PR #2, opened 2024-05-27
[ci] Init Makefile strategy and base internal tools package

rogercoll:
- Initializes the `tools` package with two tools: misspell and golangci-lint
- Adds the `Makefile.Common` and `Makefile` files with the `build-tools`, `misspell` and `lint` targets.


ChrsMark (approved):
lgtm

PR #4, opened 2024-05-30
add Go Makefile targets

rogercoll:
This PR adds the following Go commands to ease the interaction with any Go module: 

- go mod download
- go test
- go fmt
- go tidy
- go generate

To use any of these Makefile targets, the common Makefile.common must be included in each submodule's Makefile:

```
include ../../Makefile.common
```

The idea is to use them for the CI checks too.


ChrsMark (approved):
lgtm

PR #6, opened 2024-06-04
add a collector distribution for the repository components

rogercoll:
This distribution is used to ensure compatibility between Elastic's components and the OpenTelemetry collector builder. It is not intended for public release.

- ./distributions/elastic-components
- Makefile target to build the distribution using the OpenTelemetry Collector builder
- CI built check


ChrsMark (approved):
LGTM. I would love @andrzej-stencel input here as well though :).

andrzej-stencel:
Looks great, thanks @rogercoll! :tada: 

This builds the collector binary, but does not run the built collector with the sample config `distributions/elastic-components/config.yaml`, right? We could add running the binary as a next step to verify that there are no issues in runtime.

rogercoll:
> This builds the collector binary, but does not run the built collector with the sample config distributions/elastic-components/config.yaml, right? We could add running the binary as a next step to verify that there are no issues in runtime.

Good point! Make sense to add a runtime check too. I will do the followup in another PR. Thanks!

PR #7, opened 2024-06-05
Adding the ElasticProcessor

ishleenk17:
Add the ElasticProcessor Code to the Components Repo.
This Processor will be using the elastic/opentelemetry-lib for the translations of metrics from Hostmetrics to System


ishleenk17:
As discussed with @rogercoll , this PR adds the code for the ElasticInfraMetricsProcesor. The addition of component to the distro will be done as a separate PR.

processor/elasticinframetricsprocessor/config.go
```diff
@@ -0,0 +1,9 @@
+package elasticinframetricsprocessor // import "github.com/elastic/opentelemetry-collector-components/processor/elasticinframetricsprocessor"
+
+type Config struct {
+	AddSystemMetrics bool `mapstructure:"add_system_metrics"`
+}
+
+func (c *Config) Validate() error {
```
ChrsMark, line 7: Is this required if it has no actual effect?

processor/elasticinframetricsprocessor/config.go
```diff
@@ -0,0 +1,9 @@
+package elasticinframetricsprocessor // import "github.com/elastic/opentelemetry-collector-components/processor/elasticinframetricsprocessor"
+
+type Config struct {
```
ChrsMark, line 3: Wondering if we should add config tests like those that exists in the processors/components of the upstream repo: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/testdata/config.yaml

processor/elasticinframetricsprocessor/processor.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	return &ElasticProcessor{
+		cfg:       cfg,
+		logger:    set.Logger,
+		remappers: remappers,
+	}
+}
+
+// processMetrics processes the given metrics and applies remappers if configured.
+func (p *ElasticProcessor) processMetrics(_ context.Context, md pmetric.Metrics) (pmetric.Metrics, error) {
```
ChrsMark, line 42: No tests for the processor?

processor/elasticinframetricsprocessor/config.go
rogercoll: +1 Example test file that uses the testdata dir: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/config_test.go#L20

processor/elasticinframetricsprocessor/config.go
rogercoll: Good point! It has no effect in the processor module by itself, but it has once you bundle it into a collector. All components within the collector must implement this interface. On start, the collector goes one by one to each component and runs this function to validate the configuration: https://github.com/open-telemetry/opentelemetry-collector/blob/v0.101.0/otelcol/config.go#L45

rogercoll: To make this more explicit, what do you think if we add the following type assertion:

```Go
var _ component.Config = (*Config)(nil)
```

ChrsMark: I had the impression that in some components this Validate() was missing but maybe it was some kind of "inheritance" behind the scenes and that's why that wasn't an issue.

Since it's required, we can keep either way.

ishleenk17: I will keep the validate code as it is then.

processor/elasticinframetricsprocessor/config.go
ishleenk17: I have added the tests.

processor/elasticinframetricsprocessor/processor.go
ishleenk17: Added the tests.

processor/elasticinframetricsprocessor/factory.go
```diff
@@ -0,0 +1,81 @@
...
+	return processorhelper.NewMetricsProcessor(
+		ctx,
+		set,
+		cfg,
+		nextConsumer,
+		elasticinframetricsProcessor.processMetrics,
+		processorhelper.WithCapabilities(processorCapabilities))
+}
+
+func createLogsProcessor(
```
rogercoll, line 51: I think it can be removed if we are not processing logs.

processor/elasticinframetricsprocessor/factory.go
```diff
@@ -0,0 +1,81 @@
...
+	return processorhelper.NewLogsProcessor(
+		ctx,
+		set,
+		cfg,
+		nextConsumer,
+		elasticinframetricsProcessor.processLogs,
+		processorhelper.WithCapabilities(processorCapabilities))
+}
+
+func createTracesProcessor(
```
rogercoll, line 67: I think it can be removed if we are not processing traces.

processor/elasticinframetricsprocessor/processor.go
```diff
@@ -0,0 +1,63 @@
...
+	return md, nil
+}
+
+func (p *ElasticinframetricsProcessor) processLogs(_ context.Context, ld plog.Logs) (plog.Logs, error) {
+	return ld, nil
+}
+
+func (p *ElasticinframetricsProcessor) processTraces(_ context.Context, td ptrace.Traces) (ptrace.Traces, error) {
+	return td, nil
+}
```
rogercoll, line 63: Can we remove these functions? Related to previous comment

processor/elasticinframetricsprocessor/config_test.go
```diff
@@ -0,0 +1,38 @@
...
+		{
+			name:    "AddSystemMetrics false",
+			config:  Config{AddSystemMetrics: false},
+			wantErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := tt.config.Validate(); (err != nil) != tt.wantErr {
```
rogercoll, line 33: What do you think of testing the configuration by initializing the factory instead? That way, we will be able to use upstream helper functions to validate the configuration `component.ValidateConfig`.

Example: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/config_test.go#L194

In addition, I would suggest using the `testdata` directory to include some YAML configurations which will test the unmarshalling phase too.

Let me know if you want me to provide more details/help with the implementation.

processor/elasticinframetricsprocessor/config.go
```diff
@@ -0,0 +1,9 @@
...
+	AddSystemMetrics bool `mapstructure:"add_system_metrics"`
```
rogercoll, line 4: What does the `opentelemetry-lib` package do with the system metrics? Does it only add additional metrics? Or it also transforms actual ones?

I was wondering if maybe `translate_system_metrics` would be a more descriptive name.

processor/elasticinframetricsprocessor/processor_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (m *mockRemapper) Remap(scopeMetric pmetric.ScopeMetrics, metricSlice pmetric.MetricSlice, resource pcommon.Resource) {
+	m.called = true
+	m.scopeMetric = scopeMetric
+	m.metricSlice = metricSlice
+	m.resource = resource
+}
+
+func TestProcessMetrics(t *testing.T) {
```
rogercoll, line 30: What do you think adding a test case for the processor when created with newProcessor? In that case, we will be asserting the actual `opentelemetry-lib hostmetrics.NewRemapper` too.

processor/elasticinframetricsprocessor/config.go
ishleenk17: It adds system metrics, doesn't translate the existing ones.

processor/elasticinframetricsprocessor/factory.go
```diff
@@ -0,0 +1,81 @@
...
+
+// NewFactory returns a new factory for the Filter processor.
+func NewFactory() processor.Factory {
+	return processor.NewFactory(
+		typeStr,
+		createDefaultConfig,
+		processor.WithMetrics(createMetricsProcessor, component.StabilityLevelAlpha),
+		processor.WithLogs(createLogsProcessor, component.StabilityLevelAlpha),
+		processor.WithTraces(createTracesProcessor, component.StabilityLevelAlpha),
+	)
```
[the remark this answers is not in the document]
ishleenk17: Yeah, I think we have now limited the scope only to metrics. LEt me remove the logs and traces functions

processor/elasticinframetricsprocessor/config_test.go
ishleenk17: I have used the upstream validators and the testdata folder now. 
Please see if this looks fine.

processor/elasticinframetricsprocessor/processor_test.go
ishleenk17: Done

processor/elasticinframetricsprocessor/processor_test.go
```diff
@@ -0,0 +1,95 @@
...
+	testCases := []struct {
+		name               string
+		cfg                *Config
+		createMetrics      func() pmetric.Metrics
+		expectedHostname   string
+		expectedMetricName string
+		expetedMetricValue int64
+	}{
+		{
+			name: "ProcessMetrics when AddSystemMetrics is enabled",
```
rogercoll, line 24: Do we rely on the `opentelemetry-lib` tests? Should we verify a hostmetric remap?  E.g. that `"system.cpu.total.pct"` is added

processor/elasticinframetricsprocessor/factory.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+var processorCapabilities = consumer.Capabilities{MutatesData: true}
+
+// NewFactory returns a new factory for the Filter processor.
+func NewFactory() processor.Factory {
+	return processor.NewFactory(
+		typeStr,
+		createDefaultConfig,
+		processor.WithMetrics(createMetricsProcessor, component.StabilityLevelAlpha),
```
rogercoll, line 23: If we think the configuration/logic is still in a very early stage, we could use the `Development` stability level instead.

rogercoll (approved):
LGTM, just a couple of follow up comments that can be addressed in a future PR

processor/elasticinframetricsprocessor/processor_test.go
ishleenk17: I think the remapping-specific logic is being tested in the opentelemetry-lib. 
So that should be sufficient.

processor/elasticinframetricsprocessor/factory.go
ishleenk17: If I see the difference between [Development](https://github.com/open-telemetry/opentelemetry-collector/tree/main?tab=readme-ov-file#development) and [Alpha](https://github.com/open-telemetry/opentelemetry-collector/tree/main?tab=readme-ov-file#alpha), I feel alpha to be more suitable. 
Although a comment of "it might not be available as part of any distributions yet" in development is true for us. 
But since this is in Elastic repo, I think we should be good with Alpha. 
Since we will be asking some users to use this processor soon.

rogercoll: Sounds good! We can still do breaking changes `The configuration options might change often without backwards compatibility guarantees.` Thanks for the clarification

ChrsMark (approved):
Thank's for adding the tests!

PR #11, opened 2024-06-11
Add the elasticinframetricsprocessor to the components distribution manifest

rogercoll:
- Update manifest file to reference local package
- Validate make target: compile + configuration validation


ishleenk17 (approved):
Looks good!

PR #14, opened 2024-06-17
Add add_kubernetes_metrics configuration

tetianakravchenko:
Description:

Add kubernetes metrics remapping and corresponding `add_k8s_metrics` configuration

Testing:

### DIFF
--- a/processor/elasticinframetricsprocessor/config.go
+++ b/processor/elasticinframetricsprocessor/config.go
@@ -2,7 +2,7 @@ package elasticinframetricsprocessor // import "github.com/elastic/opentelemetry
 
 type Config struct {
 	AddSystemMetrics bool `mapstructure:"add_system_metrics"`
-	AddK8sMetrics    bool `mapstructure:"add_kubernetes_metrics"`
+	AddK8sMetrics    bool `mapstructure:"add_k8s_metrics"`
 }
 
 func (c *Config) Validate() error {
--- a/processor/elasticinframetricsprocessor/config.go
+++ b/processor/elasticinframetricsprocessor/config.go
@@ -2,6 +2,7 @@ package elasticinframetricsprocessor // import "github.com/elastic/opentelemetry
 
 type Config struct {
 	AddSystemMetrics bool `mapstructure:"add_system_metrics"`
+	AddK8sMetrics    bool `mapstructure:"add_kubernetes_metrics"`
 }
 
 func (c *Config) Validate() error {
--- a/processor/elasticinframetricsprocessor/config_test.go
+++ b/processor/elasticinframetricsprocessor/config_test.go
@@ -24,13 +24,18 @@ func TestLoadConfig(t *testing.T) {
 		},
 		{
 			configPath: "config.yaml",
-			id:         component.NewIDWithName(metadata.Type, "2"),
+			id:         component.NewIDWithName(metadata.Type, "all"),
 			expected:   &Config{AddSystemMetrics: true, AddK8sMetrics: true},
 		},
 		{
 			configPath: "config.yaml",
 			id:         component.NewIDWithName(metadata.Type, "with_system_metrics"),
 			expected:   &Config{AddSystemMetrics: true, AddK8sMetrics: false},
+		},
+		{
+			configPath: "config.yaml",
+			id:         component.NewIDWithName(metadata.Type, "with_kubernetes_metrics"),
+			expected:   &Config{AddSystemMetrics: false, AddK8sMetrics: true},
 		},
 	}
 
--- a/processor/elasticinframetricsprocessor/config_test.go
+++ b/processor/elasticinframetricsprocessor/config_test.go
@@ -37,6 +37,11 @@ func TestLoadConfig(t *testing.T) {
 			id:         component.NewIDWithName(metadata.Type, "with_kubernetes_metrics"),
 			expected:   &Config{AddSystemMetrics: false, AddK8sMetrics: true},
 		},
+		{
+			configPath: "config.yaml",
+			id:         component.NewIDWithName(metadata.Type, "without_system_and_k8s_metrics"),
+			expected:   &Config{AddSystemMetrics: false, AddK8sMetrics: false},
+		},
 	}
 
 	for _, tt := range tests {
--- a/processor/elasticinframetricsprocessor/config_test.go
+++ b/processor/elasticinframetricsprocessor/config_test.go
@@ -19,13 +19,18 @@ func TestLoadConfig(t *testing.T) {
 	}{
 		{
 			configPath: "config.yaml",
-			id:         component.NewIDWithName(metadata.Type, "default"),
-			expected:   &Config{AddSystemMetrics: true},
+			id:         component.NewID(metadata.Type),
+			expected:   &Config{AddSystemMetrics: true, AddK8sMetrics: true},
+		},
+		{
+			configPath: "config.yaml",
+			id:         component.NewIDWithName(metadata.Type, "2"),
+			expected:   &Config{AddSystemMetrics: true, AddK8sMetrics: true},
 		},
 		{
 			configPath: "config.yaml",
 			id:         component.NewIDWithName(metadata.Type, "with_system_metrics"),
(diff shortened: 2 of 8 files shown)


processor/elasticinframetricsprocessor/config_test.go
```diff
@@ -19,13 +19,18 @@ func TestLoadConfig(t *testing.T) {
...
+		},
+		{
+			configPath: "config.yaml",
+			id:         component.NewIDWithName(metadata.Type, "2"),
+			expected:   &Config{AddSystemMetrics: true, AddK8sMetrics: true},
 		},
 		{
 			configPath: "config.yaml",
 			id:         component.NewIDWithName(metadata.Type, "with_system_metrics"),
-			expected:   &Config{AddSystemMetrics: true},
```
gizas, line 28: Should we create a new test with_k8s_metrics?

processor/elasticinframetricsprocessor/testdata/config.yaml
```diff
@@ -1,8 +1,12 @@
-default:
+elasticinframetrics:
+elasticinframetrics/2:
```
gizas, line 2: I would change it to /all (and not /2 ). WDYT?

processor/elasticinframetricsprocessor/processor.go
```diff
@@ -29,6 +30,11 @@ func newProcessor(set processor.CreateSettings, cfg *Config) *Elasticinframetric
 			hostmetrics.NewRemapper(set.Logger, hostmetrics.WithSystemIntegrationDataset(true)),
 		}
 	}
+	if cfg.AddK8sMetrics {
```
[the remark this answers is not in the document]
gizas: From collector side yes it can. I mean the users can have both enabled

processor/elasticinframetricsprocessor/testdata/config.yaml
```diff
@@ -1,8 +1,12 @@
...
+elasticinframetrics/with_system_metrics:
   add_system_metrics: true
+  add_kubernetes_metrics: false
 
-without_system_metrics:
-  add_system_metrics: false
\ No newline at end of file
+elasticinframetrics/without_system_metrics:
+  add_system_metrics: false
+  add_kubernetes_metrics: false
```
ishleenk17, line 12: Since we are using these metrics to power just the Inventory (part of curated UI's). Should we either both enabled/disabled. Rather than giving different combinations to the user? WDYT ?

processor/elasticinframetricsprocessor/config_test.go
ishleenk17: Agree

processor/elasticinframetricsprocessor/testdata/config.yaml
tetianakravchenko: done https://github.com/elastic/opentelemetry-collector-components/pull/14/commits/cf215e8e1fe25354459ee6e668f97a1e9f170a94

processor/elasticinframetricsprocessor/processor.go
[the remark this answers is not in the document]
tetianakravchenko: thank you for pointing it out! `system_metrics` and `k8s_metrics` can be enabled at the same time, and both are enabled by default, fixed - https://github.com/elastic/opentelemetry-collector-components/pull/14/commits/cf215e8e1fe25354459ee6e668f97a1e9f170a94 also added test for this

processor/elasticinframetricsprocessor/config_test.go
tetianakravchenko: done - https://github.com/elastic/opentelemetry-collector-components/pull/14/commits/cf215e8e1fe25354459ee6e668f97a1e9f170a94

processor/elasticinframetricsprocessor/testdata/config.yaml
ChrsMark: These are just tests so I believe it does makes sense to test the various combinations unless we enforce the both_enabled/both_disabled rule/idea in the the implementation.
If the latter is the proposal I would suggest we first discuss and validate this with the wider `otel-dev` group.

tetianakravchenko: > Should we have either both enabled/disabled

now both `add_system_metrics`/`add_kubernetes_metrics` are enabled by default - https://github.com/elastic/opentelemetry-collector-components/blob/cf215e8e1fe25354459ee6e668f97a1e9f170a94/processor/elasticinframetricsprocessor/factory.go#L25-L28
I agree that it is unlikely that users would need to change the default configuration, but I would keep it configurable to explicitly disable specific remapper if needed

ishleenk17: I have checked about this in the #otel-dev. Lets see

processor/elasticinframetricsprocessor/config.go
```diff
@@ -2,6 +2,7 @@ package elasticinframetricsprocessor // import "github.com/elastic/opentelemetry
 
 type Config struct {
 	AddSystemMetrics bool `mapstructure:"add_system_metrics"`
+	AddK8sMetrics    bool `mapstructure:"add_kubernetes_metrics"`
```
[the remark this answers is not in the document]
tetianakravchenko: done https://github.com/elastic/opentelemetry-collector-components/pull/14/commits/9b5b652a486075a96eef1b94440dbc13e5ec3a17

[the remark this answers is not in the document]
ChrsMark: Thank's!

rogercoll (approved):
LGTM

ChrsMark (approved):
LGTM.

Not a blocker for this PR but in a more generic note I wonder if the naming is accurate here (`add_system_metrics` and `add_k8s_metrics` ).
I'm not sure if we have discussed about it already but the point is that this `add_` prefix implies that we add something additional/new but the truth is that we mostly transform what is coming in. @felixbarny @AlexanderWert do you have any feedback on this?

ishleenk17:
> LGTM.
> 
> Not a blocker for this PR but in a more generic note I wonder if the naming is accurate here (`add_system_metrics` and `add_k8s_metrics` ). I'm not sure if we have discussed about it already but the point is that this `add_` prefix implies that we add something additional/new but the truth is that we mostly transform what is coming in. @felixbarny @AlexanderWert do you have any feedback on this?

We are actually adding metrics while keeping the existing metrics as it is. Hence the add_*

processor/elasticinframetricsprocessor/testdata/config.yaml
```diff
@@ -1,8 +1,16 @@
...
 
-with_system_metrics:
+elasticinframetrics/with_system_metrics:
   add_system_metrics: true
+  add_k8s_metrics: false
 
-without_system_metrics:
-  add_system_metrics: false
\ No newline at end of file
+elasticinframetrics/without_system_metrics:
```
ishleenk17, line 10: More appropriate name for this ?
As we are disabling not only system but k8s as well

tetianakravchenko: done https://github.com/elastic/opentelemetry-collector-components/pull/14/commits/a8c5b77fa34c74ddb64d1c3d119c2df924f2cad9, also added it to tests - this one was missing for some reason

ishleenk17:
@tetianakravchenko : Looks good!
I hope we have tested it out as well that both the system and K8s metrics flow in when both configurations are enabled. 
By building a custom collector with this processor PR.

tetianakravchenko:
> @tetianakravchenko : Looks good!
I hope we have tested it out as well that both the system and K8s metrics flow in when both configurations are enabled.

@ishleenk17 I've updated the description with the testing information. I've tested the default config:
```
  processors:
    elasticinframetrics:
```
so both are enabled, but since I used only the `kubeletstats` receiver I focused only on the k8s related metrics (also used ingest pipeline contains only k8s related metrics).

I think your concern should be covered by this tests - https://github.com/elastic/opentelemetry-collector-components/pull/14#discussion_r1644412148 where number of initialized remappers is checked for different configurations

Issue #9, opened 2024-06-07
Setup Dependentbot for keeping opentelemetry-lib module updated


ishleenk17:
Add dependentbot.yml to trigger the upcoming elasticinframetricsprocessor to always use the latest version of opentelemetry-lib module.

PR #16, opened 2024-06-19
ci: add Go directories to dependabot configuration 

rogercoll:
Specify which directories contain Go modules. For the moment, we need to replicate each dependabot entry configuration as wildcard patterns are not supported: https://github.com/dependabot/dependabot-core/issues/2178

Fixes: https://github.com/elastic/opentelemetry-collector-components/issues/9


ishleenk17 (approved):
Looks good!

PR #39, opened 2024-07-18
PoC(wip): APM central config extension

rogercoll:


extension/apmconfigextension/internal/examples/agent/.gitignore
```diff
@@ -0,0 +1 @@
+/bin/
```
florianl, line 1: As `/bin/` starts with a `/` the absolute path of the root of the git repository is used. Is this intended?

extension/apmconfigextension/internal/examples/agent/agent/agent.go
```diff
@@ -0,0 +1,365 @@
...
+	instanceId uuid.UUID
+
+	agentDescription *protobufs.AgentDescription
+
+	opampClient client.OpAMPClient
+
+	remoteConfigStatus *protobufs.RemoteConfigStatus
+}
+
+func NewAgent(logger types.Logger, agentType string, agentVersion string) *Agent {
```
florianl, line 74: Is this function declaration enforced in some way? If not, it should be changed to `(*Agent, error)` for proper error handling.

extension/apmconfigextension/internal/examples/go.mod
```diff
@@ -0,0 +1,30 @@
...
+	github.com/rogpeppe/go-internal v1.11.0 // indirect
+	github.com/stretchr/testify v1.10.0 // indirect
+	golang.org/x/sys v0.19.0 // indirect
+	google.golang.org/protobuf v1.36.2 // indirect
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
+	gopkg.in/yaml.v2 v2.2.8 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+replace github.com/elastic/opentelemetry-collector-components/extension/apmconfigextension => ../../
```
florianl, line 30: Is there a reason for this absolute pathed replace statement?

florianl (commented):
first quick look

extension/apmconfigextension/elastic/centralconfig/fetcher.go
```diff
@@ -0,0 +1,80 @@
...
+		return apmconfig.RemoteConfig{}, fmt.Errorf("%w: service.name attribute must be provided", apmconfig.UnidentifiedAgent)
+	}
+	result, err := fw.configFetcher.Fetch(ctx, agentcfg.Query{
+		Service: serviceParams,
+	})
+	if err != nil {
+		return apmconfig.RemoteConfig{}, err
+	}
+
+	return otelapmconfig.RemoteConfig{Hash: []byte(result.Source.Etag), Attrs: result.Source.Settings}, nil
```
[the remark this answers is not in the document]
rogercoll: That will look simpler, changed in https://github.com/elastic/opentelemetry-collector-components/pull/39/commits/3272729816ae33bdf913d6cdfd29cadf74f6dfeb

extension/apmconfigextension/apmconfig/client.go
```diff
@@ -0,0 +1,41 @@
...
+// RemoteConfigClient is an adapter interface that can be used between different
+// remote configuration providers.
+type RemoteConfigClient interface {
+	// RemoteConfig returns the upstream remote configuration that needs to be applied. Empty RemoteConfig Attrs if no remote configuration is available for the specified service.
+	RemoteConfig(context.Context, *protobufs.AgentToServer) (RemoteConfig, error)
+}
+
+// RemoteConfig holds an agent remote configuration.
+type RemoteConfig struct {
+	Hash []byte
```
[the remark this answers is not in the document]
rogercoll: End up removing this custom type in favor of the already available `protobus.AgentRemoteConfig` https://github.com/elastic/opentelemetry-collector-components/pull/39/commits/3272729816ae33bdf913d6cdfd29cadf74f6dfeb

extension/apmconfigextension/config.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"time"
+
+	"github.com/elastic/opentelemetry-lib/config/configelasticsearch"
+	"go.opentelemetry.io/collector/component"
+)
+
+type Config struct {
+	RemoteConfig RemoteConfig `mapstructure:"agent_config"`
```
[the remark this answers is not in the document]
rogercoll: renamed to `AgentConfig` https://github.com/elastic/opentelemetry-collector-components/pull/39/commits/4440bfce623c6171df6cb0b96a11f13f05d34ee2

extension/apmconfigextension/elastic/centralconfig/fetcher.go
```diff
@@ -0,0 +1,80 @@
...
+	cacheDuration time.Duration
+
+	// OpAMP instanceID to service mapping
+	uidToService map[string]agentcfg.Service
+
+	logger *zap.Logger
+}
+
+func NewFetcherAPMWatcher(fetcher agentcfg.Fetcher, cacheDuration time.Duration, logger *zap.Logger) *fetcherAPMWatcher {
+	return &fetcherAPMWatcher{
```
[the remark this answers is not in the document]
rogercoll: :+1:  fixed https://github.com/elastic/opentelemetry-collector-components/pull/39/commits/c11db46b099bc661e0cab7dafa174fd0b746601b

extension/apmconfigextension/elastic/centralconfig/fetcher.go
```diff
@@ -0,0 +1,80 @@
...
+		fetcher,
+		cacheDuration,
+		make(map[string]agentcfg.Service),
+		logger,
+	}
+}
+
+func (fw *fetcherAPMWatcher) RemoteConfig(ctx context.Context, agentMsg *protobufs.AgentToServer) (apmconfig.RemoteConfig, error) {
+	var serviceParams agentcfg.Service
+	if agentMsg.AgentDescription != nil {
```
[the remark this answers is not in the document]
rogercoll: In the Go protobufs package the field can be `nil` and there's no `AgentDescription()` function but a `GetAgentDescription()` one: https://github.com/open-telemetry/opamp-go/blob/main/protobufs/opamp.pb.go#L643

extension/apmconfigextension/internal/examples/agent/agent/agent.go
rogercoll: Why should the error be propagated in this case? The agent type is always created, but if there is an error on `connect` (unexported method) just log it using the internal logger. Unfortunately, propagating the error won't help in this example because there are no fallback/methods that can be called for the Agent's struct. 

Note that this is just an example to test e2e the extension, the code was copied from https://github.com/open-telemetry/opamp-go/tree/main/internal/examples/agent and it should be removed once the OpAMP client is available in any of the Otel  SDKs.

extension/apmconfigextension/elastic/centralconfig/fetcher.go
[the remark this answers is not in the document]
dmathieu: Then, shouldn't this be

```suggestion
	if agentMsg.GetAgentDescription() != nil {
```

extension/apmconfigextension/elastic/centralconfig/fetcher.go
```diff
@@ -0,0 +1,80 @@
...
+		for _, attr := range agentMsg.GetAgentDescription().GetIdentifyingAttributes() {
```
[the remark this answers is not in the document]
rogercoll: good point, https://github.com/elastic/opentelemetry-collector-components/pull/39/commits/62a96128d0620bec3b436d13775ba2082d344879

extension/apmconfigextension/internal/examples/agent/.gitignore
rogercoll: good catch, it should be relative https://github.com/elastic/opentelemetry-collector-components/pull/39/commits/62a96128d0620bec3b436d13775ba2082d344879

extension/apmconfigextension/internal/examples/agent/agent/agent.go
```diff
@@ -0,0 +1,365 @@
...
+		agent.logger.Errorf(context.Background(), "Cannot connect OpAMP client: %v", err)
+		return nil
+	}
+
+	return agent
+}
+
+func (agent *Agent) connect() error {
+	httpClient := client.NewHTTP(agent.logger)
+	httpClient.SetPollingInterval(5 * time.Second)
```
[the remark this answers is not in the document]
rogercoll: The default opamp interval is 30s, but this code is just an example app so messages communication can be seen more frequently.

extension/apmconfigextension/internal/examples/agent/agent/agent.go
```diff
@@ -0,0 +1,365 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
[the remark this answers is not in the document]
rogercoll: @dmathieu @florianl I end up removing the example given that we can rely on the upstream example https://github.com/open-telemetry/opamp-go/tree/main/internal/examples/agent or https://github.com/elastic/opamp-java

The example was intended for locally e2e the extension, as OpAMP is not bundled in any SDK yet.

extension/apmconfigextension/internal/examples/go.mod
rogercoll: example removed

extension/apmconfigextension/go.mod
```diff
@@ -0,0 +1,105 @@
...
+	golang.org/x/net v0.34.0 // indirect
+	golang.org/x/sys v0.29.0 // indirect
+	golang.org/x/text v0.21.0 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20250102185135-69823020774d // indirect
+	google.golang.org/grpc v1.70.0 // indirect
+	google.golang.org/protobuf v1.36.4
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+replace github.com/elastic/opentelemetry-collector-components/internal/testutil => ../../internal/testutil
```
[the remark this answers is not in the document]
rogercoll: It forces using the local https://github.com/elastic/opentelemetry-collector-components/blob/main/internal/testutil/go.mod internal packages. When using a Go monorepo, this replacement statements help keeping all the modules in sync. The OpenTelemetry community uses the [crosslink](https://pkg.go.dev/go.opentelemetry.io/build-tools/crosslink#section-readme) to automate this process.

extension/apmconfigextension/README.md
```diff
@@ -0,0 +1,55 @@
...
+(optional) attributes. These attributes **must** be set on the [AgentDescription.identifying_attributes](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionidentifying_attributes) field during the first send [AgentToServer](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoserver-message) message. As the `AgentDescription` should not be send if not changed, the extension will maintain and internal mapping between the `agent.instance_id` and its service identifing attributes.
+
+The [ServerToAgent.ReportFullState flag](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#servertoagentflags) will be set in the following cases:
+
+- The agent did not include the `service.name` identifing attributes during the
+  first message.
+- The OpAMP server was not able to identify the agent.
+
+The agent **must** return a message with the corresponding `AgentDescription.identifying_attributes`.
+
```
florianl, line 24: What is the recommended way to debug this? Are the two named cases related to each other with a OR or an AND connection wrt ServerToAgent.ReportFullState flag?

extension/apmconfigextension/elastic/centralconfig/fetcher.go
```diff
@@ -0,0 +1,94 @@
...
+		configFetcher: fetcher,
+		cacheDuration: cacheDuration,
+		uidToService:  make(map[string]agentcfg.Service),
+		logger:        logger,
+	}
+}
+
+func (fw *fetcherAPMWatcher) RemoteConfig(ctx context.Context, agentMsg *protobufs.AgentToServer) (*protobufs.AgentRemoteConfig, error) {
+	var serviceParams agentcfg.Service
+	if agentDescription := agentMsg.GetAgentDescription(); agentDescription != nil {
```
florianl, line 58: It looks like there are some logical bugs in this if-else branch:

Should the `else` branch with `serviceParams = fw.uidToService[string(agentMsg.GetInstanceUid())]` be considered a fast path or fall back?

With `fw.uidToService[string(agentMsg.GetInstanceUid())] = serviceParams` in the loop over the attributes, `serviceParams` with invalid values, e.g. empty Name, can be written into the cache. For both cases of `attr.GetValue().GetStringValue()` there should be a validation step.

extension/apmconfigextension/factory.go
```diff
@@ -0,0 +1,77 @@
...
+	extCfg := cfg.(*Config)
+	elasticsearchRemoteConfig := func(ctx context.Context, host component.Host, telemetry component.TelemetrySettings) (apmconfig.RemoteConfigClient, error) {
+		esClient, err := extCfg.AgentConfig.Elasticsearch.ToClient(ctx, host, telemetry)
+		if err != nil {
+			return nil, err
+		}
+		fetcher := agentcfg.NewElasticsearchFetcher(esClient, extCfg.AgentConfig.CacheDuration, telemetry.Logger)
+		go func() {
+			err := fetcher.Run(ctx)
+			if err != nil {
```
florianl, line 69: nit: simplify code and limit scope of `err`
```suggestion
			if err := fetcher.Run(ctx); err != nil {
```

extension/apmconfigextension/factory.go
```diff
@@ -0,0 +1,77 @@
...
+
+func createExtension(_ context.Context, set extension.Settings, cfg component.Config) (extension.Extension, error) {
+	extCfg := cfg.(*Config)
+	elasticsearchRemoteConfig := func(ctx context.Context, host component.Host, telemetry component.TelemetrySettings) (apmconfig.RemoteConfigClient, error) {
+		esClient, err := extCfg.AgentConfig.Elasticsearch.ToClient(ctx, host, telemetry)
+		if err != nil {
+			return nil, err
+		}
+		fetcher := agentcfg.NewElasticsearchFetcher(esClient, extCfg.AgentConfig.CacheDuration, telemetry.Logger)
+		go func() {
```
florianl, line 67: Depending on the Go scheduler, there can be a race condition, where this anonymous Go function is never started, when `return centralconfig.NewFetcherAPMWatcher(..)` is faster.

extension/apmconfigextension/metadata.yaml
```diff
@@ -0,0 +1,15 @@
+type: apmconfig
+
+status:
+  class: extension
+  stability:
+    development: [extension]
+  distributions: []
+  codeowners:
+    active: [rogercoll]
```
florianl, line 9: Should a codeowner always be a person or a team?

extension/apmconfigextension/README.md
rogercoll: They are related in the sense that the server won't be able to proceed for the connected client. afaik the only OpAMP mechanism to communicate to the client that some information is missing is by replying with the `ServerToAgent.ReportFullState` flag (and the error message). https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#servertoagentflags

extension/apmconfigextension/metadata.yaml
rogercoll: Good point, do we have a team's handle? I think using `otel-dev` involves many subteams.

extension/apmconfigextension/elastic/centralconfig/fetcher.go
rogercoll: nice catch, I fixed by only writing to the cahe if service.name is set https://github.com/elastic/opentelemetry-collector-components/pull/39/commits/bdbc6fa56ed01690bd294a275f8805f8d9ef3fd3

extension/apmconfigextension/factory.go
rogercoll: Would be using a sync.Waitgroup and releasing it at first inside the anonymous Go function solve this race condition?

extension/apmconfigextension/metadata.yaml
rockdaboot: @sboomsma can possibly set up a new team for the (new) otel data team. Would that be sufficient?

extension/apmconfigextension/factory.go
florianl: yes, that is the most straight forward way to make sure this go routine is created and exists.

rogercoll: great! will do that. Do you know about any external source that explains this behavior? My understanding was that the `go func()` statement was already registering the routine immediately to the scheduler, even if the parent function returns earlier (in this case it is not an issue if NewFetcherAPMWatcher is scheduled first).

extension/apmconfigextension/elastic/centralconfig/fetcher.go
```diff
@@ -0,0 +1,97 @@
...
+	uidToService map[string]agentcfg.Service
+
+	logger *zap.Logger
+}
+
+func NewFetcherAPMWatcher(fetcher agentcfg.Fetcher, cacheDuration time.Duration, logger *zap.Logger) *fetcherAPMWatcher {
+	return &fetcherAPMWatcher{
+		configFetcher: fetcher,
+		cacheDuration: cacheDuration,
+		uidToService:  make(map[string]agentcfg.Service),
```
florianl, line 51: With every call of `RemoteConfig()` this map `uidToService` is populated and never cleaned up. Therefore it just grows without limits.There is also the risk, that `RemoveConfig()` fetches something from this map - via `serviceParams, ok := fw.uidToService[string(agentMsg.GetInstanceUid())]` that is out of date and not work any longer.

extension/apmconfigextension/factory.go
florianl: I'm not aware of any guarantees by the scheduler that make sure that `go func()` is executed and does not race with the immediate `return` statement. But I had to debug and investigate such issues often enough, that I'm looking for such things and name it.

rogercoll: sounds good, let's keep it safe for now https://github.com/elastic/opentelemetry-collector-components/pull/39/commits/56ba5a9a7a8841a12536c0e83d238eede9256923

extension/apmconfigextension/elastic/centralconfig/fetcher.go
rogercoll: Good point, I think removals should be done on the agent disconnects (AgentDisconnect message), but it would be more appropriate to handle in a follow-up PR https://github.com/elastic/opentelemetry-collector-components/issues/398

florianl (approved):

PR #40, opened 2024-07-18
[elasticinframetrics]Fix failing test due to upgrade to v0.105.0

lahsivjar:
This is a minimal fix for the current failures happening in main. Creating this since https://github.com/elastic/opentelemetry-collector-components/pull/38 had concerns about backward compatibility.

### DIFF
--- a/processor/elasticinframetricsprocessor/generated_component_test.go
+++ b/processor/elasticinframetricsprocessor/generated_component_test.go
@@ -66,7 +66,7 @@ func TestComponentLifecycle(t *testing.T) {
 	cfg := factory.CreateDefaultConfig()
 	sub, err := cm.Sub("tests::config")
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


rogercoll:
I think we should go with https://github.com/elastic/opentelemetry-collector-components/pull/38 

This fix seems not to solve the linter issue yet:

```
Error: elasticinframetricsprocessor/factory.go:50:6: SA1019: processor.CreateSettings is deprecated: [v0.103.0] Use processor.Settings instead. (staticcheck)
```

Wdyt?

lahsivjar:
> I think we should go with https://github.com/elastic/opentelemetry-collector-components/pull/38

+1, the only reason would be the backward compatibility. I am not sure if that would be a concern but I am not familiar with the details of our release distro.

lahsivjar:
Closing this as there is agreement to go with #38

PR #41, opened 2024-07-19
Add stateless span metrics connector

lahsivjar:
The span metrics connector v2 is similar to the upstream [spanmetricsconnector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector) but is stateless. The idea is that aggregations can be performed on a requirement basis using more apt components (like interval processor). Will be pitched to OTel community as an upgrade to the older span metrics connector once it is mature enough.

Related to: https://github.com/elastic/opentelemetry-dev/issues/335


connector/spanmetricsconnectorv2/connector.go
```diff
@@ -0,0 +1,85 @@
...
+				multiError = errors.Join(multiError, spansHist.update(ctx, span.Attributes(), durationInNs))
+			}
+		}
+
+		if len(spansHist.counts) == 0 {
+			continue // don't add an empty resource
+		}
+
+		processedResource := processedMetrics.ResourceMetrics().AppendEmpty()
+		resourceSpan.Resource().Attributes().CopyTo(processedResource.Resource().Attributes())
```
axw, line 73: Should there be configuration to filter resource attributes, like in spanmetricsconnector?

axw (commented):
I've only skimmed - looks good, just a couple of high level questions about the config

connector/spanmetricsconnectorv2/go.mod
```diff
@@ -0,0 +1,61 @@
+module github.com/elastic/opentelemetry-collector-components/connector/spanmetricsconnectorv2
```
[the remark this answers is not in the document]
lahsivjar: Yeah, I was planning to create a separate followup PR for this once this one is merged.

connector/spanmetricsconnectorv2/config.go
```diff
@@ -0,0 +1,165 @@
...
+
+// Config for the connector
+type Config struct {
+	Spans map[string]MetricInfo `mapstructure:"spans"`
+}
+
+// MetricInfo for a data type
+type MetricInfo struct {
+	Description string            `mapstructure:"description"`
+	Attributes  []AttributeConfig `mapstructure:"attributes"`
```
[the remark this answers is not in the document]
lahsivjar: The `countconnector` calls it `attributes` and it made more sense to me at the time of writing since it relates to the OTel data model. Though, it might be prudent to keep it dimension if we are planning on upstreaming this. What was your reasoning behind using dimensions?

connector/spanmetricsconnectorv2/connector.go
lahsivjar: I didn't add it because we didn't need it at the moment. Also, my reasoning for not adding it was because I wanted the connector to only do what is required for span -> metrics production. Other processing (like dropping attributes) if needed can be done by a separate processor.

connector/spanmetricsconnectorv2/config.go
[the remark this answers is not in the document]
axw: Just to minimise UX differences between spanmetricsconnector and spanmetricsconnectorv2. But if `attributes` is in use in the `countconnector`, I can also buy consolidating on one and switching to `attributes`.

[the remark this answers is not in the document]
lahsivjar: If the proposal gets accepted to replace the current span metrics connector then it would be a breaking change anyway (producing aggregated metrics vs non-aggregated ones) so we could take the opportunity to consolidate the configuration. If required, we can provide a form of aliasing of dimensions -> attributes for compatibility.

connector/spanmetricsconnectorv2/histogram.go
```diff
@@ -0,0 +1,155 @@
...
+					if v != 0 {
+						countAttrs.PutDouble(attr.Key, float64(v))
+					}
+				}
+			}
+		}
+
+		// Missing necessary attributes to be counted
+		if countAttrs.Len() != len(md.attrs) {
+			continue
```
carsonip, line 100: question: when do we expect attributes to be missing? Should it log / raise an error?

lahsivjar: > question: when do we expect attributes to be missing? 

If the span added doesn't have any of the attributes that we configured. The `md.attrs` is the configured attribute.

> Should it log / raise an error?

No, this is expected behavior. Look at the `with_attributes` and `with_missing_attributes` test files.

connector/spanmetricsconnectorv2/config.go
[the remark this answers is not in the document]
axw: Fair enough. Switching to `attributes` sounds fine to me.

connector/spanmetricsconnectorv2/connector.go
axw: Ah true, we could indeed drop attributes in a subsequent processor since it's stateless. SGTM.

connector/spanmetricsconnectorv2/connector.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (sm *spanMetrics) ConsumeTraces(ctx context.Context, td ptrace.Traces) error {
+	var multiError error
+	processedMetrics := pmetric.NewMetrics()
+	processedMetrics.ResourceMetrics().EnsureCapacity(td.ResourceSpans().Len())
+	for i := 0; i < td.ResourceSpans().Len(); i++ {
+		resourceSpan := td.ResourceSpans().At(i)
+		// TODO (lahsivjar): add support for exponential histogram and summary
+		spansHist := newExplicitHistogram(sm.spansMetricDefs)
```
felixbarny, line 51: Is the idea to use exponential histograms by default once support for that has been added? I think that would be preferable over using explicit boundaries.

lahsivjar: I haven't thought about what should be the default but it makes sense to make exponential histograms as default.

connector/spanmetricsconnectorv2/histogram.go
```diff
@@ -0,0 +1,155 @@
...
+				switch typeAttr := attrVal.Type(); typeAttr {
+				case pcommon.ValueTypeInt:
+					countAttrs.PutInt(attr.Key, attrVal.Int())
+				case pcommon.ValueTypeDouble:
+					countAttrs.PutDouble(attr.Key, attrVal.Double())
+				default:
+					countAttrs.PutStr(attr.Key, attrVal.Str())
+				}
+			} else if attr.DefaultValue != nil {
+				switch v := attr.DefaultValue.(type) {
```
axw, line 81: Since the cases below aren't exhaustive of all JSON types, we should probably validate the default value types.

connector/spanmetricsconnectorv2/histogram.go
```diff
@@ -0,0 +1,155 @@
...
+
+	bounds []float64
+	counts []uint64
+}
+
+func newAttrExplicitHistogram(attrs pcommon.Map, bounds []float64) *attrExplicitHistogram {
+	return &attrExplicitHistogram{
+		attrs:  attrs,
+		bounds: bounds,
+		counts: make([]uint64, len(bounds)+1),
```
axw, line 62: Maybe leave a comment that the last bucket is for `[bounds[len(bounds)-1]..+Inf)` (IIUC)?

connector/spanmetricsconnectorv2/config.go
```diff
@@ -0,0 +1,165 @@
...
+		return errors.New("histogram buckets missing")
+	}
+	return nil
+}
+
+func (i *MetricInfo) validateAttributes() error {
+	for _, attr := range i.Attributes {
+		if attr.Key == "" {
+			return fmt.Errorf("attribute key missing")
+		}
```
axw, line 102: If there's no better option, can we validate the DefaultValue types here?

connector/spanmetricsconnectorv2/connector.go
```diff
@@ -0,0 +1,85 @@
...
+			scopeSpan := resourceSpan.ScopeSpans().At(j)
+
+			for k := 0; k < scopeSpan.Spans().Len(); k++ {
+				span := scopeSpan.Spans().At(k)
+				var durationInNs float64
+				startTime := span.StartTimestamp()
+				endTime := span.EndTimestamp()
+				if endTime > startTime {
+					durationInNs = float64(endTime - startTime)
+				}
```
axw, line 63: Maybe pass in as a `time.Duration`?

connector/spanmetricsconnectorv2/histogram.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+}
+
+func (c *explicitHistogram) update(ctx context.Context, attrs pcommon.Map, value float64) error {
+	var multiError error
+	for name, md := range c.metricDefs {
+		countAttrs := pcommon.NewMap()
+		for _, attr := range md.attrs {
```
axw, line 70: I suppose it's OK for now, and we can refactor when adding summary & exponential histogram support, but this seems like it should be up a level in the main connector code, rather than in the explicitHistogram code.

connector/spanmetricsconnectorv2/testdata/with_custom_histogram_buckets/config.yaml
```diff
@@ -0,0 +1,8 @@
+spanmetricsv2:
+  spans:
+    trace.span.duration:
+      description: Span duration with custom histogram buckets
+      unit: "s"
+      histogram:
+        explicit:
+          buckets: [0.001, 0.1, 10, 100]
```
axw, line 8: Maybe alter the buckets so one of the span durations is greater than the greatest bucket value? i.e. to cover the case where SearchFloat64s returns len(buckets)

axw (commented):
LGTM overall, just a handful of minor details.

connector/spanmetricsconnectorv2/config.go
```diff
@@ -0,0 +1,165 @@
...
+type MetricInfo struct {
+	Description string            `mapstructure:"description"`
+	Attributes  []AttributeConfig `mapstructure:"attributes"`
+	Unit        MetricUnit        `mapstructure:"unit"`
+	Histogram   HistogramConfig   `mapstructure:"histogram"`
+}
+
+type AttributeConfig struct {
+	Key          string `mapstructure:"key"`
+	DefaultValue any    `mapstructure:"default_value"`
```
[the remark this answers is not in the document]
lahsivjar: I don't think we can do anything about `any` but we should be able to parse this and convert it to `pcommon.Value`, let me give that a shot.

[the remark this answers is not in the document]
lahsivjar: Fixed this in https://github.com/elastic/opentelemetry-collector-components/pull/41/commits/1f310e83dc7cc1cbaa37fa1fbe310edb1137f3fb

connector/spanmetricsconnectorv2/histogram.go
lahsivjar: I have updated the code to parse `any` directly to `pcommon.Value` which is then directly copied to the attribute map (https://github.com/elastic/opentelemetry-collector-components/pull/41/commits/1f310e83dc7cc1cbaa37fa1fbe310edb1137f3fb). Let me know if that solves this concern.

connector/spanmetricsconnectorv2/histogram.go
lahsivjar: Added comments to the `attrExplicitHistogram` struct.

connector/spanmetricsconnectorv2/config.go
lahsivjar: Checkout https://github.com/elastic/opentelemetry-collector-components/commit/1f310e83dc7cc1cbaa37fa1fbe310edb1137f3fb

connector/spanmetricsconnectorv2/histogram.go
lahsivjar: Yes, I plan to abstract explicit histogram, exponential histogram, and summary behind a common interface.

connector/spanmetricsconnectorv2/testdata/with_custom_histogram_buckets/config.yaml
lahsivjar: Done!

connector/spanmetricsconnectorv2/connector.go
lahsivjar: Done in https://github.com/elastic/opentelemetry-collector-components/pull/41/commits/8e2b75defcef43f2471516988d9bb299bf931117

connector/spanmetricsconnectorv2/histogram.go
axw: Perfect, thank you

connector/spanmetricsconnectorv2/histogram.go
```diff
@@ -0,0 +1,156 @@
...
+	metricName string, attrs pcommon.Map, value float64, hCfg HistogramConfig,
+) error {
+	if _, ok := c.counts[metricName]; !ok {
+		c.counts[metricName] = make(map[[16]byte]*attrExplicitHistogram)
+	}
+
+	key := noAttributes
+	if attrs.Len() > 0 {
+		key = pdatautil.MapHash(attrs)
+	}
```
axw, line 119: I worry about the lack of attribute comparison, and possibility of hash collisions. Should be quite unlikely since this is only within a single batch, per metric name, though.

axw (approved):
Looks great, thank you!

connector/spanmetricsconnectorv2/histogram.go
axw: It does, thank you.

connector/spanmetricsconnectorv2/config.go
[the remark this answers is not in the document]
axw: Heh yeah it might be a bit, but since other metrics (e.g. ones coming from SDKs) can have arbitrary attributes, I think it's fine.

connector/spanmetricsconnectorv2/histogram.go
lahsivjar: I share the concern. I will later look into doing something about this.

connector/spanmetricsconnectorv2/config.go
```diff
@@ -0,0 +1,170 @@
...
+func (c *Config) Validate() error {
+	for name, info := range c.Spans {
+		if name == "" {
+			return errors.New("spans: metric name missing")
+		}
+		if info.Unit == "" {
+			return errors.New("spans: metric unit missing")
+		}
+		if err := info.validateHistogram(); err != nil {
+			return fmt.Errorf("sapn attributes: metric %q: %w", name, err)
```
carsonip, line 80:
```suggestion
			return fmt.Errorf("spans histogram: metric %q: %w", name, err)
```

carsonip (commented):
nit, otherwise lgtm

connector/spanmetricsconnectorv2/histogram.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if _, ok := c.counts[metricName][key]; !ok {
+		c.counts[metricName][key] = newAttrExplicitHistogram(attrs, hCfg.Explicit.Buckets)
+	}
+
+	hist := c.counts[metricName][key]
+	hist.count++
```
axw, line 126: Just occurred to me: we're not compensating for sampling here, like we do in apm-aggregation. Is that intentional? Maybe for a follow up?

lahsivjar: Yup, planned as a followup. I didn't think it would be too vital for the PoC so I have it a bit further down in the TODO list.

connector/spanmetricsconnectorv2/config.go
lahsivjar: Fixed!

carsonip (approved):

PR #66, opened 2024-07-30
Add dependabot grouping for otel-collector-contrib repo

lahsivjar:
Should make it a bit easier to review and merge the OTel version bump PRs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,11 @@ updates:
       - automation
     groups:
       otel-dependencies:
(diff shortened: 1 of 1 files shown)


rogercoll (approved):
Great!

rogercoll:
I don't have a strong opinion, but I agree that the otel core happens first, and the contrib is updated accordingly. This repo can be seen as an extension of the contrib one, thus we can keep the same strategy (different groups). If the repo becomes too overwhelmed, I don't think there would be an issue joining them into a single otel group.

PR #95, opened 2024-08-15
[connector/spanmetricsv2] Scale spans based on adjusted count

lahsivjar:
Calculates the adjusted count based on [OTEP-235](https://github.com/open-telemetry/oteps/blob/main/text/trace/0235-sampling-threshold-in-trace-state.md). Adjusted count represents the number of spans that a single sampled span represents -- similar to APM's representative count. OTEP-235 is now [supported by probabilistic sampler](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31894) too.

The logic should probably be controlled with a config option but that is not done yet.

Related to: https://github.com/elastic/opentelemetry-dev/issues/378
Also, related to: https://github.com/elastic/opentelemetry-dev/issues/372


gregkalapos (approved):
The logic here looks good to me.

I asked agent devs if we get `r,p` value or `th` by default, but it seems even `th` is not always propagated by probabilistic samplers. I think there isn't much we can do about this now and the PR makes the situation better, so 👍 from me.

connector/spanmetricsconnectorv2/connector.go
```diff
@@ -88,3 +92,58 @@ func (sm *spanMetrics) ConsumeTraces(ctx context.Context, td ptrace.Traces) erro
...
+		return 0
+	case p > 63:
+		// Invalid value, default to 1
+		return 1
+	default:
+		// TODO (lahsivjar): Should we handle fractional adjusted count?
+		// One way to do this would be to scale the values in the histograms
+		// for some precision.
+		return uint64(math.Pow(2, float64(p)))
+	}
```
axw, line 148: I'm not sure about this...

My understanding is that the old (experimental) p-value/r-value spec is superseded by OTEP 235, which introduces t-value.

If we need to support both, perhaps it should be a separate processor that transforms the old tracestate? We might want to do something similar to handle the Elastic APM tracestate.

axw (commented):
Looks good overall. If we're planning to upstream this, then I think having the p-value handling in there may be problematic.

> The logic should probably be controlled with a config option but that is not done yet.

In case you haven't seen it yet: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33632

connector/spanmetricsconnectorv2/connector.go
lahsivjar: > My understanding is that the old (experimental) p-value/r-value spec is superseded by OTEP 235, which introduces t-value.

+1. I didn't know that [probabilistic sampling processor already supported the new specs](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31894) and wanted to continue supporting that processor.

> If we need to support both, perhaps it should be a separate processor that transforms the old tracestate? We might want to do something similar to handle the Elastic APM tracestate.

This makes more sense. I have updated the PR to remove the p-value handling in favor of OTEP 235.

axw (approved):
looks great, thanks for simplifying!

PR #109, opened 2024-09-09
Add lsmintervalprocessor for persistence backed aggregation

lahsivjar:
Moving the LSM interval processor from the PoC branch to this repository. Needed to copy the [internal identity package from opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/internal/exp/metrics/identity).

Will follow-up with some other suggested improvements to the component incrementally.

Related to:

https://github.com/elastic/opentelemetry-dev/issues/316
https://github.com/elastic/opentelemetry-dev/issues/390


processor/lsmintervalprocessor/internal/merger/model.go
```diff
@@ -0,0 +1,389 @@
...
+	case pmetric.NumberDataPointValueTypeDouble:
+		to.SetDoubleValue(to.DoubleValue() + from.DoubleValue())
+	}
+}
+
+func mergeDeltaHistogramDP(from, to pmetric.HistogramDataPoint) {
+	// Explicit bounds histogram should have same pre-defined buckets.
+	// However, it is possible that the boundaries got updated. In such
+	// scenarios we can't calculate the histogram for conflicting
+	// boundaries. In practical situations, we should not see such
```
axw, line 371: Could we assume uniform distribution across buckets, and split the buckets while merging and distribute the value according to the split weight? e.g. if we needed to split the bucket `[1..100]` into `[1..25]` and `[26..100]`, then we would assign 25% of the value to the first bucket and 75% to the second one.

The new buckets would be based on merging overlapping intervals: https://www.enjoyalgorithms.com/blog/merge-overlapping-intervals

(I think this can be a TODO, just saying I don't think it's so intractable)

processor/lsmintervalprocessor/processor.go
```diff
@@ -0,0 +1,388 @@
...
+			return fmt.Errorf("failed to close database: %w", err)
+		}
+		// All future operations are invalid after db is closed
+		p.db = nil
+	}
+	return nil
+}
+
+func (p *Processor) Capabilities() consumer.Capabilities {
+	return consumer.Capabilities{MutatesData: true}
```
axw, line 202: I wonder if the "MutatesData" capability is a bit too coarse-grained. IIANM we're only removing metrics from the slice, but we don't otherwise modify the metric contents, and this capability would cause them all to be cloned I think.

Maybe something to keep an eye on in benchmarks, and open an issue about upstream.

processor/lsmintervalprocessor/processor.go
```diff
@@ -0,0 +1,388 @@
...
+		return errors.Join(append(errs, fmt.Errorf("failed to marshal value to proto binary: %w", err))...)
+	}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	keys := make([][]byte, len(p.intervals))
+	for i, ivl := range p.intervals {
+		key := merger.NewKey(ivl, p.processingTime)
+		keys[i], err = key.Marshal()
```
axw, line 242: Could we pre-marshal the keys when the processing time ticks over?

axw (approved):
Looks great. I didn't look at the tests much, so didn't entirely validate correctness, but I think we can get it in and iterate.

Are you intending to add a cardinality limit later? I'm wondering what the configuration would look like.

distributions/elastic-components/manifest.yaml
```diff
@@ -19,11 +19,13 @@ receivers:
 processors:
   - gomod: github.com/elastic/opentelemetry-collector-components/processor/elasticinframetricsprocessor v0.0.0
   - gomod: github.com/elastic/opentelemetry-collector-components/processor/elastictraceprocessor v0.0.0
+  - gomod: github.com/elastic/opentelemetry-collector-components/processor/lsmintervalprocessor v0.0.0
```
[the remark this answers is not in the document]
lahsivjar: Adding to distribution is different than making it a part of Elastic's distribution. It is used to ensure things are compatible with the OTel binary.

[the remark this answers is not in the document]
lahsivjar: For more information: https://github.com/elastic/opentelemetry-collector-components/blob/main/distributions/elastic-components/README.md?plain=1

processor/lsmintervalprocessor/internal/merger/model.go
lahsivjar: > Could we assume uniform distribution across buckets

I was initially hesitant to make this assumption but now that I think about it, this state will be temporary -- maybe when the service is updating with a new histogram config. And, assuming a temporary state, we can make this assumption.

> The new buckets would be based on merging overlapping intervals:

If we make the assumption of uniform distribution, we can, technically, keep the same buckets as before and normalize the new buckets. Not sure if updating the buckets is a good idea since they are explicit buckets specified by the users. If we change the buckets we might end up with concerns about unknown buckets, though this concern would not be a big deal for our (Elastic's) use cases but it may end up being troublesome if a user just wants to roll up the metrics and index them using this processor.

I will update the TODO and create an issue for this.

processor/lsmintervalprocessor/processor.go
```diff
@@ -0,0 +1,388 @@
...
+
+func (p *Processor) ConsumeMetrics(ctx context.Context, md pmetric.Metrics) error {
+	var errs []error
+	v := merger.Value{Metrics: pmetric.NewMetrics()}
+	md.ResourceMetrics().RemoveIf(func(rm pmetric.ResourceMetrics) bool {
+		rm.ScopeMetrics().RemoveIf(func(sm pmetric.ScopeMetrics) bool {
+			sm.Metrics().RemoveIf(func(m pmetric.Metric) bool {
+				switch t := m.Type(); t {
+				case pmetric.MetricTypeEmpty, pmetric.MetricTypeGauge, pmetric.MetricTypeSummary:
+					return false
```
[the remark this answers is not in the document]
lahsivjar: I will add a TODO. Dealing with summaries and gauges both is a bit of a pain, especially with delta temporality.

processor/lsmintervalprocessor/processor.go
lahsivjar: Good idea. I will give it a shot.

processor/lsmintervalprocessor/internal/merger/model.go
```diff
@@ -0,0 +1,389 @@
...
+	"encoding/binary"
+	"errors"
+	"time"
+
+	"go.opentelemetry.io/collector/pdata/pmetric"
+
+	"github.com/elastic/opentelemetry-collector-components/processor/lsmintervalprocessor/internal/identity"
+)
+
+// TODO: Think about multitenancy, should be part of the key
```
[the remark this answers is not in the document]
lahsivjar: > I think we briefly discussed this but I can't recall the details of our discussion: an alternative may be to create a separate directory/PebbleDB instance per tenant

Yeah, but I don't think we concluded anything. This is one way, however, we would need to identify the tenant first. Also, I think using key taints is better than separate directory in pebble since the keys would partition the data and there is almost negligible chances of any data mixing between tenants due to how data is kept at rest.

> We can keep memory usage in check by limiting concurrency of merges.

Currently, we have this config as `1`, so no concurrent merges.

processor/lsmintervalprocessor/processor.go
lahsivjar: Hmm, I decided to add a TODO for now. I think we should pause doing this a bit and see what ends up in the key. For example: if the tenant information ends up in the key then it might not make sense to cache things.

PR #135, opened 2024-09-24
[connector/spanmetricsv2] Aggregate spans as 2 counters

lahsivjar:
Allows aggregating the span as 2 separate counters with customizable suffixes. The counters represent sum of span duration and count of spans respectively.

Related to: https://github.com/elastic/opentelemetry-dev/issues/372

### DIFF
--- a/connector/spanmetricsconnectorv2/config/config.go
+++ b/connector/spanmetricsconnectorv2/config/config.go
@@ -37,6 +37,9 @@ const (
 	// error of less than 5%.
 	// Ref: https://opentelemetry.io/docs/specs/otel/metrics/sdk/#base2-exponential-bucket-histogram-aggregation
 	defaultExponentialHistogramMaxSize = 160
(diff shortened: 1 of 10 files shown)


connector/spanmetricsconnectorv2/config/config.go
```diff
@@ -57,6 +60,82 @@ type Config struct {
...
+			}
+			if info.Counters.CountSuffix == "" {
+				info.Counters.CountSuffix = defaultCountersCountSuffix
+			}
+		}
+		c.Spans[k] = info
+	}
+	return nil
+}
+
```
lahsivjar, line 138: [**For reviewers**] This bit of code is mostly moved from below. The only change is using `info.noAggregatorDefined` to check if no aggregator has been defined for a metric configuration.

carsonip (approved):
lgtm thanks

Issue #145, opened 2024-09-26
[processor/lsminterval] Support summary


lahsivjar:
Support for merging summaries by the lsminterval processor is required.

PR #146, opened 2024-09-26
Add support for merging summary metrics

lahsivjar:
This PR allows merging summaries, however, there are some assumptions involved. Assumptions are required because summary metrics does not have a defined temporality. The documentation in the PR goes over the assumptions.

Closes: https://github.com/elastic/opentelemetry-collector-components/issues/145

### DIFF
--- a/processor/lsmintervalprocessor/config.go
+++ b/processor/lsmintervalprocessor/config.go
@@ -29,9 +29,9 @@ type Config struct {
 	// Directory is the data directory used by the database to store files.
 	// If the directory is empty in-memory storage is used.
 	Directory string `mapstructure:"directory"`
-	// PassThrough is a configuration that determines whether gauge and
-	// summary metrics should be passed through as they are or aggregated.
-	// This is because they lead to lossy aggregations.
+	// PassThrough is a configuration that determines whether summary
+	// metrics should be passed through as they are or aggregated. This
+	// is because they lead to lossy aggregations.
 	PassThrough PassThrough `mapstructure:"pass_through"`
 	// Intervals is a list of interval configuration that the processor
 	// will aggregate over. The interval duration must be in increasing


processor/lsmintervalprocessor/config.go
```diff
@@ -29,6 +29,10 @@ type Config struct {
 	// Directory is the data directory used by the database to store files.
 	// If the directory is empty in-memory storage is used.
 	Directory string `mapstructure:"directory"`
+	// PassThrough is a configuration that determines whether gauge and
+	// summary metrics should be passed through as they are or aggregated.
+	// This is because they lead to lossy aggregations.
```
axw, line 34: Should it just be "summary" rather than "gauge and summary"?

processor/lsmintervalprocessor/processor_test.go
```diff
@@ -39,32 +39,42 @@ import (
...
+						`set(resource.attributes["dependent_attr"], Concat([attributes["aaa"], "dependent"], "-"))`,
+					},
+				},
+			},
+			PassThrough: func() (pc PassThrough) {
+				if tc.passThrough {
+					pc.Summary = true
+				}
+				return
+			}(),
```
axw, line 75:
```suggestion
			PassThrough: PassThrough{
				Summary: tc.passThrough,
			},
```

axw (commented):
Looks good, but why have non-summary metric tests changed?

lahsivjar:
> Looks good, but why have non-summary metric tests changed?

Sorry, I should have created a separate PR for that. We were earlier ignoring the timestamp comparison with `pmetrictest.IgnoreTimestamp()`. I am still not sure why and I didn't look into it much but the `golden` package seems to behave weirdly with comparing less than 1 s timestamps - I found this while working on the PR and updated the timestamps to min 1sec while removing the ignore timestamp check. Let me know if it is a hindrance in the review process, I will create a separate PR for that.

processor/lsmintervalprocessor/processor_test.go
lahsivjar: 🤦 , thanks!

processor/lsmintervalprocessor/config.go
lahsivjar, line 34:
```suggestion
	// PassThrough is a configuration that determines whether summary
	// metrics should be passed through as they are or aggregated. This
	// is because they lead to lossy aggregations.
```

axw (approved):

PR #148, opened 2024-09-28
Allow resource attribute filtering and adding ephemeral IDs as resource attribute

lahsivjar:
Prior to this PR, resource attribute filtering for aggregated metrics was supposed to be done in the aggregation data pipeline using some other processor. This would have made maintaining the logic complicated and we would also have difficulty maintaining metric identity since we would need to drop a lot of resource attributes for our aggregated metrics. The current PR adds all this into spanmetricsv2 connector. The idea is that the metrics produced by the connector could be modified as required, including for resource attributes. Since modifying resource attributes could affect metric identity and single-writer principle, the PR also introduces another attribute which if enabled will add an ephemeral ID as resource attribute to the produced metric.

With all the above, the PR allows to deal with stripping resource attributes as well as dealing with its consequences with a single component. It does affect the possibility to upstream this but I think the changes could be reasoned with the community and introduced behind feature flags.

Closes: https://github.com/elastic/opentelemetry-dev/issues/387


connector/spanmetricsconnectorv2/internal/aggregator/aggregator.go
```diff
@@ -27,6 +27,11 @@ import (
 	"go.opentelemetry.io/collector/pdata/pmetric"
 )
 
+const (
+	scopeName            = "otelcol/spanmetricsconnectorv2"
```
carsonip, line 31: non-blocker: do we want a longer scope name like the hostmetrics one? Or is it better to keep "elastic repo" out of the picture?

carsonip (approved):
lgtm, a few questions

connector/spanmetricsconnectorv2/internal/aggregator/aggregator.go
lahsivjar: Yup, that will be a followup - didn't want to mix PRs.

connector/spanmetricsconnectorv2/internal/aggregator/aggregator.go
```diff
@@ -59,131 +72,124 @@ func NewAggregator() *Aggregator {
...
-			a.timestamp,
-			destExpHist,
-			destExplicitHist,
-			destSummary,
-			destCountersSum,
-			destCountersCount,
-		)
+// Finalize finalizezs the aggregations performed by the aggregator so far into
+// the pmetric.Metrics used to create this instance of the aggregator. The current
+// design of aggregator is not meant to be used once and should not be used after
```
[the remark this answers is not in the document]
lahsivjar: Result of brain<>finger mis-coordination 😅 . Fixed this.

carsonip (approved):

PR #150, opened 2024-10-01
test: add testdata to help visualize metrics

rogercoll:


gizas:
Nice !!!!

There are two main things that should be checked: -

1. Setting the scope.Name to https://github.com/elastic/opentelemetry-lib/blob/main/remappers/kubernetesmetrics/k8smetrics.go#L31
2. 
3. And setting timestamp values to the Metric because of the checks we do here: https://github.com/elastic/opentelemetry-lib/blob/main/remappers/kubernetesmetrics/kubelet.go#L41

gizas (approved):

rogercoll:
@gizas sgtm, thanks for the feeback. 

One unrelated issue that I see, based on the input/output files, is that if only one Otel metric is present, all `kubernetes` metrics will be added with value 0.

Issue #126, opened 2024-09-13
[processor/lsminterval] Support exponential histograms


lahsivjar:
Currently, the lsm-interval processor doesn't support merging exponential histograms.

PR #151, opened 2024-10-01
[processor/lsminterval] Add support for exponential histogram

lahsivjar:
Merges delta histogram with downscaling if required to merge histograms with 2 different scales.

Closes: https://github.com/elastic/opentelemetry-collector-components/issues/126


processor/lsmintervalprocessor/internal/merger/model.go
```diff
@@ -406,3 +417,126 @@ func mergeDeltaHistogramDP(from, to pmetric.HistogramDataPoint) {
...
+	}
+
+	to.SetSum(to.Sum() + from.Sum())
+	to.SetCount(to.Count() + from.Count())
+	// TODO (lahsivjar): Zero count depends on the value of zero threshold.
+	// If zero threshold are not the same then we can't merge them directly,
+	// instead, we will need to expand the zero threshold to be inclusive
+	// of the smaller zero count. If a zero threshold encroaches a bucket,
+	// then we will need to expand the zero count till the end of the bucket.
+	to.SetZeroCount(to.ZeroCount() + from.ZeroCount())
```
lahsivjar, line 439: [**For reviewers**] Will do this in a followup.

processor/lsmintervalprocessor/internal/merger/model.go
```diff
@@ -406,3 +417,126 @@ func mergeDeltaHistogramDP(from, to pmetric.HistogramDataPoint) {
...
+	scaleDownHistogramBuckets(hist.Positive(), diff)
+	scaleDownHistogramBuckets(hist.Negative(), diff)
+}
+
+// scaleDownHistogramBuckets scales down the histogram buckets by collapsing 2^(scaleDiff)
+// number of buckets into one bucket while handling the offset as required.
+// TODO (lahsivjar): Currently we don't trim the trailing zero entries from the buckets.
+// This is because OTel doesn't expose the data-model to efficiently slice the buckets,
+// instead, trimming trailing zero entries will require reallocation of the whole slice.
+// Is it worth it? (See test cases)
```
lahsivjar, line 500: [**For reviewers**] I don't think this is a big issue (keeping trailing zeroes) but let me know if there are any concerns.

processor/lsmintervalprocessor/internal/merger/model.go
```diff
@@ -406,3 +417,125 @@ func mergeDeltaHistogramDP(from, to pmetric.HistogramDataPoint) {
 		toCounts.SetAt(i, fromCounts.At(i)+toCounts.At(i))
 	}
 }
+
+func mergeDeltaExponentialHistogramDP(from, to pmetric.ExponentialHistogramDataPoint) {
+	if to.Count() == 0 {
```
lahsivjar, line 422: We should short-circuit these 2 conditions by just copying the whole DO. I haven't done this because I wanted to confirm if zero count is included in count but then it slipped my mind. I will fix this

processor/lsmintervalprocessor/internal/merger/model.go
axw: FYI there's logic you could copy at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/deltatocumulativeprocessor/internal/data/expo/zero.go

I wonder if we should just vendor and use that code, under the assumption that lsmintervalprocessor can be moved to opentelemetry-collector-contrib. At that point we could create a common (internal?) package for merging exponential histograms -- or maybe metrics in general.

processor/lsmintervalprocessor/internal/merger/model.go
```diff
@@ -406,3 +417,125 @@ func mergeDeltaHistogramDP(from, to pmetric.HistogramDataPoint) {
...
+	}
+	return b
+}
+
+func max[V constraints.Integer | constraints.Float](a, b V) V {
+	if a > b {
+		return a
+	}
+	return b
+}
```
axw, line 541: Nit: use `cmp.Ordered` and `cmp.Less` instead of the experimental constraints package?

processor/lsmintervalprocessor/internal/merger/model.go
axw: I wouldn't worry about it for now.

axw (commented):
Before I dive into the algorithms: is there a good reason why we shouldn't just vendor/copy a bit of code from deltatocumulativeprocessor temporarily? If not, I'd prefer we do that.

processor/lsmintervalprocessor/internal/merger/model.go
1pkg: Better yet use min and max builtin from the standard library https://pkg.go.dev/builtin#min, they are available since 1.21

axw: Yes! I thought there were functions, just couldn't remember where they were.

processor/lsmintervalprocessor/internal/merger/model.go
lahsivjar: > FYI there's logic you could copy

Ah, thanks! This would be helpful.

> I wonder if we should just vendor and use that code

I don't think we can. Looked into the code briefly and it only merges buckets of same scale: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a1915508175a498b731d17c951d776761ce8e824/processor/deltatocumulativeprocessor/internal/data/expo/merge.go#L10-L12

Maybe I will try to upstream the changes in this PR to support merges across scale.

processor/lsmintervalprocessor/internal/merger/model.go
lahsivjar: > Better yet use min and max builtin from the standard library https://pkg.go.dev/builtin#min, they are available since 1.21

wth, I spent time trying to find these and failed 🤦

lahsivjar: Fixed, thanks for this!! I don't know why I remembered it being in some other package and not in standard library 🤦

processor/lsmintervalprocessor/internal/merger/model.go
lahsivjar: This is done!

lahsivjar:
> Before I dive into the algorithms: is there a good reason why we shouldn't just vendor/copy a bit of code from deltatocumulativeprocessor temporarily? If not, I'd prefer we do that.

@axw Replied at https://github.com/elastic/opentelemetry-collector-components/pull/151#discussion_r1783988351. TL;DR, that one only supports merges for same scale.

processor/lsmintervalprocessor/internal/merger/model.go
lahsivjar: 👍 For posterity (and for me as well when I read this back), this is easy to do by just doing right-shift on the effective len and reslicing. The only reason to not do is due to the extra reallocation involved. Sample code:

```
	end := (hist.BucketCounts().Len() + int(hist.Offset()) - 1) >> scaleDiff // using offset before scaledown
	hist.BucketCounts().FromRaw(hist.BucketCounts().AsRaw()[:(end + 1)])
```

processor/lsmintervalprocessor/internal/merger/model.go
lahsivjar: > Maybe I will try to upstream the changes in this PR to support merges across scale.

Just realized that it doesn't make sense for deltatocumulative to support multiple scales. Maybe we can create a common exphistogram package in future when we try to upstream this.

axw: @lahsivjar it's true that the Merge function only works with buckets of the same scale, but there's also code to downscale: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a1915508175a498b731d17c951d776761ce8e824/processor/deltatocumulativeprocessor/internal/data/expo/scale.go#L42

They're all put together here: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a1915508175a498b731d17c951d776761ce8e824/processor/deltatocumulativeprocessor/internal/data/add.go#L68-L110

lahsivjar: Hmm, the logic is very similar but I am a bit reluctant since we will have to vendor the code and build `Merge` on top of it.

lahsivjar: I think the collapse logic is a bit complicated upstream in terms of how it is written - nothing major. It seems I am getting attached to my code 😅 . It might be better to just vendor. Let me update!

processor/lsmintervalprocessor/internal/merger/model.go
```diff
@@ -406,3 +416,120 @@ func mergeDeltaHistogramDP(from, to pmetric.HistogramDataPoint) {
...
+
+	if to.Offset() > from.Offset() {
+		// Expand `to` to accomodate the merged buckets with offset as `from.Offset()`
+		newBucketsSize := max(fromSize, toSize) - int(from.Offset())
+		newBuckets := pcommon.NewUInt64Slice()
+		newBuckets.Append(make([]uint64, newBucketsSize)...)
+
+		// Copy `to` to the new bucket
+		for i := 0; i < toBuckets.Len(); i++ {
+			newBuckets.SetAt(int(to.Offset())+i, toBuckets.At(i))
```
carsonip, line 473: how will this work when both to and from offsets are negative?

processor/lsmintervalprocessor/internal/merger/model.go
```diff
@@ -406,3 +416,120 @@ func mergeDeltaHistogramDP(from, to pmetric.HistogramDataPoint) {
...
+
+func mergeExponentialHistogramBuckets(from, to pmetric.ExponentialHistogramDataPointBuckets) {
+	fromBuckets := from.BucketCounts()
+	toBuckets := to.BucketCounts()
+	fromSize := int(from.Offset()) + fromBuckets.Len()
+	toSize := int(to.Offset()) + toBuckets.Len()
+
+	if to.Offset() > from.Offset() {
+		// Expand `to` to accomodate the merged buckets with offset as `from.Offset()`
+		newBucketsSize := max(fromSize, toSize) - int(from.Offset())
```
carsonip, line 467: Q: I'm slightly confused by this line. Is the meaning of "size" here the same as fromSize and toSize, or does it mean bucket length? the most intuitive way to me to get the new number of buckets is max(fromBuckets.Len(),toBuckets.Len()) + abs(from.Offset()-to.Offset()). Is this the same?

processor/lsmintervalprocessor/internal/merger/model.go
lahsivjar: I don't think offsets can be negative, they are number of zeroes and negative buckets are separated. Anway, check this https://github.com/elastic/opentelemetry-collector-components/pull/151#discussion_r1783675061 (TL;DR, we are vendoring an upstream internal implementation)

processor/lsmintervalprocessor/internal/merger/model.go
lahsivjar: Size here is the bucket size after merging. If you include the offset then you can calculate the size using above formula.

processor/lsmintervalprocessor/internal/merger/model.go
lahsivjar: Updated the code to vendor the internal package. Will also update histogram to use that instead in a separate PR.

processor/lsmintervalprocessor/internal/merger/model.go
lahsivjar: For posterity, Carson is correct here. The offset cannot be taken directly to calculate indexes, they must be normalized first by taking the difference of max of the upper bounds and min of the lower bounds.

processor/lsmintervalprocessor/internal/data/datatest/equal_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	is.Equal(want, got)
+
+	// Output:
+	// equal_test.go:55: Negative().BucketCounts().AsRaw(): [1 1 2] != [4]
+	// equal_test.go:55: Negative().BucketCounts().Len(): 3 != 1
+	// equal_test.go:55: Positive().BucketCounts().AsRaw(): [1 1 2] != [4]
+	// equal_test.go:55: Positive().BucketCounts().Len(): 3 != 1
+	// equal_test.go:55: Scale(): 0 != 1
+}
```
lahsivjar, line 63: [**For reviewers**] This bit is changed from upstream (line number from 35 -> 55) because the test is a bit brittle. It parses these lines (which has line numbers) and compares it with the actual output 😞

processor/lsmintervalprocessor/internal/merger/model_test.go
```diff
@@ -46,3 +48,171 @@ func TestKeyOrdered(t *testing.T) {
 		before = after
 	}
 }
+
+func TestMergeDeltaExponentialHistogram(t *testing.T) {
```
axw, line 52: Do we need all of these test cases here? I would think we can defer to the testing of package expo, and just verify here that things are wired up correctly with a basic test.

axw (approved):
LGTM, thanks for the changes :)

processor/lsmintervalprocessor/internal/merger/model_test.go
lahsivjar: Yup, sound good. Let me remove that!

lahsivjar: Done!

PR #163, opened 2024-10-04
Use OTTL statement for value and count

lahsivjar:
The PR is based on the discussion at https://github.com/elastic/opentelemetry-collector-components/pull/160#discussion_r1787082903. It allows the connector to be much more generic and utilizes OTTL to get `count`, and `value` for generating metrics.


lahsivjar:
superseded by `signaltometricsconnector`

PR #169, opened 2024-10-07
Add signaltometrics connector to produce metrics from all signals

lahsivjar:
`signaltometrics` is built as a connector that is capable of producing metrics from all signal types. The component utilizes OTTL to allow configuring the data and the metric types used to generate the metrics.

Related to: https://github.com/elastic/opentelemetry-dev/issues/372


connector/signaltometrics/config/config.go
```diff
@@ -0,0 +1,337 @@
...
+		info.ensureDefaults()
+		c.Logs[i] = info
+	}
+	return nil
+}
+
+// MetricInfo for a data type
+type MetricInfo struct {
+	Name        string `mapstructure:"name"`
+	Description string `mapstructure:"description"`
```
axw, line 116: Should we have a configurable unit in here too?

connector/signaltometrics/config/config.go
```diff
@@ -0,0 +1,337 @@
...
+	// default supports a high-resolution histogram able to cover a
+	// long-tail latency distribution from 1ms to 100s with a relative
+	// error of less than 5%.
+	// Ref: https://opentelemetry.io/docs/specs/otel/metrics/sdk/#base2-exponential-bucket-histogram-aggregation
+	defaultExponentialHistogramMaxSize = 160
+)
+
+var defaultHistogramBuckets = []float64{
+	2, 4, 6, 8, 10, 50, 100, 200, 400, 800, 1000, 1400, 2000, 5000, 10_000, 15_000,
+}
```
axw, line 47: Do other processors or connectors have default explicit bounds buckets? Maybe we should make it mandatory to specify them?

connector/signaltometrics/config/config.go
```diff
@@ -0,0 +1,337 @@
...
+
+type Summary struct {
+	Count string `mapstructure:"count"`
+	Value string `mapstructure:"value"`
+}
+
+type Sum struct {
+	Double string `mapstructure:"double"`
+	Int    string `mapstructure:"int"`
```
axw, line 336: Why does Sum need an explicit Double/Int distinction, and others don't?

connector/signaltometrics/connector.go
```diff
@@ -0,0 +1,281 @@
...
+						continue
+					}
+					var filteredResAttrs pcommon.Map
+					if len(md.IncludeResourceAttributes) > 0 {
+						filteredResAttrs = getFilteredAttributes(resourceAttrs, md.IncludeResourceAttributes)
+					} else {
+						// Copy resource attrs to avoid mutating data
+						filteredResAttrs = pcommon.NewMap()
+						resourceAttrs.CopyTo(filteredResAttrs)
+					}
```
axw, line 89: I think we should find a way to do this once per ResourceSpans per metric definition

axw (commented):
Thanks, I think this is looking great.

So just to make sure I understand: summary and histogram metrics have a default count of either 1 (log records & datapoints) or adjusted count (spans)? And then users can override this if they wish. I think that is sensible, just double checking my understanding.

Then where we currently aggregate a total span/transaction duration & count, we would create two metric definitions: one that sums up the durations, and one that sums up the adjusted counts. I think for the latter we'll need to either expose a new OTTL path for the adjusted count, or add a new OTTL converter, as discussed on Slack.

connector/signaltometrics/connector.go
```diff
@@ -0,0 +1,281 @@
...
+	otTraceState := w3cTraceState.OTelValue()
+	if otTraceState == nil {
+		return 1
+	}
+	if len(otTraceState.TValue()) == 0 {
+		// For non-probabilistic sampler OR always sampling threshold, default to 1
+		return 1
+	}
+	// TODO (lahsivjar): Handle fractional adjusted count. One way to do this
+	// would be to scale the values in the histograms for some precision.
```
felixbarny, line 265: Another option would be to apply some kind of probabilistic rounding. For example, if the adjusted count is 2.3, you return 2 with a 60% chance and 3 with a 30% chance.

connector/signaltometrics/connector.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	if multiError != nil {
+		return multiError
+	}
+	aggregator.Finalize(sm.dpMetricDefs)
+	return sm.next.ConsumeMetrics(ctx, processedMetrics)
+}
+
+func (sm *signalToMetrics) ConsumeLogs(ctx context.Context, logs plog.Logs) error {
+	var multiError error
```
felixbarny, line 206: Here and in other places, maybe check if there are metric definitions first, and short-circuit if there aren't.

connector/signaltometrics/config/config.go
lahsivjar: The SDKs and the [spanmetrics connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/679374a8656d39b5a38515c7d5fdaf708a999e4c/connector/spanmetricsconnector/config.go#L22-L24) at least have defaults.

connector/signaltometrics/config/config.go
lahsivjar: Hmm, I removed the `unit` (it was present in `spanmetricsv2`) because, unit could be configured with the OTTL converters. Do you see any other use for them?

connector/signaltometrics/config/config.go
lahsivjar: Sum can record either integer values or double values, others (histograms and summaries) don't need this since they record double values and integer counts.

connector/signaltometrics/connector.go
lahsivjar: In a follow-up, I will be adding collector's service information as observer (for single-writer). This will be added based on metric definition meaning we will have to allocate the underlying array accordingly for as per configuration in metric definition. We can keep a copy of the resource attributes and avoid the necessary copy, but I am not sure caching filtered attributes at the current point would give us any benefits.  Does it sound good to revisit this on the PR for the single-writer?

connector/signaltometrics/connector.go
lahsivjar: Probabilistic rounding is a good idea. I will create an issue for handling this and add the ideas there.

connector/signaltometrics/config/config.go
felixbarny: Could we just have a single `value` configuration and do the right thing depending on what the OTTL returns?

connector/signaltometrics/config/config.go
felixbarny: > unit could be configured with the OTTL converters

Do you mean with something like `Milliseconds`? While this converts the value to ms, a user of the metric doesn't know what unit it is in. So I think it would be useful if the unit can be explicitly specified.

connector/signaltometrics/config/config.go
axw: :+1: sounds like we should leave it then, thanks

connector/signaltometrics/config/config.go
axw: Right, if we store the unit in Elasticsearch then the metric can be self-describing to the degree that Kibana (or some other client) could visualise it more meaningfully. e.g. if it's "By", then Kibana could autoscale the charts to bytes, megabytes, gigabytes, etc.

lahsivjar: Ah, I see. So will the unit be added as a dimension?

connector/signaltometrics/config/config.go
axw: Sounds good. If it turns out to be problematic, we could always add another property just for sums like `type: int|double` or something like that.

connector/signaltometrics/config/config.go
axw: See https://github.com/elastic/opentelemetry-dev/blob/main/docs/design-decisions/ingest/mapping.md#units

I think it would need to be considered a dimension since it would influence grouping.

lahsivjar: We could always add these dimensions using attributes, something like:

```
signalttometrics:
  attributes:
    - key: unit
       default_value: ms
```

A little bit hacky, but I am a bit reluctant to introduce this because it doesn't directly impact the data manipulation/metric generation and is a bit specific to our (Elastic's) requirement. We could add a unit config and make it a bit more generic with something like:

```
unit:
  enabled: true
  attribute_name: unit
```

but I am not sure overall if we need this.

axw: Unit is part of the structured data, it's not a free-form attribute and is not vendor-specific. See https://github.com/open-telemetry/opentelemetry-collector/blob/b06236cc794982916cc956f20828b3e18eb33264/pdata/pmetric/generated_metric.go#L72-L81

I was thinking something like:

```yaml
- name: http.trace.span.duration
      description: Span duration for HTTP spans as a exponential histogram metric
      unit: ms
      attributes:
        - key: http.response.status_code
      exponential_histogram:
        value: Milliseconds(end_time - start_time)
```

felixbarny: Unit is already mapped as a dimension in a top-level field. See https://github.com/elastic/elasticsearch/blob/6d6fc66e90da3eddcd9f37d93361b77a6596fed9/x-pack/plugin/otel-data/src/main/resources/component-templates/metrics-otel%40mappings.yaml#L14-L17

lahsivjar: > Unit is part of the structured data, it's not a free-form attribute and is not vendor-specific

Thanks for the link, I didn't know the OTel model allowed adding unit explicitly 🤦 . Will fix.

connector/signaltometrics/config/config.go
felixbarny: This leads me to another question: What happens if the OTTL expression returns an invalid value, such as a string that can't be coerced to a number? I think what should happen is that we ignore the event. This may happen when using a custom attribute that uses different types across different services.

lahsivjar: We will ignore but with an error in the logs.

connector/signaltometrics/testdata/traces/exponential_histograms/config.yaml
```diff
@@ -0,0 +1,32 @@
...
+        value: Milliseconds(end_time - start_time)
+    - name: with_custom_count
+      description: Spans with custom count OTTL expression as a exponential histogram metric
+      exponential_histogram:
+        value: Milliseconds(end_time - start_time)
+        count: "2" # count each span twice
+    - name: http.trace.span.duration
+      description: Span duration for HTTP spans as a exponential histogram metric
+      attributes:
+        - key: http.response.status_code
```
felixbarny, line 17: Checking my understanding of how this is working. This implicitly filters events that have a value for this attribute. Events that don't have a value are ignored and neither the count nor the value is used. But when a default_value is supplied, there's no filtering. Is my understanding correct?

connector/signaltometrics/config/config.go
lahsivjar: Fixed in https://github.com/elastic/opentelemetry-collector-components/pull/169/commits/cb1555a18b8e989632da93b7eff19d790771194a

connector/signaltometrics/config/config.go
lahsivjar: Fixed in https://github.com/elastic/opentelemetry-collector-components/pull/169/commits/44a0fe792f5c3a3da8637200cf49461b2856d1df

connector/signaltometrics/connector.go
lahsivjar: Fixed!

connector/signaltometrics/testdata/traces/exponential_histograms/config.yaml
lahsivjar: > But when a default_value is supplied, there's no filtering

There is still filtering but if the key is not found in the attributes then the default value is used. For example, for the following definition of an attribute `attr.foo`:

```
- key: attr.foo
   default_value: bar
```

and events as:

```
event1:
  attributes:
    - key: attr.foo
       value: foo1
  ...
event2:
  attributes:
    - key: attr.foo
       value: foo2
  ...
event3:
  attributes: []
```

There will be 3 datapoints:

```
dp1: # for event 1
  attributes:
    - key: attr.foo
       value: foo1
dp2: # for event 2
  attributes:
    - key: attr.foo
       value: foo2
dp3: # for event 3 with attr.foo set to the default value of `bar`
  attributes:
    - key: attr.foo
       value: bar
```

Examples are also available in `testdata/traces/{metric_type}/{config.yaml, output.yaml}` (the input file is `testdata/traces/traces.yaml`

connector/signaltometricsconnector/README.md
```diff
@@ -0,0 +1,3 @@
+# Signal to metrics connector
+
+WIP
```
lahsivjar, line 3: [**For reviewers**] I will update this in a follow-up after all the configuration changes are in.

connector/signaltometricsconnector/internal/aggregator/aggregator.go
```diff
@@ -0,0 +1,336 @@
...
+	// smLookup maps resourceID against scope metrics since the aggregator
+	// always produces a single scope.
+	smLookup    map[[16]byte]pmetric.ScopeMetrics
+	valueCounts map[model.MetricKey]map[[16]byte]map[[16]byte]*valueCountDP
+	sums        map[model.MetricKey]map[[16]byte]map[[16]byte]*sumDP
+	timestamp   time.Time
+}
+
+// NewAggregator creates a new instance of aggregator.
+func NewAggregator[K any](metrics pmetric.Metrics) *Aggregator[K] {
```
lahsivjar, line 48: [**For reviewers**] This is currently missing unit-tests and is only tested via the connector tests, I will add these as a follow-up.

connector/signaltometricsconnector/connector_test.go
```diff
@@ -0,0 +1,355 @@
...
+			require.NoError(t, connector.ConsumeTraces(ctx, inputTraces))
+			require.Len(t, next.AllMetrics(), 1)
+			assertAggregatedMetrics(t, expectedMetrics, next.AllMetrics()[0])
+		})
+	}
+}
+
+func TestConnectorWithMetrics(t *testing.T) {
+	testCases := []string{
+		"sum",
```
lahsivjar, line 79: [**For reviewers**] I haven't added tests for histograms for the metric to metric case. This is because, I couldn't think of a good example for these without implementing `conditions` (will be a follow-up). Use-cases like gauge to histograms OR histograms to exponential histograms will only make sense if we can add conditions on metric types or metric names. So, these will be added with the PR for conditions.

lahsivjar (commented):
@axw @felixbarny The PR is ready for review.

connector/signaltometricsconnector/metadata.yaml
```diff
@@ -0,0 +1,30 @@
...
+          value: Microseconds(end_time - start_time)
+      - name: span.duration.exponential_histogram
+        description: "Span duration as exponential histogram"
+        exponential_histogram:
+          value: Microseconds(end_time - start_time)
+    logs:
+      - name: logrecord.count
+        description: "Count of log record"
+        sum:
+          value: Int("1") # Count each log record as 1
```
felixbarny, line 25: Is this equivalent to the default? I.e. can you just do `sum: {}`? Also, can you instead do `value: 1`, or does it need to Int converter?

connector/signaltometricsconnector/testdata/logs/exponential_histograms/config.yaml
```diff
@@ -0,0 +1,21 @@
+signaltometrics:
+  logs:
+    - name: total.logrecords.exphistogram
+      description: Logrecords as exponential histogram with log.duration from attributes
+      exponential_histogram:
+        count: Int(1)
+        value: Double(attributes["log.duration"])
```
felixbarny, line 7: Is the Double converter required, or can you also provide `attributes["log.duration"]`?

axw (commented):
Beautiful! Mostly LGTM with minor comments. I do think we need a plan for start timestamp - have you already thought about what we should do there?

connector/signaltometrics/connector.go
lahsivjar: Created an issue: https://github.com/elastic/opentelemetry-collector-components/issues/170

connector/signaltometricsconnector/config/config.go
```diff
@@ -0,0 +1,318 @@
...
+	// IncludeResourceAttributes is a list of resource attributes that
+	// needs to be included in the generated metric. If no resource
+	// attribute is included in the list then all attributes are included.
+	// Note that configuring this setting might cause the produced metric
+	// to lose its identity or cause identity conflict.
+	IncludeResourceAttributes []Attribute           `mapstructure:"include_resource_attributes"`
+	Attributes                []Attribute           `mapstructure:"attributes"`
+	Histogram                 *Histogram            `mapstructure:"histogram"`
+	ExponentialHistogram      *ExponentialHistogram `mapstructure:"exponential_histogram"`
+	Summary                   *Summary              `mapstructure:"summary"`
```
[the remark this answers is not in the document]
lahsivjar: 💯 , I agree. Will remove it.

.github/dependabot.yml
```diff
@@ -62,6 +62,21 @@ updates:
       - dependency-name: "*"
         update-types: ["version-update:semver-major"]
 
+  - package-ecosystem: "gomod"
+    directory: "/connector/signaltometricsconnector"
```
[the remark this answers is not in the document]
lahsivjar: The dependabot.yml links to https://github.com/dependabot/dependabot-core/issues/2178 issue which has been fixed now so probably we can use wildcards. I have created an issue for this: https://github.com/elastic/opentelemetry-collector-components/issues/171

connector/signaltometricsconnector/config/config.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+	for i, info := range c.Logs {
+		info.ensureDefaults()
+		c.Logs[i] = info
+	}
+	return nil
+}
+
+// MetricInfo for a data type
+type MetricInfo struct {
```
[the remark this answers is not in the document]
lahsivjar: I use `MetricDef` for the parsed config (in the `model/*`), maybe we can use `MetricConfig` here? WDYT?

connector/signaltometricsconnector/internal/ottlget/ottlget.go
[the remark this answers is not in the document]
lahsivjar: I don't think so because set editor doesn't return anything: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/985dddba71e26212e7e0535965fbec205135958f/pkg/ottl/ottlfuncs/func_set.go#L36-L46

connector/signaltometrics/config/config.go
lahsivjar: I will create an issue for this.

connector/signaltometricsconnector/connector.go
```diff
@@ -0,0 +1,306 @@
...
+			// No signals were consumed so return an error
+			return fmt.Errorf("failed to consume signal: %w", err)
+		}
+		// At least some signals were partially consumed, so log the error
+		// and pass the processed metrics to the next consumer.
+		sm.logger.Warn(
+			"failed to consume all signals, some signals were partially processed",
+			zap.Error(err),
+			zap.Int("successful_data_points", dpCount),
+		)
```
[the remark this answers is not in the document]
lahsivjar: Good point, I think it should be.

connector/signaltometricsconnector/connector_test.go
```diff
@@ -0,0 +1,355 @@
...
+				},
+			},
+			Attributes: []config.Attribute{
+				{
+					Key: "http.response.status_code",
+				},
+			},
+			Histogram: &config.Histogram{
+				Buckets: []float64{2, 4, 6, 8, 10, 50, 100, 200, 400, 800, 1000, 1400, 2000, 5000, 10_000, 15_000},
+				Value:   "Double(1.4)",
```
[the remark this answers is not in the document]
lahsivjar: We can, but we can't use it directly as float (like 1.4) in the yaml file and we have to cast it to string -- because parser accepts string. I was debugging something and left all the `Double` and `Int` lying around 🤦 . Will fix it.

connector/signaltometricsconnector/internal/aggregator/aggregator.go
```diff
@@ -0,0 +1,336 @@
...
+			ctx, tCtx,
+			md.Summary.Value,
+			md.Summary.Count,
+			defaultCount,
+		)
+		if err != nil {
+			return err
+		}
+		return a.aggregateValueCount(md, resAttrs, srcAttrs, val, count)
+	case md.Sum != nil && md.Sum.Value != nil:
```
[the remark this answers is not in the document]
lahsivjar: Remanants of all the changes in config validation. Will fix, thanks.

connector/signaltometricsconnector/internal/aggregator/aggregator.go
```diff
@@ -0,0 +1,336 @@
...
+			metrics := a.smLookup[resID].Metrics()
+			destMetric := metrics.AppendEmpty()
+			destMetric.SetName(md.Key.Name)
+			destMetric.SetDescription(md.Key.Description)
+			destMetric.SetUnit(md.Unit)
+			destCounter := destMetric.SetEmptySum()
+			destCounter.SetAggregationTemporality(pmetric.AggregationTemporalityDelta)
+			destCounter.DataPoints().EnsureCapacity(len(dpMap))
+			for _, dp := range dpMap {
+				// TODO (lahsivjar): Before merge: differentiate float vs int
```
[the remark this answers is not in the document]
lahsivjar: Ah, forgot to remove that TODO.

connector/signaltometricsconnector/internal/aggregator/exphistogram.go
```diff
@@ -0,0 +1,77 @@
...
+	dp.attrs.CopyTo(dest.Attributes())
+	dest.SetZeroCount(dp.data.ZeroCount())
+	dest.SetScale(dp.data.Scale())
+	dest.SetCount(dp.data.Count())
+	dest.SetSum(dp.data.Sum())
+	if dp.data.Count() > 0 {
+		dest.SetMin(dp.data.Min())
+		dest.SetMax(dp.data.Max())
+	}
+	// TODO determine appropriate start time
```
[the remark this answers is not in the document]
lahsivjar: Yeah, a few basic ideas. Created https://github.com/elastic/opentelemetry-collector-components/issues/172, will add details there.

connector/signaltometricsconnector/metadata.yaml
lahsivjar: I didn't add the default behavior for `sum` metric so far. With histograms, it is implicit what count would be but for sum metric, it didn't feel right to assume the behavior. No strong objection to having the defaults though.

> Also, can you instead do value: 1, or does it need to Int converter?

We can do `value: "1"` as it has to be a string but we don't need the int converter - that is remnant from my debugging something. I will simplify these.

connector/signaltometricsconnector/testdata/logs/exponential_histograms/config.yaml
lahsivjar: Not required, will simplify.

connector/signaltometricsconnector/config/config.go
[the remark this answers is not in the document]
lahsivjar: Done in https://github.com/elastic/opentelemetry-collector-components/pull/169/commits/e5dc6d5a654575aa3391600c039f19e6e42afb36

connector/signaltometrics/config/config.go
lahsivjar: Created an issue for error mode: https://github.com/elastic/opentelemetry-collector-components/issues/173

I will prioritize the follow-up PR for this.

lahsivjar:
@axw @felixbarny Addressed the concerns, for some of them I have created issues. Should be ready for another pass.

connector/signaltometricsconnector/metadata.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - name: span.duration.histogram
+        description: "Span duration as histogram"
+        unit: ms
+        histogram:
+          value: Microseconds(end_time - start_time)
+      - name: span.duration.exponential_histogram
+        description: "Span duration as exponential histogram"
+        unit: ms
+        exponential_histogram:
+          value: Microseconds(end_time - start_time)
```
felixbarny, line 22:
```suggestion
      - name: span.duration.histogram
        description: "Span duration as histogram"
        unit: micros
        histogram:
          value: Microseconds(end_time - start_time)
      - name: span.duration.exponential_histogram
        description: "Span duration as exponential histogram"
        unit: micros
        exponential_histogram:
          value: Microseconds(end_time - start_time)
```

felixbarny (approved):
LGTM! Just a small nit about using the appropriate unit in the test case.

connector/signaltometricsconnector/internal/ottlget/ottlget.go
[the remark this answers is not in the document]
felixbarny: Maybe we could set the value in [`cache["value"]`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/contexts/ottldatapoint/README.md#paths) and retrieve it from there? Having said that, I don't have a preference regarding introducing `get` vs using `set`.

connector/signaltometricsconnector/metadata.yaml
lahsivjar: Ah, thanks for catching this. Should it be `micros` or `us`? Do we have the units defined somewhere?

connector/signaltometricsconnector/internal/ottlget/ottlget.go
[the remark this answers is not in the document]
lahsivjar: > Maybe we could set the value in [cache["value"]](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/contexts/ottldatapoint/README.md#paths) and retrieve it from there

Hmm, I am not quite sure how to populate the cache. Also, the definition says:

>  the value of the current transform context's temporary cache. cache can be used as a temporary placeholder for data during complex transformations

so we would need to manually populate the cache with all of the values that could be used in the OTTL expression - I am not sure if that would be a good idea.

> Having said that, I don't have a preference regarding introducing get vs using set.

I don't think we can use `set` since it doesn't return the value from the getter ([ref](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/985dddba71e26212e7e0535965fbec205135958f/pkg/ottl/ottlfuncs/func_set.go#L36-L46)).

connector/signaltometricsconnector/metadata.yaml
felixbarny: You're right, it should be `us`. Units are defined by UCUM. See also https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units.

lahsivjar: Fixed in https://github.com/elastic/opentelemetry-collector-components/pull/169/commits/fe5438de6c7c3431720d55edd301d5a28cc25bbe

connector/signaltometricsconnector/internal/ottlget/ottlget.go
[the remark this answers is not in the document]
axw: > I don't think so because set editor doesn't return anything: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/985dddba71e26212e7e0535965fbec205135958f/pkg/ottl/ottlfuncs/func_set.go#L36-L46

I had in mind we would just use the statement for parsing (until there's a way to parse an isolated value), and then evaluate the `SetArguments.Value`. But it seems there's no way to get to the concrete statement type, so that doesn't work.

Never mind then, we can simplify it later.

connector/signaltometricsconnector/config/config.go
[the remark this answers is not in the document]
axw: Does it matter if they have the same name? I think it's still a metric definition from the user's perspective. I wouldn't go with MetricConfig, since it's in the config package.

Anyway, not critical that we address this atm, we can come back to it.

axw (approved):
Love it, thank you!

Issue #159, opened 2024-10-03
[connector/signaltometrics] Standardize ephemeral resource ID based on telemetry resources


lahsivjar:
As part of the [TelemetrySettings](https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/telemetry.go#L38-L39), OTel provides `service.{name, namespace, instance.id}` for the current instance of the collector. These attributes are [standard](https://opentelemetry.io/docs/specs/semconv/attributes-registry/service/#service-attributes) which uniquely identifies a service and applies to OTel collector also. We should include these attributes instead of a random unique ID.

Since we also need the `service.*` attributes to signal the originating service for a metric, we need to put these attributes under another suffix, maybe `observer.service.{name, namespace, instance.id}`?

lahsivjar:
Related discussion upstream: https://github.com/open-telemetry/opentelemetry.io/issues/4368

PR #178, opened 2024-10-10
[connector/signaltometrics]Add otelcol instance info as res attribute

lahsivjar:
With `include_resource_attributes` white listining which
resource attributes to add as resource attributes to the
final metrics, single-writer is affected. Add the collectors
instance information restores single-writer.

Closes: https://github.com/elastic/opentelemetry-collector-components/issues/159

### DIFF
--- a/connector/signaltometricsconnector/config/config.go
+++ b/connector/signaltometricsconnector/config/config.go
@@ -140,6 +140,12 @@ type MetricInfo struct {
 	// Unit, if not-empty, will set the unit associated with the metric.
 	// See: https://github.com/open-telemetry/opentelemetry-collector/blob/b06236cc794982916cc956f20828b3e18eb33264/pdata/pmetric/generated_metric.go#L72-L81
 	Unit string `mapstructure:"unit"`
+	// CollectorInfoAsResourceAttributes (experimental) appends the
+	// collector instance information, retrieved from the telemetry
+	// settings, as resource attributes to the produced metric if set
+	// to true. This is important to ensure single-writer if resource
+	// attributes are whitelisted using `include_resource_attributes`.
+	CollectorInfoAsResourceAttributes bool `mapstructure:"collector_info_as_resource_attributes"`
 	// IncludeResourceAttributes is a list of resource attributes that
 	// needs to be included in the generated metric. If no resource
 	// attribute is included in the list then all attributes are included.
--- a/connector/signaltometricsconnector/config/config.go
+++ b/connector/signaltometricsconnector/config/config.go
@@ -144,7 +144,7 @@ type MetricInfo struct {
 	// collector instance information, retrieved from the telemetry
 	// settings, as resource attributes to the produced metric if set
 	// to true. This is important to ensure single-writer if resource
-	// attributes are whitelisted using `include_resource_attributes`.
+	// attributes are included using `include_resource_attributes`.
 	CollectorInfoAsResourceAttributes bool `mapstructure:"collector_info_as_resource_attributes"`
 	// IncludeResourceAttributes is a list of resource attributes that
 	// needs to be included in the generated metric. If no resource
--- a/connector/signaltometricsconnector/config/config.go
+++ b/connector/signaltometricsconnector/config/config.go
@@ -140,17 +140,14 @@ type MetricInfo struct {
 	// Unit, if not-empty, will set the unit associated with the metric.
 	// See: https://github.com/open-telemetry/opentelemetry-collector/blob/b06236cc794982916cc956f20828b3e18eb33264/pdata/pmetric/generated_metric.go#L72-L81
 	Unit string `mapstructure:"unit"`
-	// CollectorInfoAsResourceAttributes (experimental) appends the
-	// collector instance information, retrieved from the telemetry
-	// settings, as resource attributes to the produced metric if set
-	// to true. This is important to ensure single-writer if resource
-	// attributes are included using `include_resource_attributes`.
-	CollectorInfoAsResourceAttributes bool `mapstructure:"collector_info_as_resource_attributes"`
 	// IncludeResourceAttributes is a list of resource attributes that
 	// needs to be included in the generated metric. If no resource
 	// attribute is included in the list then all attributes are included.
-	// Note that configuring this setting might cause the produced metric
-	// to lose its identity or cause identity conflict.
+	// Metric data streams MUST obey single-writer. The component produces
+	// metrics from non-metric signals as well as alters resource attributes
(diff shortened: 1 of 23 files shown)


felixbarny:
Is the collector's instance.id guaranteed to be always set and unique, or does it depend on the user having to configure it? I'm also wondering if we should only add the instance.id and omit the service.name and service.namespace because they're not required to maintain single-writer.

lahsivjar:
> Is the collector's instance.id guaranteed to be always set and unique, or does it depend on the user having to configure it?

I haven't dig deeper into where/how it is set but it doesn't require users to configure it. Testing on otel collector runs, even directly running the collectors built using OCB without any extra configs return all 3 of the values set.

> I'm also wondering if we should only add the instance.id and omit the service.name and service.namespace because they're not required to maintain single-writer.

As per [the docs](https://opentelemetry.io/docs/specs/semconv/attributes-registry/service/#service-attributes):

> [1]: MUST be unique for each instance of the same service.namespace,service.name pair (in other words service.namespace,service.name,service.instance.id triplet MUST be globally unique). The ID helps to distinguish instances of the same service that exist at the same time (e.g. instances of a horizontally scaled service).

So, I think we should keep all 3 of them.

lahsivjar:
> I haven't dig deeper into where/how it is set but it doesn't require users to configure it.

Did a bit of digging to find the [code](https://github.com/open-telemetry/opentelemetry-collector/blob/82117770f77ad2e2b2eaabd8e372e7029fc778f4/service/internal/resource/config.go#L17-L41), it could be specified by the users but if it not then the instance ID is autogenerated and the service.name is taken from the build information. So we should be good here.

connector/signaltometricsconnector/testdata/logs/exponential_histograms/config.yaml
```diff
@@ -5,6 +5,14 @@ signaltometrics:
       exponential_histogram:
         count: "1"
         value: attributes["log.duration"]
+    - name: total.logrecords.resource.foo.exphistogram
+      description: Logrecords with resource attribute foo as exponential histogram with log.duration from attributes
+      collector_info_as_resource_attributes: true
+      include_resource_attributes:
```
felixbarny, line 11: I'm wondering whether we should implicitly add the collector info if `include_resource_attributes` is used. Maybe with an opt-out. But by default, it seems that using `include_resource_attributes`  always bears the risk of violating single-writer.

lahsivjar: I deliberated on that, and I chose the current approach because it seemed a bit less confusing and magical. However, I do agree that include_resource_attributes has a high risk of violating single-writer.

To avoid confusion, we can do something like, always (include_resource_attributes present or not) add the collector info as resource attribute and give an opt-out option to the users?

connector/signaltometricsconnector/factory.go
```diff
@@ -34,6 +34,8 @@ import (
 	"github.com/elastic/opentelemetry-collector-components/connector/signaltometricsconnector/internal/model"
 )
 
+const collectorInstanceResourceAttributePrefix = "observer"
```
lahsivjar, line 37: I think this should also be configurable since there is no convention/standard around such use-cases where single writer is violated due to removing resource attribute and then restored by adding otel collector's instance information.

connector/signaltometricsconnector/internal/model/model.go
```diff
@@ -178,6 +180,71 @@ func (md *MetricDef[K]) FromMetricInfo(
...
+func (md *MetricDef[K]) FilterResourceAttributes(
+	attrs pcommon.Map,
+	collectorInfo *CollectorInstanceInfo,
+) pcommon.Map {
+	if len(md.IncludeResourceAttributes) == 0 {
+		// No resource attributes are whitelisted, return all as a copy.
+		// Copy is performed to avoid mutating the data.
+		m := pcommon.NewMap()
+		attrs.CopyTo(m)
+		return m
```
axw, line 197: Shouldn't we copy the collector info to the map in this case also?

connector/signaltometricsconnector/config/config.go
```diff
@@ -140,6 +140,12 @@ type MetricInfo struct {
 	// Unit, if not-empty, will set the unit associated with the metric.
 	// See: https://github.com/open-telemetry/opentelemetry-collector/blob/b06236cc794982916cc956f20828b3e18eb33264/pdata/pmetric/generated_metric.go#L72-L81
 	Unit string `mapstructure:"unit"`
+	// CollectorInfoAsResourceAttributes (experimental) appends the
+	// collector instance information, retrieved from the telemetry
+	// settings, as resource attributes to the produced metric if set
+	// to true. This is important to ensure single-writer if resource
+	// attributes are whitelisted using `include_resource_attributes`.
```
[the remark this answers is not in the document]
axw: Not sure that it should be configurable though, seems to me it shouldn't

connector/signaltometricsconnector/testdata/logs/exponential_histograms/config.yaml
felixbarny: > IIANM even if you included all resource attributes (i.e. include_resource_attributes is left unconfigured) then you could still violate the single-writer principle.

Why is that? Is it because signals other than metrics don't require to have a single-writer, so if you create metrics from these signals, you can't assume a single writer either?

axw: As discussed live, yes, because single-writer is only relevant for metrics, and a log record may have non-unique resource attributes. @lahsivjar please correct me if I'm wrong.

axw: ++ unless @felixbarny has objections let's remove the config and always add it

felixbarny: sgtm

connector/signaltometricsconnector/factory.go
lahsivjar: Updated!

connector/signaltometricsconnector/testdata/logs/exponential_histograms/config.yaml
lahsivjar: Updated to always add the collector info as resource attribute.

connector/signaltometricsconnector/internal/model/model.go
lahsivjar: Fixed.

axw (approved):
LGTM!

PR #180, opened 2024-10-10
[connector/signaltometrics]AdjustedCount as span converter

lahsivjar:
The PR adds `AdjustedCount` as an OTTL converter function, it only works for spans - similar to [IsRootSpan](https://pkg.go.dev/github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlfuncs#readme-isrootspan)) upstream.


connector/signaltometricsconnector/internal/trace/adjustedcount.go
```diff
@@ -0,0 +1,45 @@
...
+	otTraceState := w3cTraceState.OTelValue()
+	if otTraceState == nil {
+		return 1
+	}
+	if len(otTraceState.TValue()) == 0 {
+		// For non-probabilistic sampler OR always sampling threshold, default to 1
+		return 1
+	}
+	// TODO (lahsivjar): Handle fractional adjusted count. One way to do this
+	// would be to scale the values in the histograms for some precision.
```
felixbarny, line 43: In the context of this being called from a span converter now, I think the approach to scale the histograms isn't really feasible.

I wonder if the adjusted count should return a floating-point number so that users can decide what to do with it. But that makes it harder on the user of this converter and it means that we'd probably need something like a `ProbabilisticRound()` converter at some point. Maybe it makes sense to pull forward the probabilistic rounding so that this returns an integer, which is easer to work with?

felixbarny (commented):
Maybe we should work on upstreaming this even before the signalmetricsconnector. I feel a little uneasy about providing an Elastic-specific flavor of OTTL. But I think the approach makes sense and it's also scoped to just the signalmetricsconnector, not general OTTL.

connector/signaltometricsconnector/internal/trace/adjustedcount.go
lahsivjar: Hmm good point on the return type of floating-point. I am not sure if I fully understand all the concerns.

> I wonder if the adjusted count should return a floating-point number so that users can decide what to do with it.

This sounds like a good idea to me, I didn't quite get why it would get harder for the user of this converter as they could do one of the following:

1. `Int(AdjustedCount())` - if the want int
2. `Double(AdustedCount())` - if they want float
3. `ProbabilisticRound(AdustedCount())` - if they want rouding

felixbarny: > I didn't quite get why it would get harder for the user of this converter

It's just that they need to know how to deal with the floating-point number and they may not know what's be best strategy in which situation. Also, there are more things that can go wrong. For example, when just using `AdustedCount()` when an integer is expected. We'll also need to try to upstream both `AdjustedCount()` and `ProbabilisticRound()`.

Despite all the tradeoffs, it's probably still the best option.

lahsivjar: > It's just that they need to know how to deal with the floating-point number and they may not know what's be best strategy in which situation

Internally, we would have to cast the float64 to int to be able to use it as a count for histogram. So default usage of `AdustedCount()` should also work as expected.

Also, the [upstream package](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/47118f27b17fb275eddf2725a4a373939fbce2a1/pkg/sampling/oteltracestate.go#L182) we use returns `float64` which should make the behavior more consistent.

axw (commented):
Looks great. I like the ProbabilisticRound + float result suggestion, seems the best option.

felixbarny (approved):

lahsivjar:
@axw @felixbarny Updated the code for `AdjustedCount` to return float.

axw (approved):

PR #191, opened 2024-10-22
Initialise receiver/elasticapmreceiver

axw:
Create a skeleton for the `elasticapm` receiver, for receiving data from Elastic APM/RUM agents.


receiver/elasticapmreceiver/factory.go
```diff
@@ -0,0 +1,110 @@
...
+		return newElasticAPMReceiver(oCfg, set)
+	})
+	if err != nil {
+		return nil, err
+	}
+	r.Unwrap().nextTraces = consumer
+	return r, nil
+}
+
+var receivers = sharedcomponent.NewMap[*Config, *elasticAPMReceiver]()
```
lahsivjar, line 110: It is so bothersome that some important packages in contrib are internal causing us to duplicate things 😞

AlexanderWert:
cc @rogercoll

receiver/elasticapmreceiver/internal/sharedcomponent/sharedcomponent.go
```diff
@@ -0,0 +1,168 @@
...
+
+func NewMap[K comparable, V component.Component]() *Map[K, V] {
+	return &Map[K, V]{
+		components: map[K]*Component[V]{},
+	}
+}
+
+// Map keeps reference of all created instances for a given shared key such as a component configuration.
+type Map[K comparable, V component.Component] struct {
+	lock       sync.Mutex
```
lahsivjar, line 26: Hmm, I am not sure if we require a mutex here. Based on my understanding, the component initialization is done sequentially from the same goroutine. Do you have a reason for adding this?

The [upstream shared component](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/sharedcomponent/sharedcomponent.go) also doesn't do this. Also, should we just copy that internal package?

receiver/elasticapmreceiver/factory.go
axw: Indeed :(

receiver/elasticapmreceiver/internal/sharedcomponent/sharedcomponent.go
axw: There's actually two implementations, I copied the one from the core collector repo: https://github.com/open-telemetry/opentelemetry-collector/tree/main/internal/sharedcomponent

lahsivjar: Ah, TIL. Interestingly, the mutex was added as part of the effort to expose the sharedcomponent API: https://github.com/open-telemetry/opentelemetry-collector/pull/9157

rogercoll (approved):
(some CI checks failing, should be fixed by running `make license-update`)

Skeleton looks great to me. I think it would be interesting to start with the following in a follow-up PR:

- Translators: move the APM Event to Otel pdata translations to a custom package. Similar to https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/translator/opencensus
  - Do you know if there is any repository/source where we can get json `modelpb.Batch` examples for unit testing?
- README.md with support matrix
- `testdata`: Integration test with the expected generated signals

receiver/elasticapmreceiver/generated_component_test.go
```diff
@@ -0,0 +1,83 @@
+// Code generated by mdatagen. DO NOT EDIT.
```
lahsivjar, line 18: Need to run the `make license-update` again after generation.

receiver/elasticapmreceiver/internal/sharedcomponent/sharedcomponent.go
```diff
@@ -0,0 +1,168 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package sharedcomponent exposes functionality for components
```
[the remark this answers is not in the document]
axw: :+1: done

receiver/elasticapmreceiver/receiver.go
```diff
@@ -0,0 +1,283 @@
...
+		Logger:       r.settings.Logger,
+		MaxEventSize: maxEventSize,
+		Semaphore:    sem,
+	})
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		statusCode := http.StatusAccepted
+		baseEvent := &modelpb.APMEvent{}
+		baseEvent.Event = &modelpb.Event{}
+		baseEvent.Event.Received = modelpb.FromTime(time.Now())
```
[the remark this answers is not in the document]
axw: Hmm, actually I don't recall, I don't think it's needed though. I copied it from elswhere. I've removed it for now.

receiver/elasticapmreceiver/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+type: elasticapm
```
lahsivjar, line 1:
```suggestion
type: elasticapm
scope_name: github.com/elastic/opentelemetry-collector-components/receiver/elasticapmreceiver
github_project: elastic/opentelemetry-collector-components
```

This will help mdatagen based auto generation code to use elastic's repository rather than the default upstream one.

receiver/elasticapmreceiver/receiver.go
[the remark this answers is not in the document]
lahsivjar: In APM data-model we have `event.received` field which basically tracks when a specific event was received by APM server.

lahsivjar (approved):
LGTM!

receiver/elasticapmreceiver/generated_component_test.go
lahsivjar: Ah, already done... reviewed the older version. Resolving.

lahsivjar:
@axw Also, looks like the `mdatagen` used to generate the code is old. You may need to update that. OTOH, I think there is something wrong with the tools setup of mdatagen in the repository... it seems to use the global one.

axw:
> @axw Also, looks like the mdatagen used to generate the code is old. You may need to update that in you local. OTOH, I think there is something wrong with the tools setup of mdatagen in the repository... it seems to use the global one.

Indeed, I've pushed a fix to Makefile.Common to add the tools dir to $PATH

receiver/elasticapmreceiver/receiver.go
```diff
@@ -0,0 +1,283 @@
...
+				s.SetName(span.GetName())
+			}
+		default:
+			return fmt.Errorf("unhandled event type %q", event.Type())
+		}
+	}
+	var errs []error
+	if numRecords := ld.LogRecordCount(); numRecords != 0 {
+		ctx := r.obsreport.StartLogsOp(ctx)
+		err := r.nextLogs.ConsumeLogs(ctx, ld)
```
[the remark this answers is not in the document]
axw: I don't think it's necessary. For the `Consume*` methods to be invoked, the relevant factory function (`create*Receiver`) must have been invoked; these functions set the "next*" consumer fields. This pattern is due to the receiver being a shared component, and matches how the otlpreceiver works, for example.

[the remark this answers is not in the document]
axw: Merging as is, feel free to propose a change in a followup

PR #200, opened 2024-10-23
ci: allow dependabot workflow checks

rogercoll:


lahsivjar (approved):
Not entirely sure why we need `always()` but since it was already there LGTM!

rogercoll:
Unrelated CI error

PR #212, opened 2024-10-24
github-actions: set contents: read permission at the top level

v1v:
does what it says in the tin


.github/workflows/checks.yml
```diff
@@ -10,6 +10,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
+permissions:
+  content: read
```
v1v, line 14:
```suggestion
  contents: read
```

rogercoll (approved):

PR #232, opened 2024-11-25
[processor/lsminterval] Add benchmarks

carsonip:
Add benchmarks to facilitate perf optimization work


processor/lsmintervalprocessor/processor_test.go
```diff
@@ -118,3 +120,72 @@ func TestAggregation(t *testing.T) {
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	for _, tc := range testCases {
+		config := &Config{
+			Intervals: []IntervalConfig{
+				{
+					Duration: time.Hour,
+					Statements: []string{
```
lahsivjar, line 147: This will utilize OTTL, let's separate this also as a variable in the benchmark function. We can do all the benchmarks with and without OTTL.

processor/lsmintervalprocessor/processor_test.go
```diff
@@ -118,3 +120,90 @@ func TestAggregation(t *testing.T) {
...
+	defer cancel()
+
+	for _, ottlStatements := range [][]string{
+		{},
+		{
+			`set(resource.attributes["custom_res_attr"], "res")`,
+			`set(instrumentation_scope.attributes["custom_scope_attr"], "scope")`,
+			`set(attributes["custom_dp_attr"], "dp")`,
+			`set(resource.attributes["dependent_attr"], Concat([attributes["aaa"], "dependent"], "-"))`,
+		},
```
lahsivjar, line 150: _nit_: I would rather split the benchmarks into 2, like `BenchmarkAggregation` and `BenchmarkAggregationWithOTTL` and refactor the test to use a helper method that would run all the test cases with the ottl statements as an input.

lahsivjar (approved):
A minor nit but LGTM! Thanks for adding this.

processor/lsmintervalprocessor/processor_test.go
carsonip: good call!

processor/lsmintervalprocessor/processor_test.go
```diff
@@ -76,6 +75,9 @@ func TestAggregation(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
+			ctx, cancel := context.WithCancel(context.Background())
```
carsonip, line 78: [To reviewer] Not sure why it was outside of the tc loop in the first place

lahsivjar (approved):
Thanks!

Issue #141, opened 2024-09-25
[processor/lsminterval] Handle overflow for metrics aggregations


lahsivjar:
Aggregations could be boundless with huge cardinality OR with buggy instrumentation. To protect against these, the aggregated metrics should have limits and overflow to specific buckets after those limits are breached. This would be similar to what is implemented in https://github.com/elastic/apm-aggregation

lahsivjar:
### Current overflow handling in APM Server using [apm-aggregation](https://github.com/elastic/apm-aggregation)

APM aggregation implements overflow handling logic based on a set of factors, each having their own threshold. These factors are configured in APM Server based on the available resources or hard-coded if the use-case is known. By default, APM Server sets all the limits based on the memory available using a linear scaling based on the formula `mem_available * constant_factor`. For example: for services, the constant factor is `1000`, i.e. 8GB APM-Server will have a limit on the maximum number of services as `8000` ([ref](https://github.com/elastic/apm-server/blob/1674e8e0a5ec3d3bd1251c39d6b723d6b2ab4968/internal/beater/beater.go#L196-L201)).

#### Maximum number of services

The number of unique services for the purpose of aggregated metrics is defined as the cardinality of the following set of labels (identified as service aggregation key):

- `timestamp` truncated based on the interval
- `service.name`
- `service.environment`
- `service.language.name`
- `agent.name`
- Global labels 

When the max service limit for an aggregation interval is reached then a new `service_summary` metric is produced with the service name set to `_other` and the timestamp set to the processing timestamp - service summary metrics in APM server doesn't record any numeric value. In addition, another metric with the name of `service_summary.aggregation.overflow_count` is produced that records service aggregation keys overflowed for the given interval.

This overflow service bucket acts like a catch-all for all overflow aggregated metrics i.e. in addition to the service count, the overflow services also record service transaction, transaction, and span metrics in their corresponding data types (histograms, aggregate_metric_double, or counters) with the following overflow identifiers

- Overflow service transaction is identified by `transaction.type` set to `_other` with `service.name` set to `_other`
- Overflow transaction is identified by `transaction.name` set to `_other` with `service.name` set to `_other`
- Overflow span is identified by `service.target.name` set to `_other` with `service.name` set to `_other`

#### Maximum number of service transactions

[...]

PR #235, opened 2024-11-30
[processor/lsminterval] Define cardinality limits and handle overflows

lahsivjar:
Related to: https://github.com/elastic/opentelemetry-collector-components/issues/141

For more details checkout https://github.com/elastic/opentelemetry-collector-components/issues/141#issuecomment-2508751134 comment.


axw (commented):
At a high level I think the approach looks good. As mentioned in a comment, I think we should have limits on all parts of the hierarchy, including metrics.

I ran the benchmarks, and it seems there's quite a drop in throughput.

processor/lsmintervalprocessor/config/config.go
```diff
@@ -39,7 +39,10 @@ type Config struct {
 	// TODO (lahsivjar): Make specifying interval easier. We can just
 	// optimize the timer to run on differnt times and remove any
 	// restriction on different interval configuration.
-	Intervals []IntervalConfig `mapstructure:"intervals"`
+	Intervals       []IntervalConfig `mapstructure:"intervals"`
+	ResourceLimits  LimitConfig      `mapstructure:"resource_limit"`
+	ScopeLimits     LimitConfig      `mapstructure:"scope_limit"`
+	DatapointLimits LimitConfig      `mapstructure:"datapoint_limit"`
```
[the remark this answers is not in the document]
lahsivjar: Yes, I will push it with a few more fixes in the next commit.

lahsivjar:
> I ran the benchmarks, and it seems there's quite a drop in throughput.

Yeah, there is a considerable overhead. This is what I am working on right now -- to improve performance and drop some of the unneeded complexity around attribute matching and filtering.

lahsivjar:
@axw I have addressed all of the comments other than one and I am marking this ready for review now. The performance consideration is still there and I have a few threads to chase down for improvements but I don't have a lot of hope from them, so, if you have any ideas (even rough) I will be more than happy to chase them down.

I have left adding metrics limits to a follow-up PR. I will also add a few more detailed overflow tests.

processor/lsmintervalprocessor/internal/merger/limits/tracker.go
```diff
@@ -0,0 +1,149 @@
...
+				))
+			}
+			return true
+		}
+		return false
+	})
+	if len(errs) > 0 {
+		return fmt.Errorf("failed to unmarshal tracker: %w", errors.Join(errs...))
+	}
+	return nil
```
axw, line 136: Should we check that we found the expected keys here?

processor/lsmintervalprocessor/config/config.go
```diff
@@ -63,6 +66,20 @@ type IntervalConfig struct {
...
+}
+
+type OverflowConfig struct {
+	Attributes []Attribute `mapstructure:"attributes"`
+}
+
+type Attribute struct {
+	Key   string `mapstructure:"key"`
+	Value any    `mapstructure:"value"`
+}
```
axw, line 81: Could we add some doc comments to the types & fields please?

processor/lsmintervalprocessor/internal/merger/merger.go
```diff
@@ -21,35 +21,44 @@ import (
...
 
 	"github.com/cockroachdb/pebble"
+	"github.com/elastic/opentelemetry-collector-components/processor/lsmintervalprocessor/config"
 )
 
 var _ pebble.ValueMerger = (*Merger)(nil)
 
 type Merger struct {
-	current Value
+	current          Value
```
axw, line 30: Now that Merger methods have non-pointer receivers, they'll be copying the internals. Won't copying Value be pretty expensive?

axw (commented):
Preliminary comments, ran out of time today

processor/lsmintervalprocessor/internal/merger/merger.go
lahsivjar: > Won't copying Value be pretty expensive?

Hmm, not by much I think because most of the value is reference type. Even pmetric is kinda wrapper around the proto OTel data model references. Though, I didn't intend to change the merger to value receiver but only add the limit config explicitly 😅 I think this happened by mistake, I will fix this.

processor/lsmintervalprocessor/internal/merger/limits/tracker.go
lahsivjar: Good point, the prefix counter key must be present if the tracker was decoded so it should be an error if tracker was expected and not found. Will add this.

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+		scopeDPLimitCfg:  dpLimit,
+		source:           pmetric.NewMetrics(),
+		resourceLimits:   limits.NewTracker(resLimit.MaxCardinality),
+		resLookup:        make(map[identity.Resource]resourceMetrics),
+		scopeLookup:      make(map[identity.Scope]scopeMetrics),
+		metricLookup:     make(map[identity.Metric]metric),
+		numberLookup:     make(map[identity.Stream]numberDataPoint),
+		summaryLookup:    make(map[identity.Stream]summaryDataPoint),
+		histoLookup:      make(map[identity.Stream]histogramDataPoint),
+		expHistoLookup:   make(map[identity.Stream]exponentialHistogramDataPoint),
```
axw, line 97: Should we defer instantiating these maps until first use? Particularly for single batches, some of them may never be used.

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+// MarshalProto marshals the value into binary. Before marshaling the metric,
+// the overflow values are encoded as attributes with pre-defined keys.
+func (s *Value) MarshalProto() ([]byte, error) {
+	rms := s.source.ResourceMetrics()
+	if rms.Len() > 0 {
+		// Encode resource tracker at the 0th resource metrics
+		// TODO (lahsivjar): Is this safe? We don't ever remove
+		// resource metrics so it should be but best to check.
+		// Also, the limits checker should ensure max cardinality
+		// is greater than zero.
```
axw, line 110: Another approach we could take that does not involve piggybacking on attributes would be to first encode the non-overflowed ResourceMetrics, and then encode a tracker for each node in the tree in traversal order. i.e. one for the Resource, then for each Scope in that resource, etc.

That may increase the storage size slightly when there are no/few overflows, but would avoid having to iterate through the attributes when unmarshalling. If we were to do this, we would probably want to add some versioning to the format, e.g. so we can add another level of limits.

Did you consider this? Not sure if it's better at all, what you've done is fairly straightforward - I mostly worry about the performance cost of iterating through attributes when unmarshalling individual batches.

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+		if err := decorate(
+			overflowDP.Attributes(),
+			s.scopeDPLimitCfg.Overflow.Attributes,
+		); err != nil {
+			return pmetric.Metrics{}, fmt.Errorf("failed to finalize merged metric: %w", err)
+		}
+		overflowDP.SetIntValue(int64(sm.datapointsLimits.EstimateOverflow()))
+	}
+	// Remove any hanging resource or scope which failed to have any entries
+	// due to children reaching their limits.
```
axw, line 271: Seeing as the scope limit is per resource, and the metric/dp limit is per scope, when would we add an empty resource or scope?

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+		}
+		// Merge any overflow estimators for scope or datapoints. Note that here
+		// we assume that the limits for both metrics being merged are identical
+		// and thus if any of the metric has overflowed then the target metric
+		// for merge will definitely overflow. Thus, merging the estimators is
+		// safe and required to correctly estimate the total number of overflow.
+		// Note that overflow merging is not deterministic and the estimates are
+		// always estimates as it is possible that the oveflows have hashes which
+		// haven't overflowed in the source metric.
+		res := v.resLookup[resID]
```
axw, line 326: Should addResourceMetrics return resourceMetrics rather than just the identity? Then we don't need to do a separate lookup. Same for addScopeMetrics & the scopeLookup below

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+// addResourceMetrics adds a new resource metrics to the store while also
+// applying resource limiters. If a limit is configured and breached by
+// adding the provided resource metric, then, a new overflow resource
+// metric is created and returned.
+func (s *Value) addResourceMetrics(
+	otherRm pmetric.ResourceMetrics,
+) (identity.Resource, error) {
+	resID := identity.OfResource(otherRm.Resource())
+	if _, ok := s.resLookup[resID]; ok {
+		return resID, nil
```
axw, line 354: Is it possible that two ResourceMetrics have identical Resources, but different SchemaUrls? Do we need to consider this? Same for ScopeMetrics

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+	}
+	if s.resourceLimits.CheckOverflow(resID.Hash().Sum64()) {
+		// Overflow, get/prepare an overflow bucket
+		overflowResID, err := s.getOverflowResourceBucketID()
+		if err != nil {
+			return identity.Resource{}, err
+		}
+		if _, ok := s.resLookup[overflowResID]; !ok {
+			overflowRm := s.source.ResourceMetrics().AppendEmpty()
```
axw, line 363: Should we be setting a schema URL? Same for ScopeMetrics

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+	otherMetric pmetric.Metric,
+) identity.Metric {
+	metricID := identity.OfMetric(scopeID, otherMetric)
+	if _, ok := s.metricLookup[metricID]; ok {
+		return metricID
+	}
+	scope := s.scopeLookup[scopeID]
+
+	// Metrics doesn't have overflows (only datapoints have)
+	// Clone it *without* the datapoint data
```
axw, line 451: Can we add a TODO to add a limit for metrics?

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+		// Datapoints overflow detected. In this case no action has to be
+		// done at this point since data point overflow should create a new
+		// overflow metric of sum type recording the number of unique
+		// datapoints. This number will be recorded in the limit tracker
+		// and the metric will be populated on demand.
+		return pmetric.NumberDataPoint{}, false
+	}
+	dp := metric.Sum().DataPoints().AppendEmpty()
+	s.numberLookup[streamID] = dp
+	return dp, true
```
axw, line 507: Perhaps the caller could be simplified if we called `otherDP.CopyTo(dp)` here? Then instead of returning true here, we would return true only when the caller should update the returned data point. WDYT? (Same for other metric types of course.)

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+		merge(
+			m.ExponentialHistogram().DataPoints(),
+			metricID,
+			v.addExponentialHistogramDataPoint,
+			m.ExponentialHistogram().AggregationTemporality(),
+		)
+	}
+}
+
+func (s *Value) getOverflowResourceBucketID() (identity.Resource, error) {
```
axw, line 637: nit: getOverflowResourceIdentity?

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+	if err := decorate(
+		r.Attributes(),
+		s.resourceLimitCfg.Overflow.Attributes,
+	); err != nil {
+		return identity.Resource{}, fmt.Errorf("failed to create overflow bucket: %w", err)
+	}
+	return identity.OfResource(r), nil
+}
+
+func (s *Value) getOverflowScopeBucketID(
```
axw, line 648: nit: getOverflowScopeIdentity?

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,796 @@
...
+		toDP, ok := addDP(toMetricID, fromDP)
+		if toDP == zero {
+			// Overflow, discard the datapoint
+			continue
+		}
+		if ok {
+			// New data point is created so we can copy the old data directly
+			fromDP.CopyTo(toDP)
+			continue
+		}
```
axw, line 726: Comparing to zero feels like an unnecessarily expensive way to do this. See my other comment about returning true only when we should update. Then this could become

```go
if toDP, exists := addDP(toMetricID, fromDP); exists {
	switch fromDP := any(fromDP).(type) {
	case pmetric.NumberDataPoint:
		mergeDeltaSumDP(fromDP, any(toDP).(pmetric.NumberDataPoint))
	case pmetric.HistogramDataPoint:
		mergeDeltaHistogramDP(fromDP, any(toDP).(pmetric.HistogramDataPoint))
	case pmetric.ExponentialHistogramDataPoint:
		mergeDeltaExponentialHistogramDP(fromDP, any(toDP).(pmetric.ExponentialHistogramDataPoint))
	}
}
```

axw (commented):
Looks good overall, mostly minor comments/questions

lahsivjar:
Bringing this back to draft as I have a few good ideas for optimization and combined with @axw 's above comments I have a few more things to do here.

(I have also pushed a different way to encode limit trackers independent of the pmetric ds)

lahsivjar:
Update so far, I have been able to make some progress on optimizing the processor as a whole. I need to perform more tests  and more real-world correctness tests but with the changes so far below are the result as compared to main branch (TL;DR around 50% for allocations and runtime).

<details>
<summary> `BenchmarkAggregation` as compared to `main` branch </summary>

name                                            old time/op    new time/op    delta
Aggregation/sum_cumulative-10                     10.3µs ± 5%     5.3µs ± 2%  -48.24%  (p=0.008 n=5+5)
Aggregation/sum_delta-10                          10.4µs ± 2%     5.5µs ± 3%  -47.77%  (p=0.008 n=5+5)
Aggregation/histogram_cumulative-10               10.8µs ± 2%     5.8µs ± 4%  -45.96%  (p=0.008 n=5+5)
Aggregation/histogram_delta-10                    11.2µs ± 5%     6.1µs ± 8%  -45.48%  (p=0.008 n=5+5)
Aggregation/exphistogram_cumulative-10            12.0µs ± 4%     6.2µs ± 3%  -47.84%  (p=0.008 n=5+5)
Aggregation/exphistogram_delta-10                 11.8µs ± 3%     6.6µs ± 6%  -44.36%  (p=0.008 n=5+5)
Aggregation/summary_enabled-10                    10.9µs ± 3%     5.9µs ± 1%  -46.36%  (p=0.008 n=5+5)
Aggregation/summary_passthrough-10                1.35µs ± 2%    1.33µs ± 3%     ~     (p=0.341 n=5+5)
AggregationWithOTTL/sum_cumulative-10             12.5µs ± 3%     5.5µs ± 2%  -56.30%  (p=0.008 n=5+5)
AggregationWithOTTL/sum_delta-10                  12.8µs ± 2%     5.5µs ± 2%  -57.41%  (p=0.008 n=5+5)
AggregationWithOTTL/histogram_cumulative-10       13.1µs ± 2%     6.0µs ± 3%  -53.86%  (p=0.008 n=5+5)
AggregationWithOTTL/histogram_delta-10            13.4µs ± 2%     6.2µs ± 3%  -53.69%  (p=0.008 n=5+5)
AggregationWithOTTL/exphistogram_cumulative-10    13.3µs ± 3%     6.2µs ± 6%  -53.25%  (p=0.008 n=5+5)
AggregationWithOTTL/exphistogram_delta-10         13.4µs ± 2%     6.6µs ± 0%  -50.77%  (p=0.016 n=5+4)
AggregationWithOTTL/summary_enabled-10            12.7µs ± 1%     6.2µs ± 3%  -51.40%  (p=0.008 n=5+5)
AggregationWithOTTL/summary_passthrough-10        1.34µs ± 2%    1.36µs ± 7%     ~     (p=0.500 n=5+5)

name                                            old alloc/op   new alloc/op   delta
Aggregation/sum_cumulative-10                     17.5kB ± 1%     8.6kB ± 0%  -51.06%  (p=0.008 n=5+5)
Aggregation/sum_delta-10                          17.6kB ± 1%     8.6kB ± 2%  -51.17%  (p=0.008 n=5+5)
Aggregation/histogram_cumulative-10               19.5kB ± 1%    10.2kB ± 1%  -47.56%  (p=0.008 n=5+5)
Aggregation/histogram_delta-10                    19.4kB ± 1%    10.1kB ± 0%  -47.81%  (p=0.016 n=5+4)
Aggregation/exphistogram_cumulative-10            21.7kB ± 2%    10.1kB ± 0%  -53.39%  (p=0.008 n=5+5)
Aggregation/exphistogram_delta-10                 22.2kB ± 1%    10.6kB ± 2%  -52.34%  (p=0.008 n=5+5)
Aggregation/summary_enabled-10                    18.3kB ± 1%     8.8kB ± 0%  -51.79%  (p=0.008 n=5+5)
Aggregation/summary_passthrough-10                1.37kB ± 0%    1.32kB ± 0%   -3.99%  (p=0.016 n=4+5)
AggregationWithOTTL/sum_cumulative-10             20.2kB ± 1%     8.6kB ± 1%  -57.46%  (p=0.008 n=5+5)
AggregationWithOTTL/sum_delta-10                  20.3kB ± 1%     8.6kB ± 3%  -57.72%  (p=0.008 n=5+5)
AggregationWithOTTL/histogram_cumulative-10       22.1kB ± 3%    10.3kB ± 1%  -53.55%  (p=0.008 n=5+5)
AggregationWithOTTL/histogram_delta-10            22.1kB ± 2%    10.2kB ± 1%  -54.09%  (p=0.008 n=5+5)
AggregationWithOTTL/exphistogram_cumulative-10    23.0kB ± 4%    10.2kB ± 2%  -55.62%  (p=0.008 n=5+5)
AggregationWithOTTL/exphistogram_delta-10         23.0kB ± 1%    10.6kB ± 0%  -53.80%  (p=0.016 n=5+4)
AggregationWithOTTL/summary_enabled-10            20.8kB ± 2%     9.0kB ± 2%  -56.84%  (p=0.008 n=5+5)
AggregationWithOTTL/summary_passthrough-10        1.38kB ± 0%    1.32kB ± 0%   -4.22%  (p=0.008 n=5+5)

name                                            old allocs/op  new allocs/op  delta
Aggregation/sum_cumulative-10                        199 ± 1%       107 ± 0%  -46.23%  (p=0.016 n=5+4)
Aggregation/sum_delta-10                             200 ± 0%       111 ± 5%  -44.61%  (p=0.008 n=5+5)
Aggregation/histogram_cumulative-10                  212 ± 0%       121 ± 4%  -43.02%  (p=0.008 n=5+5)
Aggregation/histogram_delta-10                       214 ± 0%       125 ± 0%  -41.48%  (p=0.008 n=5+5)
Aggregation/exphistogram_cumulative-10               228 ± 0%       120 ± 1%  -47.60%  (p=0.016 n=4+5)
Aggregation/exphistogram_delta-10                    237 ± 0%       131 ± 6%  -44.81%  (p=0.008 n=5+5)
Aggregation/summary_enabled-10                       212 ± 0%       114 ± 1%  -46.23%  (p=0.016 n=4+5)
Aggregation/summary_passthrough-10                  37.0 ± 0%      33.0 ± 0%  -10.81%  (p=0.008 n=5+5)
AggregationWithOTTL/sum_cumulative-10                227 ± 0%       109 ± 3%  -51.89%  (p=0.008 n=5+5)
AggregationWithOTTL/sum_delta-10                     229 ± 0%       109 ± 7%  -52.40%  (p=0.008 n=5+5)
AggregationWithOTTL/histogram_cumulative-10          244 ± 0%       124 ± 0%  -49.10%  (p=0.008 n=5+5)
AggregationWithOTTL/histogram_delta-10               245 ± 0%       123 ± 3%  -49.96%  (p=0.016 n=4+5)
AggregationWithOTTL/exphistogram_cumulative-10       256 ± 0%       124 ± 4%  -51.72%  (p=0.008 n=5+5)
AggregationWithOTTL/exphistogram_delta-10            265 ± 0%       138 ± 0%  -47.92%  (p=0.029 n=4+4)
AggregationWithOTTL/summary_enabled-10               240 ± 0%       120 ± 6%  -50.17%  (p=0.008 n=5+5)
AggregationWithOTTL/summary_passthrough-10          37.0 ± 0%      33.0 ± 0%  -10.81%  (p=0.008 n=5+5)

</details>

I am working on a few more ideas for optimizations, I hope to bring the allocations and the time/op down by at least 10% 🤞

lahsivjar:
(deleted the above comment on performance improvements)

I have made a silly mistake in the `Merge` function implementation due to which the recent commits are in an incorrect state 🤦 .

processor/lsmintervalprocessor/internal/merger/value.go
lahsivjar: Updated the code to encode tracker separately. This resulted in lesser allocation and better performance. The encoding is pretty straightforward right now and there are some scopes for further improvements but I don't think they would be a bottleneck.

processor/lsmintervalprocessor/internal/merger/value.go
lahsivjar: This happens due to metric overflows, I have added a comment to explain this better:

```
	// Remove any hanging metrics, scope, or resource which failed to have any
	// entries due to datapoints overflowing. Overflowing datapoints discards
	// that metric and creates a new overflow metric which might result in the
	// original metric and its parent to exist without any datapoints.
```

processor/lsmintervalprocessor/internal/merger/value.go
lahsivjar: Done! Also, did a bit of similar changes around the add datapoint methods.

processor/lsmintervalprocessor/internal/merger/value.go
lahsivjar: Done!

processor/lsmintervalprocessor/internal/merger/value.go
lahsivjar: >  Do we need to consider this?

I don't think so, at least not for now I think since even the identity packages are not considering this.

lahsivjar:
After the above changes, below is the benchmark diff from `main`:

<details>

<summary>Benchmark diff from `main`</summary>

```
name                                            old time/op    new time/op    delta
Aggregation/sum_cumulative-10                     10.0µs ± 5%    12.2µs ± 5%  +22.80%  (p=0.008 n=5+5)
Aggregation/sum_delta-10                          10.0µs ± 3%    12.3µs ± 2%  +23.51%  (p=0.008 n=5+5)
Aggregation/histogram_cumulative-10               10.1µs ± 3%    12.8µs ± 4%  +26.10%  (p=0.008 n=5+5)
Aggregation/histogram_delta-10                    10.4µs ± 2%    12.9µs ± 2%  +24.41%  (p=0.008 n=5+5)
Aggregation/exphistogram_cumulative-10            11.1µs ± 2%    12.7µs ± 1%  +14.86%  (p=0.008 n=5+5)
Aggregation/exphistogram_delta-10                 11.2µs ± 2%    13.4µs ± 8%  +19.06%  (p=0.008 n=5+5)
Aggregation/summary_enabled-10                    10.3µs ± 1%    13.2µs ± 4%  +28.69%  (p=0.008 n=5+5)
Aggregation/summary_passthrough-10                1.30µs ± 2%    1.39µs ± 6%   +6.82%  (p=0.048 n=5+5)
AggregationWithOTTL/sum_cumulative-10             12.0µs ± 3%    14.4µs ± 6%  +19.39%  (p=0.008 n=5+5)
AggregationWithOTTL/sum_delta-10                  12.2µs ± 4%    14.3µs ± 4%  +16.75%  (p=0.008 n=5+5)
AggregationWithOTTL/histogram_cumulative-10       12.5µs ± 0%    14.4µs ± 7%  +15.09%  (p=0.016 n=4+5)
AggregationWithOTTL/histogram_delta-10            11.8µs ± 7%    15.2µs ± 6%  +28.66%  (p=0.008 n=5+5)
AggregationWithOTTL/exphistogram_cumulative-10    13.2µs ± 6%    14.3µs ± 5%   +8.42%  (p=0.016 n=5+5)
AggregationWithOTTL/exphistogram_delta-10         13.2µs ± 4%    14.9µs ± 1%  +13.16%  (p=0.008 n=5+5)
AggregationWithOTTL/summary_enabled-10            12.7µs ± 5%    14.1µs ± 3%  +11.16%  (p=0.008 n=5+5)
AggregationWithOTTL/summary_passthrough-10        1.40µs ± 5%    1.32µs ± 3%   -5.87%  (p=0.032 n=5+5)

name                                            old alloc/op   new alloc/op   delta
Aggregation/sum_cumulative-10                     17.6kB ± 3%    23.8kB ± 1%  +35.18%  (p=0.016 n=5+4)
Aggregation/sum_delta-10                          17.6kB ± 1%    23.9kB ± 1%  +35.23%  (p=0.008 n=5+5)
Aggregation/histogram_cumulative-10               19.4kB ± 2%    24.7kB ± 5%  +27.10%  (p=0.008 n=5+5)
Aggregation/histogram_delta-10                    19.4kB ± 1%    24.7kB ± 5%  +27.42%  (p=0.008 n=5+5)
Aggregation/exphistogram_cumulative-10            21.7kB ± 0%    24.2kB ± 1%  +11.70%  (p=0.008 n=5+5)
Aggregation/exphistogram_delta-10                 22.2kB ± 1%    24.9kB ± 1%  +12.11%  (p=0.008 n=5+5)
Aggregation/summary_enabled-10                    18.5kB ± 1%    21.9kB ± 4%  +18.36%  (p=0.016 n=4+5)
Aggregation/summary_passthrough-10                1.38kB ± 0%    1.38kB ± 0%   +0.52%  (p=0.032 n=5+5)
AggregationWithOTTL/sum_cumulative-10             20.2kB ± 1%    23.6kB ± 2%  +16.63%  (p=0.008 n=5+5)
AggregationWithOTTL/sum_delta-10                  20.2kB ± 1%    23.7kB ± 1%  +17.35%  (p=0.008 n=5+5)
AggregationWithOTTL/histogram_cumulative-10       22.1kB ± 3%    26.9kB ± 2%  +21.82%  (p=0.008 n=5+5)
AggregationWithOTTL/histogram_delta-10            21.9kB ± 2%    26.5kB ± 0%  +20.91%  (p=0.008 n=5+5)
AggregationWithOTTL/exphistogram_cumulative-10    23.0kB ± 5%    26.8kB ± 1%  +16.58%  (p=0.008 n=5+5)
AggregationWithOTTL/exphistogram_delta-10         23.6kB ± 3%    27.5kB ± 1%  +16.38%  (p=0.008 n=5+5)
AggregationWithOTTL/summary_enabled-10            20.7kB ± 4%    24.3kB ± 2%  +17.67%  (p=0.008 n=5+5)
AggregationWithOTTL/summary_passthrough-10        1.38kB ± 1%    1.38kB ± 0%     ~     (p=0.159 n=5+5)

name                                            old allocs/op  new allocs/op  delta
Aggregation/sum_cumulative-10                        199 ± 1%       235 ± 0%  +18.41%  (p=0.008 n=5+5)
Aggregation/sum_delta-10                             200 ± 0%       237 ± 0%  +18.06%  (p=0.008 n=5+5)
Aggregation/histogram_cumulative-10                  211 ± 1%       246 ± 0%  +16.67%  (p=0.008 n=5+5)
Aggregation/histogram_delta-10                       213 ± 0%       248 ± 0%  +16.34%  (p=0.008 n=5+5)
Aggregation/exphistogram_cumulative-10               228 ± 0%       264 ± 0%  +15.79%  (p=0.008 n=5+5)
Aggregation/exphistogram_delta-10                    237 ± 0%       273 ± 0%  +15.19%  (p=0.016 n=5+4)
Aggregation/summary_enabled-10                       212 ± 0%       248 ± 0%  +16.98%  (p=0.029 n=4+4)
Aggregation/summary_passthrough-10                  37.0 ± 0%      37.0 ± 0%     ~     (all equal)
AggregationWithOTTL/sum_cumulative-10                227 ± 0%       263 ± 0%  +15.66%  (p=0.008 n=5+5)
AggregationWithOTTL/sum_delta-10                     229 ± 0%       264 ± 0%  +15.28%  (p=0.016 n=4+5)
AggregationWithOTTL/histogram_cumulative-10          244 ± 1%       281 ± 2%  +15.33%  (p=0.008 n=5+5)
AggregationWithOTTL/histogram_delta-10               246 ± 1%       279 ± 2%  +13.07%  (p=0.008 n=5+5)
AggregationWithOTTL/exphistogram_cumulative-10       256 ± 0%       292 ± 0%  +14.06%  (p=0.029 n=4+4)
AggregationWithOTTL/exphistogram_delta-10            265 ± 0%       301 ± 0%  +13.58%  (p=0.016 n=4+5)
AggregationWithOTTL/summary_enabled-10               240 ± 0%       275 ± 0%  +14.58%  (p=0.008 n=5+5)
AggregationWithOTTL/summary_passthrough-10          37.0 ± 0%      37.0 ± 0%     ~     (all equal)
```
</details>

TL;DR, there is still ~20% performance degradation with overflows. We could eliminate the overflow path when it is not defined but that will only give us a false sense of improvement since, for our use case, we would always have overflows defined.

processor/lsmintervalprocessor/internal/merger/limits/tracker.go
```diff
@@ -0,0 +1,297 @@
...
+	var offset int
+	binary.BigEndian.PutUint64(b[offset:(offset+8)], t.observedCount)
+	offset += 8
+	if t.overflowCounts != nil {
+		hll, err := t.overflowCounts.MarshalBinary()
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal limits: %w", err)
+		}
+		b = slices.Grow(b, len(hll))[:offset+len(hll)]
+		copy(b[offset:], hll)
```
axw, line 114: We should look at making the HLL++ implement the coming-soon BinaryAppender interface: https://pkg.go.dev/encoding@master#BinaryAppender

I'll look into this: https://github.com/axiomhq/hyperloglog/pull/47

processor/lsmintervalprocessor/internal/merger/limits/tracker.go
```diff
@@ -0,0 +1,297 @@
...
+//	    panic(err)
+//	}
+func (t *Tracker) Unmarshal(d []byte) error {
+	if len(d) < 8 {
+		return errors.New("failed to unmarshal tracker, invalid length")
+	}
+
+	var offset int
+	t.observedCount = binary.BigEndian.Uint64(d[offset:(offset + 8)])
+	offset += 8
```
axw, line 132:
```suggestion
	t.observedCount = binary.BigEndian.Uint64(d[:8])
	offset := 8
```

processor/lsmintervalprocessor/internal/merger/limits/tracker.go
```diff
@@ -0,0 +1,297 @@
...
+	resourceTracker trackerType = iota
+	scopeTracker
+	dpsTracker
+)
+
+// Trackers represent multiple tracker in an ordered structure. It takes advantage
+// of the fact that pmetric DS is ordered and thus allows trackers to be created
+// for each resource, scope, and datapoint independent of the pmetric datastructure.
+// Note that this means that the order for pmetric and trackers are implicitly
+// related and removing/adding new objects to pmetric should be accompained by
```
axw, line 155:
```suggestion
// related and removing/adding new objects to pmetric should be accompanied by
```

processor/lsmintervalprocessor/internal/merger/limits/tracker.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+func (t *Trackers) GetScopeDPsTracker(i int) *Tracker {
+	if i >= len(t.scopeDPs) {
+		return nil
+	}
+	return t.scopeDPs[i]
+}
+
+func (t *Trackers) Marshal() ([]byte, error) {
```
axw, line 208: Should we add a version byte (first byte) for future proofing?

processor/lsmintervalprocessor/internal/merger/limits/tracker.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+	if t.overflowCounts == nil {
+		t.overflowCounts = other.overflowCounts.Clone()
+		return nil
+	}
+	return t.overflowCounts.Merge(other.overflowCounts)
+}
+
+// Marshal marshals the tracker to a byte slice.
+func (t *Tracker) Marshal() ([]byte, error) {
```
axw, line 100: Should this also be `AppendBinary`? Then we could avoid a slice allocation & copy in `Value.Marshal`

processor/lsmintervalprocessor/internal/merger/value.go
```diff
@@ -0,0 +1,776 @@
...
+			s.scopeDPLimitCfg.Overflow.Attributes,
+		); err != nil {
+			return pmetric.Metrics{}, fmt.Errorf("failed to finalize merged metric: %w", err)
+		}
+		overflowDP.SetIntValue(int64(sm.datapointsLimits.EstimateOverflow()))
+	}
+	// Remove any hanging metrics, scope, or resource which failed to have any
+	// entries due to datapoints overflowing. Overflowing datapoints discards
+	// that metric and creates a new overflow metric which might result in the
+	// original metric and its parent to exist without any datapoints.
```
axw, line 274: (Continuing from https://github.com/elastic/opentelemetry-collector-components/pull/235/#discussion_r1881559686.)

I still don't understand how we get into this situation. Could you give an example? If we've gotten to the point that we've overflowed, doesn't that imply that there's a non-zero amount of non-overflowing data points...?

axw (commented):
Getting back up to speed after PTO - mostly minor comments

processor/lsmintervalprocessor/internal/merger/limits/tracker.go
lahsivjar: Sounds good. We won't be able to fully utilize this though since it is not possible to estimate the size of the tracker prior to marshaling, so we could still end up reallocating the slice. Also see: https://github.com/axiomhq/hyperloglog/issues/44

processor/lsmintervalprocessor/internal/merger/value.go
lahsivjar: > Could you give an example? If we've gotten to the point that we've overflowed, doesn't that imply that there's a non-zero amount of non-overflowing data points...?

The issue arises due to 2 points:

1. Overflows can happen for datapoints/streams before they happen for resource or scope level.
2. Based on our modeling of the problem, when an overflow limit is breached then the dimensions of the overflow bucket are identified by the configuration.

For example: If we have to add a resource metric, scope metric and a datapoint, then, the order in which we will execute this is:

1. Add a resource metric - success since no overflow limit has been breached
2. Add a scope metric - success since no overflow limit has been breached
3. Add a metric - no overflow handling defined for this so far
4. Add a datapoint - overflow limit breached - no datapoint added

In such cases we have dangling resources. We could track if new entities are created and propagate the overflow status upwards - eventually removing any new parent added in case datapoints overflow but I haven't done that so far.

processor/lsmintervalprocessor/internal/merger/limits/tracker.go
lahsivjar: Done in https://github.com/elastic/opentelemetry-collector-components/pull/235/commits/4446f97f592f931916dbc9c2e146e1737a54babe

lahsivjar:
@axw I have been thinking about the limits again today and I am not sure if we need to keep the scope limits, maybe just having resource limits and resource datapoints limits are enough for all use-cases? Do you see a valid case where scope limits would be useful?

processor/lsmintervalprocessor/internal/merger/value.go
axw: Thanks for the explanation, it makes sense to me now.

processor/lsmintervalprocessor/internal/merger/limits/tracker.go
axw: Yeah no worries, we can follow up on this

axw:
> @axw I was already planning to add resource datapoints limits but now I am questioning myself if scope limits and scope datapoint limits are required or even useful, maybe just having resource limits and resource datapoints limits are enough for all use-cases? Do you see a valid case where scope limits would be useful?

I don't think that we would encounter high-cardinality scopes, but it may still be useful to have an explicit limit for scopes independent of metrics/datapoints, to avoid high cardinality metric names or dimensions from causing scopes to be washed out. That's the same reason why I think we should have a metric limit.

So what I have in mind for limits is to following the hierarchy: resource, resource-scope, resource-scope-metric, resource-scope-metric-datapoint.

Why were you planning to add a resource-datapoint limit?

axw (commented):
Looks good now, thanks for the updates. Let's resolve the discussion about limits and then get it in :)

lahsivjar:
> So what I have in mind for limits is to following the hierarchy: resource, resource-scope, resource-scope-metric, resource-scope-metric-datapoint.

This also sounds good to me.

> Why were you planning to add a resource-datapoint limit?

The main intention for adding this was to provide a way to cater to cases when we are not sure how many scope metrics can be within a scope but we don't want to waste the limit either i.e. if we expect number of scope metrics to have a high variance we could just do resource datapoint limit and prevent losing the unutilized limits for resources with lower number of scopes... but now that I think about this, I am not sure if scope metrics are meant to be high cardinality - maybe I got ahead of the problem without thinking it is actually a problem 😓 .

axw:
>> So what I have in mind for limits is to following the hierarchy: resource, resource-scope, resource-scope-metric, resource-scope-metric-datapoint.
>
> This also sounds good to me.

Do you want to do that in this PR, or in a followup?

lahsivjar:
> Do you want to do that in this PR, or in a followup?

I will do it in a followup

axw (approved):

PR #246, opened 2024-12-13
Add load generator otelcol

carsonip:
Add a load generator otel collector that replays and rewrites canned opentelemetry-demo data.


carsonip (commented):
TODO: see if it is possible to deduplicate testdata/traces to reduce file size

Makefile
```diff
@@ -89,4 +89,14 @@ builddocker:
...
\ No newline at end of file
+	docker build -t $$IMAGE_NAME -f distributions/elastic-components/Dockerfile .
+
+# Build a loadgen collector
+.PHONY: genloadgencol
+genloadgencol: $(BUILDER)
+	$(BUILDER) --config ./loadgen/manifest.yaml
+
+# Validate that the loadgen collector can run with the example configuration.
+.PHONY: loadgencol-validate
```
carsonip, line 100: TODO: add run loadgen command

carsonip (commented):
TODO:
- add docs on methodology
- how to get throughput, maybe from internal telemetry

carsonip (commented):
TODO: clean up non-compliant semconv attributes

constanca-m:
Looks really nice!

I have tested it, and it seems to work fine - and it generates logs really fast! I can see why the rate limiter would be useful for this case.

I tried to test it with the elasticsearch exporter, but then I ran into the error of missing the mandatory header project ID. For the hotel project, we add an authenticator to check it (see for example the [otlp receiver in our setup](https://github.com/elastic/hosted-otel-collector/blob/5d0c46b1ecaee1d444ff4a32af861a44a3dbf8ba/helm/templates/configmap.yaml#L39-L40)). Would it be possible to add a new field to the configuration to allow for an authenticator?

loadgen/manifest.yaml
```diff
@@ -0,0 +1,22 @@
+dist:
```
rogercoll, line 1: Do we want to keep this distribution manifest? Should we add the `loadgenreceiver` into the elastic-components distribution instead https://github.com/elastic/opentelemetry-collector-components/blob/main/distributions/elastic-components/manifest.yaml?

receiver/loadgenreceiver/logs.go
```diff
@@ -0,0 +1,151 @@
...
+	"go.opentelemetry.io/collector/receiver"
+	"go.uber.org/zap"
+)
+
+//go:embed testdata/logs.jsonl
+var demoLogs []byte
+
+type receiverLogs struct {
+	logs     plog.Logs
+	jsonSize int
```
rogercoll, line 41: I think we can remove this `jsonSize` field as we recommend controlling the throughput through the `ratelimit` processor.  (same for metrics and traces implementation)

receiver/loadgenreceiver/logs.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+func (ar *logsGenerator) nextLogs() (plog.Logs, int, error) {
+	now := pcommon.NewTimestampFromTime(time.Now())
+
+	nextLogs := plog.NewLogs()
+
+	ar.sampleLogs[ar.lastSampleIndex].logs.CopyTo(nextLogs)
+	sampledSize := ar.sampleLogs[ar.lastSampleIndex].jsonSize
```
rogercoll, line 136: `sampledSize` not needed

receiver/loadgenreceiver/logs.go
```diff
@@ -0,0 +1,151 @@
...
+			select {
+			case <-startCtx.Done():
+				return
+			default:
+				nMetrics, _, err := ar.nextLogs()
+				if err != nil {
+					ar.logger.Error(err.Error())
+					continue
+				}
+				err = ar.consumer.ConsumeLogs(startCtx, nMetrics)
```
rogercoll, line 112: qq Will this method block once the configured rate limit is reached?

carsonip:
> I tried to test it (edit: the loadgen receiver) with the elasticsearch exporter, but then I ran into the error of missing the mandatory header project ID. For the hotel project, we add an authenticator to check it (see for example the [otlp receiver in our setup](https://github.com/elastic/hosted-otel-collector/blob/5d0c46b1ecaee1d444ff4a32af861a44a3dbf8ba/helm/templates/configmap.yaml#L39-L40)). Would it be possible to add a new field to the configuration to allow for an authenticator?

That is because it only supports secret token at the moment (that's why I use apm server to test it). It will need extra support for API key and project ID.

receiver/loadgenreceiver/logs.go
carsonip: It will not. It keeps erroring out with too many requests (what apm-server otlp returns):
```
2025-01-09T15:32:17.362Z	error	loadgenreceiver/metrics.go:114	too many requests	{"kind": "receiver", "name": "loadgen", "data_type": "metrics"}
github.com/elastic/opentelemetry-collector-components/receiver/loadgenreceiver.(*metricsGenerator).Start.func1
	/home/carson/projects/opentelemetry-collector-components/receiver/loadgenreceiver/metrics.go:114
```

loadgen/manifest.yaml
carsonip: It is used to generate a purpose-built loadgencol binary under _loadgenbuild. Although in the near future I'm thinking about not using ocb and just wrap and distribute it with out own main function, so it may irrelevant soon.

Are you proposing to just add it to collector components, and to run loadgen it will just be using a specific config with this distro?

receiver/loadgenreceiver/logs.go
carsonip: We could add a little bit of backoff for "too many requests" specifically to avoid hogging the cpu.

carsonip: the too many requests is actually from ratelimiter. Might need to make ratelimiter configurable, whether to block or to return immediately.

carsonip:
> but then I ran into the error of missing the mandatory header project ID.

@constanca-m  Have you tried to configure the project header according to https://opentelemetry.io/docs/collector/configuration/#location like this?
```
make loadgencol-run --config="yaml:exporters::otlp::headers:fooheader bar"
```

receiver/loadgenreceiver/logs.go
carsonip: Disabling ratelimitprocessor for now. We can improve ratelimitprocessor later.

receiver/loadgenreceiver/logs.go
```diff
@@ -0,0 +1,132 @@
...
+	ar.cancelFn = cancelFn
+
+	go func() {
+		for {
+			select {
+			case <-startCtx.Done():
+				return
+			default:
+				if err := ar.consumer.ConsumeLogs(startCtx, ar.nextLogs()); err != nil {
+					ar.logger.Error(err.Error())
```
constanca-m, line 99: If I add a rate limiter, this will print errors non-stop, and it is very hard to read the rest of the logs from my collector container. @carsonip How did you work around this? I have tried the filter processor, but from my understanding this filter only works to filter the actual logs from the loadgenreceiver, not the logs from the logger.

constanca-m:
> @constanca-m Have you tried to configure the project header according to https://opentelemetry.io/docs/collector/configuration/#location like this?

I have not tried it yet, I will try today.

Why is there a `loadgen` and `loadgen` receiver? I am having trouble understanding that. Is it so we can send data to receivers as well using the `loadgen` similarly to `telemetrygen`?

loadgen/manifest.yaml
rogercoll: > Are you proposing to just add it to collector components, and to run loadgen it will just be using a specific config with this distro?

Exactly, I would reuse the already available distribution. That distribution is used in the CI to ensure a binary can be built from the linked components. And extend or add a new config file with the loadgen receiver https://github.com/elastic/opentelemetry-collector-components/blob/main/distributions/elastic-components/config.yaml

If the loadgen receiver needs to be included in a very small collector, I would suggest moving the `loadgen` directory under the `./distributions` one.

receiver/loadgenreceiver/logs.go
rogercoll: > the too many requests is actually from ratelimiter. Might need to make ratelimiter configurable, whether to block or to return immediately. I haven't got any 429 from apm-server yet.

(for future) What about wrapping it under an exponential backoff strategy (e.g https://github.com/cenkalti/backoff?tab=readme-ov-file)? (to auto determine an acceptable ratelimit)

receiver/loadgenreceiver/logs.go
carsonip: > If I add a rate limiter, this will print errors non-stop, and it is very hard to read the rest of the logs from my collector container. @carsonip How did you work around this? I have tried the filter processor, but from my understanding this filter only works to filter the actual logs from the loadgenreceiver, not the logs from the logger.

ratelimitprocessor now returns an error whenever the receiver exceeds the rate limit. IMO we should make it configurable such that instead of returning an error immediately when it is throttled, it can optionally block until the events are sent. And the load generator should use this option, to reduce noise in logs.

carsonip: This is specifically the TODO stated in ratelimitprocessor: https://github.com/elastic/opentelemetry-collector-components/blob/76ebae8ff6a9e10094a7a417655c1ecdb040db87/processor/ratelimitprocessor/local.go#L60

Load generator should use delayed processing.

carsonip: draft PR: https://github.com/elastic/opentelemetry-collector-components/pull/293

carsonip:
> Why is there a loadgen and loadgen receiver? I am having trouble understanding that. Is it so we can send data to receivers as well using the loadgen similarly to telemetrygen?

Will rename it to make it less confusing. Now loadgen is the name of both the directory that houses config and ocb config to loadgencol, as well as the name of the receiver.

`telemetrygen` is also overloaded as contrib has a binary to produce some very basic telemetry and send to a target.

receiver/loadgenreceiver/logs.go
carsonip: Fixed

loadgen/manifest.yaml
carsonip: Done in b649fad

rogercoll: > I'm fine with using it if it isn't used for anything critical.

Anything, we're not releasing any package/binary of it.

receiver/loadgenreceiver/logs.go
rogercoll: The idea behind the exponential backoff would be to make the `loadgen` receiver yield the cpu once it reaches the limit, thus, making the `delay` sleep on the receiver. Currently, if a low rate limit is set, the receiver will overwhelm the logs with:
```
2025-01-13T19:23:36.418+0100    error   loadgenreceiver@v0.0.0/metrics.go:99    too many requests       {"kind": "receiver", "name": "loadgen", "data_type": "metrics"}
2025-01-13T19:23:36.419+0100    error   loadgenreceiver@v0.0.0/metrics.go:99    too many requests       {"kind": "receiver", "name": "loadgen", "data_type": "metrics"}
```

But my understanding is that we are testing on a known rate limit at the moment, so I agree on adding this feature later.

rogercoll (approved):
lgtm, added some minor comments related to the repo directories

loadgen/config.example.yaml
```diff
@@ -0,0 +1,48 @@
+receivers:
```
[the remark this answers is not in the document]
carsonip: It is not useful in this PR, but i'll add a lot more files to the directory later on in https://github.com/elastic/opentelemetry-collector-components/pull/291

.github/workflows/checks.yml
```diff
@@ -227,3 +227,5 @@ jobs:
       - uses: actions/checkout@v4
       - name: Build testing component's collector and validate example configuration
         run: make elasticcol-validate
+      - name: Validate loadgen example config
+        run: make loadgen-validate
```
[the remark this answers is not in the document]
carsonip: Including it in the catch-all config.yaml defeats the purpose of having an example config that minimally illustrates the loadgen pipelines. In any case, we'll need an example config for loadgen, and this make target validates that.

receiver/loadgenreceiver/logs.go
rogercoll: > can you retry after pulling in the latest changes of this PR?

Just pulled the repository and updated the `loadgen/config.example.yaml` with this rate limit configuration:

```yaml
  ratelimit:
    strategy: records
    rate: 5
    burst: 5
    throttle_behavior: delay
```

`$ ELASTIC_APM_SERVER_URL=http://localhost:8200 ELASTIC_APM_API_KEY=some_api_key ./_build/elastic-collector-components --config loadgen/config.example.yaml`

And a big amount of logs with "too many requests" are still being generated. As mentioned, if we do not plan to work with this low rate limits, I am fine with postponing this feature

carsonip: sorry I spoke too soon. The "too many requests" are no longer there at 5000, but are still there at 5. Let me check why.

carsonip: https://pkg.go.dev/golang.org/x/time/rate#Limiter.ReserveN
> The returned Reservation’s OK() method returns false if n exceeds the Limiter's burst size.

I think we can skip this edge case first. This seems to be working as expected. If it only allows 5 item burst, you can never send 6 items in 1 go.

loadgen/config.example.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+
+processors:
+  transform/rewrite:  # Rewrite telemetry to increase cardinality
+    trace_statements:
+      - context: span
+        statements:
+          # The worst way to generate a random ID, but is the simplest in OTTL
+          # Only randomize trace ID such that span relationships are still maintained
+          - set(trace_id.string, Substring(MD5(UUID()), 0, 32))
```
axw, line 23: As mentioned on Slack, we would ideally update trace IDs and span IDs so relationships are maintained. This can be done by adjusting each ID consistently across the batches within a file.

Not sure off hand how we would do that. One option could be to add support for OTTL to access client metadata, and then have the loadgen receiver inject a random value into client metadata. Alternatively it could add it as a scope attribute or something like that.

carsonip: After discussing with Felix, as span IDs are not guaranteed to be unique across different traces, just randomizing trace ID should cover most use cases. Let's leave randomizing span ID as a follow up.

receiver/loadgenreceiver/logs.go
```diff
@@ -0,0 +1,132 @@
...
+
+	nextLogs := plog.NewLogs()
+	ar.samples.Next().CopyTo(nextLogs)
+
+	rm := nextLogs.ResourceLogs()
+	for i := 0; i < rm.Len(); i++ {
+		for j := 0; j < rm.At(i).ScopeLogs().Len(); j++ {
+			for k := 0; k < rm.At(i).ScopeLogs().At(j).LogRecords().Len(); k++ {
+				smetric := rm.At(i).ScopeLogs().At(j).LogRecords().At(k)
+				smetric.SetTimestamp(now)
```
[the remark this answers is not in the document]
carsonip: hmm, won't that create overlap between batches when we replay faster than the canned data?

PR #247, opened 2024-12-13
[processor/ratelimit] Add new component ratelimitprocessor

carsonip:
Add new processor for configurable rate limiting. Use Apache 2.0 license.

Issue #171, opened 2024-10-09
Use wildcards for dependabot


lahsivjar:
Currently, dependabot.yml file uses a separate config for each module, however, the [wildcard issue linked in the file](https://github.com/elastic/opentelemetry-collector-components/blob/9e723e51485bbb83838d004a45e6b3485979eae7/.github/dependabot.yml#L4) as a reason for this has now been [closed](https://github.com/dependabot/dependabot-core/issues/2178).

PR #248, opened 2024-12-16
Add ratelimitprocessor to dependabot.yml

carsonip:
Follow up to #247


lahsivjar (approved):
High time we get https://github.com/elastic/opentelemetry-collector-components/issues/171 done 😓

PR #267, opened 2024-12-23
ci: group dependabot directories

kruskall:
try to reduce the spam of update PRs when a new otel version is released

Closes: https://github.com/elastic/opentelemetry-collector-components/issues/171

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,12 +3,7 @@ version: 2
 updates:
   - package-ecosystem: "gomod"
     directories:
-      - "/processor/lsmintervalprocessor/"
-      - "/processor/elasticinframetricsprocessor/"
-      - "/processor/elastictraceprocessor/"
-      - "/processor/ratelimitprocessor/"
-      - "/connector/signaltometricsconnector"
-      - "/internal/tools/"
+      - "**/*"
     schedule:
       interval: "daily"
     labels:


rogercoll (commented):
Will the old otel grouping be preserved? 

```
groups:
      otel-dependencies:
        patterns:
          - "go.opentelemetry.io/*"
          - "github.com/open-telemetry/opentelemetry-collector-contrib/*"
```

I think the grouping feature is quite handy to avoid a big amount of PRs.

.github/dependabot.yml
```diff
@@ -1,69 +1,14 @@
...
-
-  - package-ecosystem: "gomod"
-    directory: "/connector/signaltometricsconnector"
+    directories:
+      - "/processor/lsmintervalprocessor/"
+      - "/processor/elasticinframetricsprocessor/"
+      - "/processor/elastictraceprocessor/"
+      - "/processor/ratelimitprocessor/"
+      - "/connector/signaltometricsconnector"
+      - "/internal/tools/"
```
lahsivjar, line 11: Can we use wildcards here?

lahsivjar: Also, I don't quite get what this PR will reduce, IIUC, we would still get as many PRs as the number of components or am I mistaken? (I would not want a single PR with all components as otel it is quite common for components to require fixes due to upgrade of otel versions and if multiple components break in a single PR then it would be quite a challenge to fix that in addition to the challenge with different teams having accountability for different components).

lahsivjar: Breaking changes are proving more difficult than having a PR owned by multiple teams 😢 . I am bringing back the proposal to group all deps together via https://github.com/elastic/opentelemetry-collector-components/pull/486/files PR.

PR #291, opened 2025-01-09
[loadgen] Separate into otelbench for benchmarking, otelsoak for soak testing

carsonip:
Separate loadgen into otelbench and otelsoak.

## otelbench

Run go benchmark on top of loadgen collector for a predefined duration `-test.benchtime`. `otelbench` accepts go benchmark flags, an otel collector config yaml, and some handy command line options to override the config. It produces go benchmark output.

Run `./otelbench -config ./config.yaml -test.benchtime=1m` with `ELASTIC_APM_SERVER_URL` and `ELASTIC_APM_SECRET_TOKEN` env vars or flags.

Example output:
```
logs-otlp       	      16	4521329779 ns/op	         0 failed_logs/s	         0 failed_metric_points/s	         0 failed_requests/s	         0 failed_spans/s	       319.2 logs/s	         0 metric_points/s	       136.5 requests/s	         0 spans/s
logs-otlphttp   	      19	3900502735 ns/op	         0 failed_logs/s	         0 failed_metric_points/s	         0 failed_requests/s	         0 failed_spans/s	       370.0 logs/s	         0 metric_points/s	       158.2 requests/s	         0 spans/s
metrics-otlp    	    3584	  20884237 ns/op	         0 failed_logs/s	         0 failed_metric_points/s	         0 failed_requests/s	         0 failed_spans/s	         0 logs/s	      5746 metric_points/s	       143.6 requests/s	         0 spans/s
metrics-otlphttp	    3506	  19158620 ns/op	         0 failed_logs/s	         0 failed_metric_points/s	         0 failed_requests/s	         0 failed_spans/s	         0 logs/s	      6263 metric_points/s	       156.6 requests/s	         0 spans/s
traces-otlp     	      84	 730912803 ns/op	         0 failed_logs/s	         0 failed_metric_points/s	         0 failed_requests/s	         0 failed_spans/s	         0 logs/s	         0 metric_points/s	       153.2 requests/s	      1034 spans/s
traces-otlphttp 	      88	 716443449 ns/op	         0 failed_logs/s	         0 failed_metric_points/s	         0 failed_requests/s	         0 failed_spans/s	         0 logs/s	         0 metric_points/s	       156.3 requests/s	      1055 spans/s
```

## otelsoak

otelsoak is the Elastic collector components distro with an example collector config file.


loadgen/cmd/otelsoak/main.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		log.Fatalf("collector new settings error: %v", err)
+	}
+
+	cmd := otelcol.NewCommand(settings)
+	cmd.Flags().AddGoFlagSet(loadgen.FlagSet)
+	runE := cmd.RunE
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		// This is to pass in parsed loadgen flags e.g. api-key, secret-token to otel collector
+		sets := loadgen.CollectorSetFromConfig()
```
carsonip, line 42: [to reviewer] this is 99% the same as the collector generated by ocb, with the exception that it supports `-api-key` and `-secret-token` etc via command line options, and translate them to `--set` in collector to override config in config.yaml.

loadgen/config.yaml
```diff
@@ -0,0 +1,53 @@
+receivers:
+  loadgen:
+
+exporters:
+  otlp:
+    endpoint: "${env:ELASTIC_APM_SERVER_URL}"
+    headers:
+     # Authorization: "Bearer ${env:ELASTIC_APM_SECRET_TOKEN}"
```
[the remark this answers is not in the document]
carsonip: In other words, do you see it as a requirement where config.yaml is not expected to be modified, and user of the load generator collector should be able to pass in either `ELASTIC_APM_API_KEY` or `ELASTIC_APM_SECRET_TOKEN` and expect it to work? Or will the user be capable of doing `--set=exporter.otlp.headers.Authorization=FOOBAR` where FOOBAR can be bearer or apikey?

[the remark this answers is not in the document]
carsonip: This is also the issue faced by https://github.com/elastic/opentelemetry-demo/issues/78 which resulted in a strange fix like https://github.com/elastic/opentelemetry-demo/pull/90

loadgen/cmd/otelbench/main.go
```diff
@@ -0,0 +1,72 @@
...
+		stop := make(chan bool)
+
+		go func() {
+			ticker := time.NewTicker(500 * time.Millisecond)
+			defer ticker.Stop()
+			for {
+				select {
+				case <-stop:
+				case <-ticker.C:
+					logs, metricPoints, spans, err := loadgen.GetTelemetrySent()
```
carsonip, line 46: TODO: check if we can mess with global registry to get from collector directly

loadgen/cmd/otelbench/main.go
```diff
@@ -0,0 +1,72 @@
...
+						b.ReportMetric(float64(metricPoints)/b.Elapsed().Seconds(), "metric_points/s")
+						b.ReportMetric(float64(spans)/b.Elapsed().Seconds(), "spans/s")
+						b.ReportMetric(float64(total)/b.Elapsed().Seconds(), "total/s")
+						return
+					}
+				}
+			}
+		}()
+
+		err := loadgen.Run(context.Background(), stop)
```
carsonip, line 65: TODO: may be good to have a configurable warm up time

loadgen/cmd/otelbench/main.go
```diff
@@ -0,0 +1,72 @@
...
+			for {
+				select {
+				case <-stop:
+				case <-ticker.C:
+					logs, metricPoints, spans, err := loadgen.GetTelemetrySent()
+					if err != nil {
+						b.Logf("error getting internal telemetry: %s", err)
+						continue
+					}
+					total := logs + metricPoints + spans
```
axw, line 51: This feels a bit wrong. `b.N` is meant to count iterations, and in this case I would expect one iteration to be sending all the configured payloads once, so each iteration is equivalent. The asynchronous check also makes me a bit uncomfortable -- it may be a bit inaccurate as a result.

Perhaps the `loadgenreceiver.NewFactory` function could take some options, one of which is a channel for controlling the iterations of the receiver? Then the otelbench command could inject a channel. We shouldn't need another goroutine then: the `testing.B` loop would be interacting with the receiver through the channel.

loadgen/cmd/otelbench/runner.go
```diff
@@ -0,0 +1,127 @@
...
+	Headers             map[string]string
+	CollectorConfigPath string
+}
+
+var FlagSet = flag.NewFlagSet("", flag.ExitOnError)
+
+func Init() {
+	// Server config
+	FlagSet.Func(
+		"server",
```
axw, line 26: Maybe "endpoint", for consistency with OTel Collector configuration?

loadgen/cmd/otelbench/runner.go
```diff
@@ -0,0 +1,127 @@
...
+		"server URL (default http://127.0.0.1:8200)",
+		func(server string) (err error) {
+			if server != "" {
+				Config.ServerURL, err = url.Parse(server)
+			}
+			return
+		})
+	FlagSet.StringVar(&Config.SecretToken, "secret-token", "", "secret token for APM Server")
+	FlagSet.StringVar(&Config.APIKey, "api-key", "", "API key for APM Server")
+	FlagSet.BoolVar(&Config.Secure, "secure", false, "validate the remote server TLS certificates")
```
axw, line 36: `insecure-skip-verify`?

loadgen/cmd/otelbench/main.go
carsonip: I'm wondering how to do this at implementation level, since logs, metrics and traces pipelines are separate. Do we want logs + metrics + traces iterations to be greater than b.N? My PR at its existing state has the same problem, which causes some imbalance between log records, metric data points and spans. Should we just benchmark the signals separately?

axw: > Should we just benchmark the signals separately?

That sounds like a good idea. They could have reasonably different performance profiles too, so combining them may be unhelpful.

loadgen/cmd/otelbench/runner.go
carsonip: I mixed up `insecure` and `insecure-skip-verify`. Fixed.

loadgen/cmd/otelbench/main.go
carsonip: Not using internal telemetry now

loadgen/cmd/otelbench/main.go
carsonip: Let's do it in a follow up.

loadgen/cmd/otelbench/main.go
carsonip: Injecting channels to NewFactory for testing.Benchmark to get notified when collector finishes sending b.N iterations of data set.

loadgen/cmd/otelbench/main.go
```diff
@@ -0,0 +1,136 @@
...
+
+					close(stop)
+				}()
+
+				err := RunCollector(context.Background(), stop, configs(exporter, signal, b.N), logsDone, metricsDone, tracesDone)
+				if err != nil {
+					b.Fatal(err)
+				}
+			})
+			fmt.Printf("%-*s\t%s\n", maxLen, benchName, result.String())
```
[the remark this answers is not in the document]
axw: I don't know either, writing to stdout seems like the right thing to do - so we can easily tee the results to a file without getting any noise from debug logs.

loadgen/README.md
```diff
@@ -1,19 +1,11 @@
...
-```
-
-Alternatively, there's `ELASTIC_APM_SERVER_URL` and `ELASTIC_APM_API_KEY` env var handling out of the box. `ELASTIC_APM_SECRET_TOKEN` is NOT supported without changing `config.example.yaml`.
-
-```
-ELASTIC_APM_SERVER_URL=http://localhost:8200 ELASTIC_APM_API_KEY=some_api_key ../_build/elastic-collector-components --config config.example.yaml
-```
-
-Even better, create your own `config.yaml` from `config.example.yaml` to fit your needs.
+otelsoak and otelbench is synonymous to apmsoak and apmbench in [elastic/apm-perf](https://github.com/elastic/apm-perf).
```
axw, line 11:
```suggestion
otelsoak and otelbench are synonymous to apmsoak and apmbench in [elastic/apm-perf](https://github.com/elastic/apm-perf).
```

loadgen/cmd/otelbench/README.md
```diff
@@ -0,0 +1,129 @@
+# otelbench
+
+otelbench wraps the collector inside a Go test benchmark. It outputs throughput numbers in standard Go benchmark format. It uses a collector config yaml as a base, and apply some overrides from command line options.
```
axw, line 3:
```suggestion
otelbench wraps the collector inside a Go test benchmark. It outputs throughput numbers in standard Go benchmark format. It uses a collector config yaml as a base, and applies some overrides from command line options.
```

loadgen/cmd/otelbench/README.md
```diff
@@ -0,0 +1,129 @@
...
+  -endpoint-otlp value
+        target server endpoint for otlp exporter (default to value in config yaml)
+  -endpoint-otlphttp value
+        target server endpoint for otlphttp exporter (default to value in config yaml)
+  -exporter-otlp
+        benchmark exporter otlp (default true)
+  -exporter-otlphttp
+        benchmark exporter otlphttp (default true)
+  -header value
+        extra headers to use when sending data to the server
```
axw, line 24: Maybe describe the expected format here? Is it k=v, and you can repeat the flag?

receiver/loadgenreceiver/factory.go
```diff
@@ -18,20 +18,44 @@
...
+func NewFactory() receiver.Factory {
+	return receiver.NewFactory(
+		metadata.Type,
+		func() component.Config {
+			return createDefaultReceiverConfig(nil, nil, nil)
+		},
+		receiver.WithMetrics(createMetricsReceiver, component.StabilityLevelDevelopment),
+		receiver.WithTraces(createTracesReceiver, component.StabilityLevelDevelopment),
+		receiver.WithLogs(createLogsReceiver, component.StabilityLevelDevelopment),
+	)
```
axw, line 36:
```suggestion
	return NewFactoryWithDone(nil, nil, nil)
```

receiver/loadgenreceiver/logs.go
```diff
@@ -130,3 +145,7 @@ func (ar *logsGenerator) nextLogs() plog.Logs {
 
 	return nextLogs
 }
+
+func (ar *logsGenerator) isDone() bool {
+	return ar.samples.LoopCount() > 0 && ar.samples.LoopCount() >= ar.cfg.Logs.MaxReplay
```
axw, line 150: Why do we need the >0 check as well as comparison?

axw (approved):
Beautiful!

receiver/loadgenreceiver/logs.go
carsonip: :facepalm: meant to type ar.cfg.Logs.MaxReplay for unlimited replays.

axw (approved):

PR #293, opened 2025-01-10
[processor/ratelimit] Add `throttle_behavior` for optional delayed processing

carsonip:
Configurable behavior on rate limit exceeded via config `throttle_behavior`.
- `error`: original behavior, return error immediately without forwarding event
- `delay`: delayed processing

### DIFF
--- a/processor/ratelimitprocessor/gubernator.go
+++ b/processor/ratelimitprocessor/gubernator.go
@@ -115,7 +115,7 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	}
 
 	if isUnderLimit := resp.GetStatus() == gubernator.Status_UNDER_LIMIT; !isUnderLimit {
-		// TODO add configurable behaviour for returning an error vs. delaying processing.
+		// TODO support `throttle_behavior` config for returning an error vs. delaying processing.
 		r.set.Logger.Error(
 			"request is over the limits defined by the rate limiter",
 			zap.Error(errTooManyRequests),
--- a/processor/ratelimitprocessor/gubernator.go
+++ b/processor/ratelimitprocessor/gubernator.go
@@ -114,7 +114,7 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 		return errors.New(resp.GetError())
 	}
 
-	if isUnderLimit := resp.GetStatus() == gubernator.Status_UNDER_LIMIT; !isUnderLimit {
+	if resp.GetStatus() != gubernator.Status_UNDER_LIMIT {
 		// Same logic as local
 		switch r.cfg.ThrottleBehavior {
 		case ThrottleBehaviorError:
--- a/processor/ratelimitprocessor/gubernator.go
+++ b/processor/ratelimitprocessor/gubernator.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/processor"
@@ -85,6 +86,7 @@ func (r *gubernatorRateLimiter) Shutdown(ctx context.Context) error {
 
 func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	uniqueKey := getUniqueKey(ctx, r.cfg.MetadataKeys)
+	createdAt := time.Now().UnixMilli()
 	getRateLimitsResp, err := r.client.GetRateLimits(ctx, &gubernator.GetRateLimitsReq{
 		Requests: []*gubernator.RateLimitReq{{
 			Name:      r.set.ID.String(),
@@ -95,9 +97,7 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 			Limit:     int64(r.cfg.Rate), // rate is per second
 			Burst:     int64(r.cfg.Burst),
 			Duration:  1000, // duration is in milliseconds, i.e. 1s
-
-			// TODO specify CreatedAt, so resulting reset time is
-			// relative to the relative clock.
+			CreatedAt: &createdAt,
 		}},
 	})
 	if err != nil {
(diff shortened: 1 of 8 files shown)


processor/ratelimitprocessor/local.go
```diff
@@ -56,9 +56,17 @@ func (r *localRateLimiter) RateLimit(ctx context.Context, hits int) error {
...
+	case OnThrottleError:
+		if ok := limiter.AllowN(time.Now(), hits); !ok {
+			return errTooManyRequests
+		}
+	case OnThrottleBlock:
+		r := limiter.ReserveN(time.Now(), hits)
+		if !r.OK() {
+			return errTooManyRequests
+		}
+		time.Sleep(r.Delay())
```
constanca-m, line 69: Thanks for this change, the non stop errors make it really hard to read anything else.

So just to be clear: if we have the pipeline loadgen -> ratelimit -> debug, once ratelimit sleeps, the whole process gets blocked?

carsonip: > So just to be clear: if we have the pipeline loadgen -> ratelimit -> debug, once ratelimit sleeps, the whole process gets blocked?

Correct

constanca-m (approved):

carsonip:
@constanca-m Lmk if you want to take over this PR. I'm not too familiar with Gubernator and I'm not sure if LeakyBucket will actually give a correct reset_time. Do you have any idea? Otherwise, we can get this merged with only local rate limiter support.

processor/ratelimitprocessor/local.go
carsonip: I've tested it with https://github.com/elastic/opentelemetry-collector-components/pull/246 and there are no more "too many requests" flooding in logs

processor/ratelimitprocessor/gubernator.go
```diff
@@ -115,7 +115,7 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	}
 
 	if isUnderLimit := resp.GetStatus() == gubernator.Status_UNDER_LIMIT; !isUnderLimit {
-		// TODO add configurable behaviour for returning an error vs. delaying processing.
+		// TODO support `throttle_behavior` config for returning an error vs. delaying processing.
```
carsonip, line 118: [to reviewer] I haven't added support for Gubernator in this PR as I'm not sure if we can do it properly without polling. Lmk if this is a blocker, or whether this PR can be merged as-is.

constanca-m:
> I'm not too familiar with Gubernator and I'm not sure if LeakyBucket will actually give a correct reset_time. Do you have any idea?

No, these changes do not affect the rate limiter if we are using gubernator, unfortunately. Yes, if you want I can open a PR for the gubernator or reuse this one, it is up to you.

constanca-m:
The change should be quite easy @carsonip (I have tested it locally):

```go
func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
	uniqueKey := getUniqueKey(ctx, r.cfg.MetadataKeys)
	createdAt := time.Now().UnixMilli()
	getRateLimitsResp, err := r.client.GetRateLimits(ctx, &gubernator.GetRateLimitsReq{
		Requests: []*gubernator.RateLimitReq{{
			Name:      r.set.ID.String(),
			UniqueKey: uniqueKey,
			Hits:      int64(hits),
			Behavior:  r.behavior,
			Algorithm: gubernator.Algorithm_LEAKY_BUCKET,
			Limit:     int64(r.cfg.Rate), // rate is per second
			Burst:     int64(r.cfg.Burst),
			Duration:  1000, // duration is in milliseconds, i.e. 1s
			CreatedAt: &createdAt,                                                                    // <-------------------- Add
		}},
	})
	if err != nil {
		return fmt.Errorf("error executing gubernator rate limit request: %w", err)
	}

	// Inside the gRPC response, we should have a single-item list of responses.
	responses := getRateLimitsResp.GetResponses()
	if n := len(responses); n != 1 {
		return fmt.Errorf("expected 1 response from gubernator, got %d", n)
	}
	resp := responses[0]
	if resp.GetError() != "" {
		return errors.New(resp.GetError())
	}

	if isUnderLimit := resp.GetStatus() == gubernator.Status_UNDER_LIMIT; !isUnderLimit {
		// Same logic as local
		switch r.cfg.ThrottleBehavior {
		case ThrottleBehaviorError:
			r.set.Logger.Error(
				"request is over the limits defined by the rate limiter",
				zap.Error(errTooManyRequests),
				zap.String("processor_id", r.set.ID.String()),
				zap.Strings("metadata_keys", r.cfg.MetadataKeys),
			)
			return errTooManyRequests
		case ThrottleBehaviorDelay:
			r.set.Logger.Error(
				"request is over the limits defined by the rate limiter",
				zap.Error(errTooManyRequests),
				zap.String("processor_id", r.set.ID.String()),
				zap.Strings("metadata_keys", r.cfg.MetadataKeys),
			)
			delay := time.Duration(resp.GetResetTime()-createdAt) * time.Millisecond
			time.Sleep(delay)
		}
	}
	return nil
}
```

carsonip:
> Now I notice that the minimum requests we can send is 1 request per second, and to test with the rate limiter + loadgen receiver, it means the logs still contain a huge amount of errors.

Not sure if I understand this. Where does the "1 request per second" come from?

vigneshshanmugam:
> Now I notice that the minimum requests we can send is 1 request per second, and to test with the rate limiter + loadgen receiver, it means the logs still contain a huge amount of errors. But I guess it is better than what was before.

I guess that depends on the [Strategy](https://github.com/elastic/opentelemetry-collector-components/blob/0f49e600a9bd55aaf524e2146d8f4446928b6ae6/processor/ratelimitprocessor/config.go#L66-L84) we are going for in the loadgen?  

Also `1 request per second` doesn't sound right to me. Are you referring to Refill rate of the Leaky bucket or anything in particular?

vigneshshanmugam (approved):

processor/ratelimitprocessor/gubernator_test.go
```diff
@@ -92,58 +93,70 @@ func TestGubernatorRateLimiter_StartStop(t *testing.T) {
...
-
-	resp.Responses = []*gubernator.RateLimitResp{}
-	err = rateLimiter.RateLimit(context.Background(), 1)
-	assert.EqualError(t, err, "expected 1 response from gubernator, got 0")
-
-	resp.Responses = []*gubernator.RateLimitResp{{Status: gubernator.Status_OVER_LIMIT}}
-	err = rateLimiter.RateLimit(context.Background(), 1)
-	assert.EqualError(t, err, "too many requests")
-
-	resp.Responses = []*gubernator.RateLimitResp{{Status: -1}} // handled like OVER_LIMIT
```
carsonip, line 133: [to reviewer] I have removed this test. Not sure what's the point of testing an invalid gubernator response. The handling is undefined for delay behavior.

axw:
> Now I notice that the minimum requests we can send is 1 request per second, and to test with the rate limiter + loadgen receiver, it means the logs still contain a huge amount of errors. But I guess it is better than what was before.
>
> @axw @vigneshshanmugam Do you see any use in decreasing the possible amount of requests? Or do we leave it like this?

You're talking about the minimum fill rate, right? As in we can't force consumers to wait for more than 1 second to send a request. I think this is OK and don't have any uses cases for that, maybe later we can make it more flexible.

constanca-m:
> Not sure if I understand this. Where does the "1 request per second" come from?

Setting:

```yaml
  rate: 1
  burst: 1
```

In the rate limiter @carsonip 

> You're talking about the minimum fill rate, right? As in we can't force consumers to wait for more than 1 second to send a request. 

:+1:

constanca-m (approved):

processor/ratelimitprocessor/gubernator.go
```diff
@@ -115,14 +115,20 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	}
 
 	if isUnderLimit := resp.GetStatus() == gubernator.Status_UNDER_LIMIT; !isUnderLimit {
```
lahsivjar, line 117:
```suggestion
	if resp.GetStatus() != gubernator.Status_UNDER_LIMIT {
```
nit

lahsivjar (approved):
LGTM, minor nit!

axw (commented):
:man_facepalming: I failed to submit, I'll send a followup

PR #294, opened 2025-01-13
ElasticAPM receiver: handle transactions and spans

gregkalapos:
This PR adds mapping for transactions and spans from the Elastic APM data model to OTel.

It also adds tests based on golden. The idea for testing: we reuse ndjson files from the apm-server repo and send those to the receiver which we compare by using golden.

TODO (will bring these up in meetings) - done: 
- [x] Share attribute name constants with `opentelemetry-lib` - we do the reverse here what's the enricher does that - but we use the exact same Elastic specific attribute names. We could share all those constants across these 2 libs.  PR: https://github.com/elastic/opentelemetry-lib/pull/141 - The PR is merged, but opentelemetry-lib will need the next release. I have a todo for this in the PR, I'd merge as is and update later. 
- [x] Testing infrustructure

Many things are still outside this PR, like errors, logs, metrics. I propose to handle those in follow-ups, as they are fairly independent.


receiver/elasticapmreceiver/internal/mappers/IntakeV2ToDerivedFields.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func SetDerivedResourceAttributes(event *modelpb.APMEvent, attributes pcommon.Map) {
+	attributes.PutStr("agent.name", event.Agent.Name)
+	attributes.PutStr("agent.version", event.Agent.Version)
+}
+
+// Shared across spans and transactions
+func SetDerivedFieldsCommon(event *modelpb.APMEvent, attributes pcommon.Map) {
+	attributes.PutDouble("duration.us", float64(event.Event.Duration))
```
AlexanderWert, line 70: Isn't that already covered by `transaction.duration.us` and `span.duration.us` in the methods above?

receiver/elasticapmreceiver/internal/mappers/IntakeV2ToDerivedFields.go
```diff
@@ -0,0 +1,82 @@
...
+	attributes.PutStr("transaction.id", event.Transaction.Id)
+	attributes.PutStr("transaction.name", event.Transaction.Name)
+
+	attributes.PutBool("transaction.sampled", event.Transaction.Sampled)
+	// from whatever reason Transaction.Root is always false. That seems to be a derived field already - I don't see that fields directly on IntakeV2 - there is only ParentId
+	attributes.PutBool("transaction.root", event.ParentId == "")
+	attributes.PutStr("transaction.type", event.Transaction.Type)
+	attributes.PutStr("transaction.result", event.Transaction.Result)
+
+	attributes.PutDouble("transaction.duration.us", float64(event.Event.Duration))
```
AlexanderWert, line 45: In the enrichment lib this is an `Int` type (not double):
https://github.com/elastic/opentelemetry-lib/blob/995b8f40588aea9477b3c5205704ed6f3aa89f48/enrichments/trace/internal/elastic/span.go#L230

receiver/elasticapmreceiver/internal/mappers/IntakeV2ToDerivedFields.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func SetDerivedFieldsForSpan(event *modelpb.APMEvent, attributes pcommon.Map) {
+
+	attributes.PutStr("processor.event", "span")
+	attributes.PutDouble("span.duration.us", float64(event.Event.Duration))
```
AlexanderWert, line 51: In the enrichment lib this is an `Int` type (not double):
https://github.com/elastic/opentelemetry-lib/blob/995b8f40588aea9477b3c5205704ed6f3aa89f48/enrichments/trace/internal/elastic/span.go#L270

receiver/elasticapmreceiver/internal/mappers/IntakeV2ToSemConv.go
```diff
@@ -0,0 +1,83 @@
...
+	attributes.PutStr(semconv.AttributeDeploymentEnvironment, event.Service.Environment)
+}
+
+func TranslateIntakeV2TransactionToOTelAttributes(event *modelpb.APMEvent, attributes pcommon.Map) {
+	if event.Http.Request != nil {
+		attributes.PutStr(semconv.AttributeHTTPMethod, event.Http.Request.Method)
+		attributes.PutInt(semconv.AttributeHTTPStatusCode, int64(event.Http.Response.StatusCode))
+
+		if event.Url != nil {
+			attributes.PutStr(semconv.AttributeHTTPURL, event.Url.Full)
```
AlexanderWert, line 43: we need the new SemConv for URL here, NOT `http.url` (that is deprecated):
- url.full
- url.scheme
- url.domain
- ...

see https://github.com/elastic/opentelemetry-lib/blob/995b8f40588aea9477b3c5205704ed6f3aa89f48/enrichments/trace/internal/elastic/span.go#L139

receiver/elasticapmreceiver/internal/mappers/IntakeV2ToSemConv.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func TranslateIntakeV2SpanToOTelAttributes(event *modelpb.APMEvent, attributes pcommon.Map) {
+
+	if event.Http != nil {
+		attributes.PutStr(semconv.AttributeHTTPMethod, event.Http.Request.Method)
+		attributes.PutInt(semconv.AttributeHTTPStatusCode, int64(event.Http.Response.StatusCode))
+
+		if event.Url != nil {
+			attributes.PutStr(semconv.AttributeHTTPURL, event.Url.Full)
```
AlexanderWert, line 61: same as above

receiver/elasticapmreceiver/internal/mappers/IntakeV2ToDerivedFields.go
gregkalapos: Correct, dropped this.

receiver/elasticapmreceiver/internal/mappers/IntakeV2ToDerivedFields.go
```diff
@@ -0,0 +1,78 @@
...
+	attributes.PutInt("transaction.duration.us", int64(event.Event.Duration))
+}
+
+func SetDerivedFieldsForSpan(event *modelpb.APMEvent, attributes pcommon.Map) {
+
+	attributes.PutStr("processor.event", "span")
+	attributes.PutDouble("span.duration.us", float64(event.Event.Duration))
+	attributes.PutStr("span.id", event.Span.Id)
+	attributes.PutStr("span.name", event.Span.Name)
+	attributes.PutStr("span.type", event.Span.Type)
```
AlexanderWert, line 52: missing:
- span.action
- span.subtype

receiver/elasticapmreceiver/internal/mappers/IntakeV2ToSemConv.go
gregkalapos: Moved to SemConv v1.27.0 - that's the latest version in the latest go package.

receiver/elasticapmreceiver/receiver.go
```diff
@@ -236,18 +246,59 @@ func (r *elasticAPMReceiver) processBatch(ctx context.Context, batch *modelpb.Ba
 			ss := rs.ScopeSpans().AppendEmpty()
 			s := ss.Spans().AppendEmpty()
 
+			traceId, err := traceIDFromHex(event.Trace.Id)
```
rockdaboot, line 249: Is there a good reason not to log an error here (and below)? An issue in the format would be silently ignored, wrong values propagate into production and are possibly hard to recognize or debug.

receiver/elasticapmreceiver/internal/mappers/IntakeV2ToDerivedFields.go
```diff
@@ -0,0 +1,86 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
dmathieu, line 1: This is a nit, but this mapper's filenames don't really match the Go convention for file naming, which is snake case.

receiver/elasticapmreceiver/receiver.go
```diff
@@ -236,18 +246,59 @@ func (r *elasticAPMReceiver) processBatch(ctx context.Context, batch *modelpb.Ba
...
+			if err == nil {
```
dmathieu, line 250: I would keep the receiver code very small, and extract all the logic into internal methods, so they can be more easily tested. But maybe that's my preference.

receiver/elasticapmreceiver/receiver.go
gregkalapos: This is a good point, thanks. We should log this indeed.

I added error level logs and also added one more test case to cover when parsing fails.

Other (in my opinion less good) options I see other than the error logs are:
- just log as warning: I think missing trace id, span id, and parent id is more than just a warning. At least error is warranted because such traces won't be very useful - e.g. correlation won't work and similar
- return with an error and don't map the event: I think this would be too brutal and it'll result by completely missing events. There is nothing after this point in the pipeline that'd prevent the empty ids, so we'd still have all the other information. I feel keeping what's possible is better for users.

So from that reason, I just added an error log as suggested and let this thing happen.

receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
```diff
@@ -0,0 +1,86 @@
...
+	"strings"
+
+	"github.com/elastic/apm-data/model/modelpb"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+)
+
+// Sets fields that are NOT part of OTel for transactions. These fields are derived by the Enrichment lib in case of OTLP input
+func SetDerivedFieldsForTransaction(event *modelpb.APMEvent, attributes pcommon.Map) {
+
+	attributes.PutStr("processor.event", "transaction")
```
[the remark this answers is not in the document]
gregkalapos: The ones from here we can get from opentelemetry-lib once a new version is released - I added a comment about that: https://github.com/elastic/opentelemetry-collector-components/pull/294/files/77f273b6c0c659bbc022d37bde892d8c2c16019e#diff-4106909dc3c7a78517bfb31ac0562b41c1d5fac660a1217b75481c8dca8563e2R21

So these constants will come from opentelemetry-lib soon - all these attribute names are already defined there. 

The `attributes.go` in this PR only contain attributes that I really don't expect to be used by other components - those are these rarely used fields defined by the elastic apm data model. We could move those constants to opentelemetry-lib as well, but so far I don't really see a reason for it.

receiver/elasticapmreceiver/internal/mappers/intakeV2ToSemConv.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Translates resource attributes from the Elastic APM model to SemConv resource attributes
+func TranslateToOtelResourceAttributes(event *modelpb.APMEvent, attributes pcommon.Map) {
+	attributes.PutStr(semconv.AttributeServiceName, event.Service.Name)
+	attributes.PutStr(semconv.AttributeServiceVersion, event.Service.Version)
+	if event.Service.Language != nil {
+		attributes.PutStr(semconv.AttributeTelemetrySDKLanguage, event.Service.Language.Name)
+	}
+	attributes.PutStr(semconv.AttributeTelemetrySDKName, "ElasticAPM")
+	attributes.PutStr(semconv.AttributeDeploymentEnvironmentName, event.Service.Environment)
```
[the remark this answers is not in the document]
gregkalapos: You are right, thanks! I added a check here.

receiver/elasticapmreceiver/receiver_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	factory := NewFactory()
+	cfg := factory.CreateDefaultConfig()
+	testData := "testdata"
+
+	set := receivertest.NewNopSettings()
+	nextTrace := new(consumertest.TracesSink)
+	rec, _ := factory.CreateTraces(context.Background(), set, cfg, nextTrace)
+
+	if err := rec.Start(context.Background(), componenttest.NewNopHost()); err != nil {
```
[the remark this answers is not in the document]
gregkalapos: Love it! Updated.

receiver/elasticapmreceiver/receiver.go
gregkalapos: Yeah, I agree. I moved out all these lines into separate functions. One nit here: https://github.com/elastic/opentelemetry-collector-components/pull/294/files#r1963186038 - but I feel it's ok.

receiver/elasticapmreceiver/receiver_test.go
```diff
@@ -0,0 +1,105 @@
...
+	receiver := extractElasticAPMReceiver(rec)
+	if receiver == nil {
+		t.Fatal("Failed to extract elasticAPMReceiver")
+	}
+
+	protocol := "https"
+	if receiver.cfg.TLSSetting == nil {
+		protocol = "http"
+	}
+	resp, err := http.Post(protocol+"://"+receiver.cfg.Endpoint+intakePath, "application/x-ndjson", bytes.NewBuffer(data))
```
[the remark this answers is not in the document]
gregkalapos: Thanks, this is a good idea. I think this PR predates `testutil.GetAvailableLocalAddress(t)`, so that's why my less nice solution here originally.

Updated the PR and now I use `testutil.GetAvailableLocalAddress(t)` - some things I adapted, since now `http.Post(...)` is in a different method (due to another point above), and I did not want to pass the endpoint string around, so I still rely on `receiver.cfg.Endpoint` - but the endpoint itself is now defined by `testutil.GetAvailableLocalAddress(t)`.

receiver/elasticapmreceiver/receiver.go
```diff
@@ -275,6 +277,103 @@ func (r *elasticAPMReceiver) processBatch(ctx context.Context, batch *modelpb.Ba
...
+	if strings.EqualFold(event.Event.Outcome, "success") {
+		s.Status().SetCode(ptrace.StatusCodeOk)
+	} else if strings.EqualFold(event.Event.Outcome, "failure") {
+		s.Status().SetCode(ptrace.StatusCodeError)
+	}
+
+	return s
+}
+
+func (r *elasticAPMReceiver) elasticTransactionToOtelSpan(s *ptrace.Span, event *modelpb.APMEvent) {
```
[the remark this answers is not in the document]
gregkalapos: I was thinking the same 

But then `elasticEventToOtelSpan` would not need to return the `ptrace.Span`, because this is the only place where we use it. Also, I think the readability is better this way - otherwise `elasticEventToOtelSpan` would do a lot. So I ended up leaving it this way.

receiver/elasticapmreceiver/receiver.go
```diff
@@ -104,6 +110,10 @@ func (r *elasticAPMReceiver) startHTTPServer(ctx context.Context, host component
 	}
 
 	r.settings.Logger.Info("Starting HTTP server", zap.String("endpoint", r.cfg.ServerConfig.Endpoint))
+
+	// TODO: Remove this once we have a proper way to configure TLS
+	r.cfg.TLSSetting = nil
```
rogercoll, line 115: nit: could we set this in the factory.go `createDefaultConfig` default method instead? (to avoid mutating the config after being parsed)

receiver/elasticapmreceiver/internal/mappers/intakeV2ToSemConv.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Translates span attributes from the Elastic APM model to SemConv attributes
+func TranslateIntakeV2SpanToOTelAttributes(event *modelpb.APMEvent, attributes pcommon.Map) {
+	if event.Http != nil {
+
+		setHttpAttributes(event, attributes)
+
+		if event.Url != nil {
+			attributes.PutStr(semconv.AttributeURLFull, event.Url.Full)
```
rogercoll, line 61: Not sure if any of this inner fields can be nil, if so we might need to add a guard function for all the setters.

rogercoll (approved):
lgtm, added a couple of nits

receiver/elasticapmreceiver/receiver_test.go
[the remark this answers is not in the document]
rogercoll: > I did not want to pass the endpoint string around, so I still rely on receiver.cfg.Endpoint - but the endpoint itself is now defined by testutil.GetAvailableLocalAddress(t).

But then you need to do a type assertion on the receiver interface, is there any reason why not to use the string endpoint?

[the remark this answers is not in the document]
gregkalapos: My only thinking was to not pass too many parameters. On the other hand, passing the string is very lightweight compared to the type assertion.

Pushed an update to get rid of the type assertion and now we pass the string.

receiver/elasticapmreceiver/receiver_test.go
```diff
@@ -0,0 +1,110 @@
...
+	testEndpoint := testutil.GetAvailableLocalAddress(t)
+	cfg := &Config{
+		ServerConfig: confighttp.ServerConfig{
+			Endpoint: testEndpoint,
+		}}
+
+	set := receivertest.NewNopSettings()
+	nextTrace := new(consumertest.TracesSink)
+	rec, _ := factory.CreateTraces(context.Background(), set, cfg, nextTrace)
+	receiver := rec.(*sharedcomponent.Component[*elasticAPMReceiver]).Unwrap()
```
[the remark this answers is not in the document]
gregkalapos: Ok, let's go with that  https://github.com/elastic/opentelemetry-collector-components/pull/294#discussion_r1964106482

receiver/elasticapmreceiver/internal/mappers/intakeV2ToSemConv.go
gregkalapos: This is actually a good point - we need to guard these. For now I just added another check. This was already setting empty strings.

In the spec the required fields are listed: https://github.com/elastic/apm-server/tree/main/docs/spec/v2

If any of those are missing, apm server rejects those documents. But `url.full` isn't required.

This whole question makes me think... we should probably discuss how to handle invalid documents - like when a required field is missing.

In any case - this specific one is handled.

PR #298, opened 2025-01-14
[processor/lsminterval]Optimize critical section

lahsivjar:
Removes the next processor call from the critical section.

### DIFF
--- a/processor/lsmintervalprocessor/processor.go
+++ b/processor/lsmintervalprocessor/processor.go
@@ -292,6 +292,22 @@ func (p *Processor) ConsumeMetrics(ctx context.Context, md pmetric.Metrics) erro
 		return errors.Join(append(errs, fmt.Errorf("failed to marshal value to proto binary: %w", err))...)
 	}
 
+	if err := p.mergeToBatch(vb); err != nil {
+		return fmt.Errorf("failed to merge the value to batch: %w", err)
+	}
+
+	// Call next for the metrics remaining in the input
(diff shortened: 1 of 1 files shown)


processor/lsmintervalprocessor/processor.go
```diff
@@ -313,28 +328,19 @@ func (p *Processor) ConsumeMetrics(ctx context.Context, md pmetric.Metrics) erro
 
 	for _, k := range keys {
 		if err := p.batch.Merge(k, vb, nil); err != nil {
-			errs = append(errs, fmt.Errorf("failed to merge to db: %w", err))
+			return fmt.Errorf("failed to merge to db: %w", err)
 		}
 	}
 
 	if p.batch.Len() >= dbCommitThresholdBytes {
 		if err := p.batch.Commit(p.wOpts); err != nil {
```
lahsivjar, line 336: [**For reviewers**] I initially wanted to move the commit and close code outside hot path by creating a channel and  reading from it in the `Start` loop, however, that would increase the complexity of draining the channel before harvest as well as orchestrating a bit on when to drain the channel. Additionally, a no-buffer channel the performance gains were not substantial for now. We could still come back to this in the future for currently I think that optimization is not required.

Here is the branch with a rough draft of the approach for anyone interested: https://github.com/elastic/opentelemetry-collector-components/compare/main...lahsivjar:opentelemetry-collector-components:poc-get-commit-outside-hotpath?expand=1

axw (approved):

Issue #303, opened 2025-01-15
[extension/apmconfig] Port apm-server code to featch APM Central configs from Elasticsearch/Kibana


rogercoll:
See: https://github.com/elastic/apm-server/tree/main/internal/agentcfg

The extension should get the remote configurations from an Elasticsearch endpoint, instead of an apm-server endpoint.

rogercoll:
Done https://github.com/elastic/opentelemetry-lib/pull/143

PR #310, opened 2025-01-16
Adds internal code to fetch APM Agents configurations through Elasticsearch

rogercoll:
Fixes https://github.com/elastic/opentelemetry-collector-components/issues/303

Two internal packages:
- agentcfg: Types and fetcher interface related to APM remote configuration
- ~~elasticsearch: Implementation of Elasticsearch fetcher interface~~: replaced with go-elasticsearch/v8

Base code: https://github.com/elastic/apm-server/tree/main/internal/agentcfg with a few changes to simplify the code for this iteration:

- Metrics and tracing removed (should be changed to using otel libs)
- Logging changed to zap.Logger so it's aligned with collector's components factories

The code is internal as it will be shared among different Go modules within this repository, the `apmconfig` extension and the `elasticapm` receiver. The following is an example of how to use the `ElasticsearchFetcher`:

```go
	cl, _ := elasticsearch.NewClient(
		&elasticsearch.Config{
			Hosts:  elasticsearch.Hosts{"https://abc.us-central1.gcp.cloud.es.io:443"},
			APIKey: "adafdsa==",
		},
	)
	
	fetcher := agentcfg.NewElasticsearchFetcher(cl, 10*time.Second, logger)
	go func() {
		fetcher.Run(context.Background())
	}()

	result, _ := fetcher.Fetch(context.Background(), agentcfg.Query{
		Service: agentcfg.Service{
			Name: "my-service-name",
		},
	})

```


gregkalapos (approved):

axw (commented):
Would it make sense to add this as a non-internal package to https://github.com/elastic/opentelemetry-lib, and reuse it in apm-server?

rogercoll:
> Would it make sense to add this as a non-internal package to https://github.com/elastic/opentelemetry-lib, and reuse it in apm-server?

That makes sense to me, moving the files in https://github.com/elastic/opentelemetry-lib/pull/143 (any suggestion on the package name is more than welcomed)

rogercoll:
Closing it in favor of https://github.com/elastic/opentelemetry-lib/pull/143

PR #313, opened 2025-01-17
[receiver/loadgen] Simulate multiple agents via new `concurrency` config

### LINKED ISSUE #305: Load generator should be able to emulate multiple agents
The OTel load generator (which is being introduced in https://github.com/elastic/opentelemetry-collector-components/pull/291) currently only emulates a single agent which limits its usability. Extend it to enable it to emulate multiple agents configurable through the collector's config yaml.

carsonip:
Add a `concurrency` config to loadgenreceiver to simulate multiple agents (`agent_replicas` in apmsoak) sending otlp requests.

Fixes #305


receiver/loadgenreceiver/config.go
```diff
@@ -25,11 +25,16 @@ type (
...
+// Config defines configuration for loadgen receiver.
 type Config struct {
 	Metrics MetricsConfig `mapstructure:"metrics"`
 	Logs    LogsConfig    `mapstructure:"logs"`
 	Traces  TracesConfig  `mapstructure:"traces"`
+
+	// Concurrency is the amount of concurrency when sending to next consumer.
+	// The concurrent workers share the amount of workload, instead of multiplying the amount of workload,
+	// i.e. loadgenreceiver still sends up to the same MaxReplay limit.
+	Concurrency int `mapstructure:"concurrency"`
```
carsonip, line 37: [to reviewer] I'm not sure what's a good name here. I started with `num_workers` but changed to `concurrency`. In apm-perf it is `agent_replicas`, although the meaning is slightly different, because agent_replicas will imply the agents each send a MaxReplay amount of load. This difference is not meaningful in long-runnning scenarios e.g. apmsoak and otelsoak.

We'll also later have to decide on the name of the flag in otelbench. I'm still thinking about `concurrency`, but in apmbench it is called `agents`, but this time in apmbench the number of agents will _share_ b.N (via `b.RunParallel` and `pb.Next()`).

receiver/loadgenreceiver/config.go
```diff
@@ -25,11 +25,16 @@ type (
...
 
-// Config defines configuration for HostMetrics receiver.
+// Config defines configuration for loadgen receiver.
 type Config struct {
 	Metrics MetricsConfig `mapstructure:"metrics"`
 	Logs    LogsConfig    `mapstructure:"logs"`
 	Traces  TracesConfig  `mapstructure:"traces"`
+
+	// Concurrency is the amount of concurrency when sending to next consumer.
+	// The concurrent workers share the amount of workload, instead of multiplying the amount of workload,
```
carsonip, line 35: [to reviewer] I've made a design decision here for it to share the workload instead of multiplying the amount of workload. The implication is that when we use this config in otelbench, as concurrency increases, time/op will decrease. If it was implemented otherwise, the time/op will be equal or higher as concurrency increases.

receiver/loadgenreceiver/logs_test.go
```diff
@@ -32,26 +33,30 @@ import (
...
-	assert.NoError(t, err)
-	defer func() {
-		assert.NoError(t, r.Shutdown(context.Background()))
-	}()
-	stats := <-doneCh
-	want := maxReplay * bytes.Count(demoLogs, []byte("\n"))
-	assert.Equal(t, want, stats.Requests)
-	assert.Equal(t, want, len(sink.AllLogs()))
-	assert.Equal(t, sink.LogRecordCount(), stats.LogRecords)
+	for _, workers := range []int{1, 2} {
```
simitt, line 36: maybe stay consistent with terminology and also call it `concurrency`?

receiver/loadgenreceiver/internal/list/bounded_looping.go
```diff
@@ -0,0 +1,52 @@
...
+	return BoundedLoopingList[T]{
+		list:      LoopingList[T]{items: items},
+		loopLimit: loopLimit,
+	}
+}
+
+// Next returns the next element in the looping list.
+// When loop limit is reached, it returns a zero value of T and ErrLoopLimitReached.
+// Safe for concurrent use.
+func (s *BoundedLoopingList[T]) Next() (T, error) {
```
simitt, line 44: Why is the BoundedLoopingList made safe for concurrent use vs. the LoopingList isn't? Logically from the perspective of a user the difference between them is that one loops infinitely, whereas the other one doesn't - it's not clear to me why this would imply difference in safety for concurrent usage.

receiver/loadgenreceiver/internal/list/bounded_looping.go
```diff
@@ -0,0 +1,52 @@
...
+package list // import "github.com/elastic/opentelemetry-collector-components/receiver/loadgenreceiver/internal/list"
+
+import (
+	"errors"
+	"sync"
+)
+
+var ErrLoopLimitReached = errors.New("loop limit reached")
+
+// BoundedLoopingList is a list that can optionally have a limit on the number of loops performed.
```
simitt, line 27: There's a `LoopingList` and a `BoundedLoopingList` - why is the `loopLimit` optional?

receiver/loadgenreceiver/internal/list/bounded_looping.go
carsonip: Given the two are pretty much the same and it is not worth the extra complexity to separate them, I've combined them to reduce confusion.

receiver/loadgenreceiver/internal/list/bounded_looping.go
carsonip: ditto

loadgen/cmd/otelsoak/config.example.yaml
```diff
@@ -1,5 +1,6 @@
 receivers:
   loadgen:
+#    concurrency: 16
```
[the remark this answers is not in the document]
carsonip: Didn't have a good idea for a magic number. Defaulting to GOMAXPROCS now.

receiver/loadgenreceiver/factory.go
```diff
@@ -39,6 +41,7 @@ func createDefaultReceiverConfig(logsDone, metricsDone, tracesDone chan Stats) c
 		Traces: TracesConfig{
 			doneCh: tracesDone,
 		},
+		Concurrency: runtime.GOMAXPROCS(0),
```
simitt, line 44: 💯

receiver/loadgenreceiver/internal/list/looping.go
```diff
@@ -0,0 +1,64 @@
...
+type LoopingList[T any] struct {
+	items     []T
+	idx       int
+	loopCnt   int
+	loopLimit int
+	mu        sync.Mutex
+}
+
+// NewLoopingList returns a LoopingList over items for an optional loopLimit.
+// Setting loopLimit to 0 causes the list to loop infinitely.
```
simitt, line 37:
```suggestion
// Setting loopLimit to <= 0 causes the list to loop infinitely.
```

receiver/loadgenreceiver/internal/list/looping.go
```diff
@@ -0,0 +1,64 @@
...
+func NewLoopingList[T any](items []T, loopLimit int) *LoopingList[T] {
+	return &LoopingList[T]{
+		items:     items,
+		loopLimit: loopLimit,
+	}
+}
+
+// Next returns the next item in list with a nil error.
```
simitt, line 45:
```suggestion
// Next returns the next item in list with a nil error.
// If LoopingList limit is reached, it returns ErrLoopLimitReached.
```

receiver/loadgenreceiver/internal/list/looping.go
```diff
@@ -0,0 +1,64 @@
...
+// Safe for concurrent use.
+func (s *LoopingList[T]) Next() (T, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.loopCnt > 0 && s.loopCnt >= s.loopLimit {
```
simitt, line 51: I don't think this works if `s.loopLimit` is `<=0` this condition will always be true after the first loop, instead of looping infinitely.  

Instead should be:
```
if s.loopLimit > 0 && s.loopCnt >= s.loopLimit {
```

receiver/loadgenreceiver/internal/list/looping.go
```diff
@@ -0,0 +1,64 @@
...
+
+	item := s.items[s.idx]
+
+	s.idx = (s.idx + 1) % len(s.items)
+	if s.idx == 0 {
+		s.loopCnt++
+	}
+
+	return item, nil
+}
```
simitt, line 64: Maybe add some basic unit test for this loop behavior.

loadgen/cmd/otelsoak/config.example.yaml
```diff
@@ -1,5 +1,6 @@
 receivers:
   loadgen:
+#    concurrency: 16  # defaults to GOMAXPROCS
```
simitt, line 3: Out of curiosity, how is `otelsoak` configured to stop at some point? I don't see a possibility to configure a time or similar?

simitt (changes requested):
Left a comment for a small fix in the loop behavior

receiver/loadgenreceiver/factory.go
carsonip: I'm second guessing this, as apmsoak agent replicas default to 1. https://github.com/elastic/apm-perf/blob/main/internal/soaktest/runner.go#L100 . For otelbench use case it doesn't matter, as it should override this config to a multiplier of GOMAXPROCS anyway.

But for loadgenreceiver users e.g. otelsoak
- `1`: it is predictable, and it works roughly the same across environments. Added benefit for consistency with apmsoak.
- `GOMAXPROCS`: otelsoak will produce a rather different load across machines with different number of cores/threads, which could be a surprise to unsuspecting users. Also I'm not 100% sure if GOMAXPROCS is enough, or it needs to be N * GOMAXPROCS, similar to what we did in `make run-benchmark-autotuned`: https://github.com/elastic/apm-server/blob/3833927f8b1a17f4f0480af5d4a8ed051910a0ca/testing/benchmark/Makefile#L120
- no default: making it a required config in loadgenreceiver, and we specify it as 1 in the example config so that it is convenient to use in otelsoak OOTB.

Now I'm slightly inclined to do either `1` or `no default`, but no strong opinion on this, and happy to merge the PR with GOMAXPROCS to unblock it, and add a note that it is different from otelsoak.

receiver/loadgenreceiver/internal/list/looping.go
carsonip: :facepalm: it is a typo

receiver/loadgenreceiver/factory.go
simitt: You are making a good point about predictability and consistency; SGTM to keep at `concurrency: 1`.

loadgen/cmd/otelsoak/config.example.yaml
carsonip: It does not terminate now. I'm not aware of a way to notify the otel collector to stop, unless we write code to make it panic or os.Exit

simitt: ok, thanks for confirming; probably something for us to look into, but in another issue/PR

receiver/loadgenreceiver/internal/list/looping.go
carsonip: You caught me putting in an undefined config value (negative range) intentionally. Let me just validate loop limit to be non-negative, so that we can define it later.

simitt (approved):

rogercoll (approved):
Looks great!

PR #324, opened 2025-01-21
[receiver/elasticapm] Add /config/v1/agents endpoint

rogercoll:
## Open questions

- ~~Should the Remote Agent Configuration feature be optional? What should the `/config/v1/agents` endpoint return if `elasticsearch` endpoint is not configured?~~ This PR has configured the endpoint to return an error and StatusForbidden error code. See https://github.com/elastic/apm/blob/main/specs/agents/configuration.md#dealing-with-errors

- ~~`elasticapm::elastic` configuration is a copy&paste of the upstream `elasticsearch exporter` configuration, should we work on unifying them?~~

## Testing

- Unit tests mocking the fetcher
- Integration tests with Elasticsearch 8.17 → CI modified to run them

##  Sample collector configuration

```yaml
extensions:
  basicauth:
    client_auth:
      username: changeme
      password: changeme

receivers:
  elasticapm:
    elasticsearch:
      endpoint: "https://127.0.0.1:9200"
      tls:
        ca_file: path_to_ca-cert.pem
      auth:
        authenticator: basicauth

processors:
exporters:
  debug:

service:
  extensions: [basicauth]
  pipelines:
    logs:
      exporters: [debug]
      processors:
      receivers: [elasticapm]
    traces:
      exporters: [debug]
      processors:
      receivers: [elasticapm]
    metrics:
      exporters: [debug]
      processors:
      receivers: [elasticapm]

```


carsonip:
> Should the Remote Agent Configuration feature be optional? What should the /config/v1/agents endpoint return if elasticsearch endpoint is not configured? This PR has configured the endpoint to return an error and StatusBadRequest error code.

Are you aware of this agent config spec? "Dealing with errors" section may be relevant. https://github.com/elastic/apm/blob/main/specs/agents/configuration.md

rogercoll:
> > Should the Remote Agent Configuration feature be optional? What should the /config/v1/agents endpoint return if elasticsearch endpoint is not configured? This PR has configured the endpoint to return an error and StatusBadRequest error code.
> 
> Are you aware of this agent config spec? "Dealing with errors" section may be relevant. https://github.com/elastic/apm/blob/main/specs/agents/configuration.md

That section answered my question, thank you!

receiver/elasticapmreceiver/factory.go
```diff
@@ -46,10 +49,16 @@ func NewFactory() receiver.Factory {
...
+	// defaultServerConfig := confighttp.NewDefaultServerConfig()
+	defaultServerConfig := confighttp.ServerConfig{}
 	defaultServerConfig.Endpoint = defaultEndpoint
+	defaultESClientConfig := configelasticsearch.NewDefaultClientConfig()
 	return &Config{
-		ServerConfig: defaultServerConfig,
+		ServerConfig:  defaultServerConfig,
+		Elasticsearch: &defaultESClientConfig,
+		// based on apm-server default https://github.com/elastic/apm-server/blob/main/internal/beater/config/agentconfig.go#L101
+		CacheDuration: 30 * time.Second,
```
gregkalapos, line 61: We chatted about this via slack, just raising here as well.

One potential issue here is that now the receiver will have a config (from what I understand it's also required), which points to elasticsearch. So it'd be something like this (as @rogercoll showed me a sample config):

```
receivers:
  elasticapm:
    elasticsearch:
      endpoint: "https://127.0.0.1:9200"
      tls:
        ca_file: [...somepath]
      auth:
        authenticator: basicauth
```

In order to fetch the config, the receiver needs an es endpoint, so it needs to configured, we clearly need this.

On the other hand, from a users's point of view, who may not even use central config, this can be very confusing - we configure a receiver here and then we force users to point the receiver to elasticsearch.
Also, users will have an elasticsearch exporter, typically pointing to the exact same cluster. I'm afraid this could confuse users.

@rogercoll said:

> ...one option that we could research is to create an extension for creating es clients. Similar to the basicauth shared above, so you just need to specify the cluster details in a config section

But overall I'd not block the PR on this, maybe worth a follow-up.

axw: Yeah I wouldn't use `agent.config`. I think `agent_config` looks good.

receiver/elasticapmreceiver/internal/testutil/testutil.go
```diff
@@ -0,0 +1,135 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
rogercoll, line 1: Copy of https://github.com/open-telemetry/opentelemetry-collector/blob/main/internal/testutil/testutil.go

gregkalapos (approved):

PR #346, opened 2025-02-05
[processor/lsminterval] Improve batching performance by using merger.Value

lahsivjar:
The PR improves the batching performed by the lsminterval processor. With this, there is no need to have batch processor before lsminterval processor.

Related to https://github.com/elastic/opentelemetry-collector-components/issues/245 and https://github.com/elastic/opentelemetry-collector-components/issues/211

<details>
<summary>Benchmark with main using defer merge logic (current PR)</summary>

```
name                                                        old time/op    new time/op    delta
pkg:github.com/elastic/opentelemetry-collector-components/processor/lsmintervalprocessor goos:darwin goarch:arm64
Aggregation/sum_cumulative                                    20.6µs ± 4%    21.7µs ±14%     ~     (p=0.310 n=5+5)
Aggregation/sum_cumulative-10                                 11.7µs ± 1%    10.2µs ± 7%  -13.24%  (p=0.008 n=5+5)
Aggregation/sum_delta                                         21.6µs ± 4%    22.2µs ± 7%     ~     (p=0.310 n=5+5)
Aggregation/sum_delta-10                                      12.0µs ± 3%     9.9µs ± 2%  -17.67%  (p=0.008 n=5+5)
Aggregation/histogram_cumulative                              20.4µs ± 3%    21.9µs ± 7%   +7.51%  (p=0.032 n=5+5)
Aggregation/histogram_cumulative-10                           12.7µs ± 5%    11.4µs ± 5%   -9.63%  (p=0.008 n=5+5)
Aggregation/histogram_delta                                   22.1µs ±10%    21.6µs ± 5%     ~     (p=0.690 n=5+5)
Aggregation/histogram_delta-10                                12.2µs ± 3%    11.3µs ± 4%   -7.42%  (p=0.008 n=5+5)
Aggregation/exphistogram_cumulative                           22.6µs ±10%    22.3µs ±14%     ~     (p=0.841 n=5+5)
Aggregation/exphistogram_cumulative-10                        12.8µs ± 5%    11.1µs ± 6%  -12.99%  (p=0.008 n=5+5)
Aggregation/exphistogram_delta                                22.4µs ± 9%    23.1µs ± 9%     ~     (p=0.548 n=5+5)
Aggregation/exphistogram_delta-10                             12.9µs ± 3%    10.8µs ± 2%  -15.86%  (p=0.008 n=5+5)
Aggregation/summary_enabled                                   24.0µs ±11%    24.0µs ± 8%     ~     (p=0.841 n=5+5)
[...]

Issue #245, opened 2024-12-13
[processor/lsminterval] Optimize hot path


lahsivjar:
Current logic creates new lookup maps in hot path for all `pmetric.Metric` passed to the `ConsumeMetrics`. This is a bit wasteful in terms of resource usage (more allocations). We can either:

1. Defer merge of this to during merge operations(compaction). We can also make the component mutating and just consume the whole metrics without any merges at this level. OR,
2. Optimize this to not use the map in a wasteful way. This could also be done by introducing pooling of values.

raultorrecilla:
moving it from it107 to it108

raultorrecilla:
moving it from it108 to it109

lahsivjar:
Goals of this issue were achieved by https://github.com/elastic/opentelemetry-collector-components/issues/459 and https://github.com/elastic/opentelemetry-collector-components/pull/290. Closing this issue now.

Issue #211, opened 2024-10-24
[processor/lsminterval] Optimize memory usage for each aggregation interval


lahsivjar:
As per the current logic, the processor uses aggregation interval and processing time as the key in the LSM database. On harvest, each key is decoded and loaded into memory. Due to this, all of a specific aggregation interval will be loaded into memory -- this piece of logic is also required to do overflows during merge as we will require ALL of the data to be available at once.

Above said, we could reduce the memory usage by relaxing the overflow during merge requirement and partitioning the metrics. We do something similar in apm-aggregation ([ref](https://github.com/elastic/apm-aggregation/blob/b22aefd839f169457eae885fda36fd68ab56a902/aggregators/converter.go#L205-L215)).

lahsivjar:
Compared to apm-aggregation, the performance of the merge operations using the opentelemetry data model is too expensive. As per some of the benchmarking, the biggest culprit is merge operations performed for exponential histograms (we use a copy of the contrib implementation). Here are some benchmarking results:

<img width="1915" alt="Image" src="https://github.com/user-attachments/assets/52b67a64-9f2c-4634-beba-1c4e1e39e62d" />
<img width="1124" alt="Image" src="https://github.com/user-attachments/assets/4a57d017-bd34-4272-a7e4-693b6c37d591" />

Playing around with GOGC and GOMEMLIMIT does produce some positive results but as a whole other performance is still marred by the huge number of allocations required for the merge operation as a whole. Note that merge is the biggest culprit but not the only one.

lahsivjar:
Diving into the source of high allocations, I ran an experiment to replace exponential histogram with histogram. While the number of buckets for the histogram was kept to the default value (which is quite less than exponential histogram default) the observed GC was much lower and the throughput was stable and in a good range:

<img width="1913" alt="Image" src="https://github.com/user-attachments/assets/e2ae68ef-0afc-416c-a5bb-7cb0e9e4094b" />

(Above was ran with go mem limit set to 4GB and normal gc disabled i.e. `GOGC=off GOMEMLIMIT=4GiB`)

[...]

PR #350, opened 2025-02-05
[processor/lsminterval]Add parallel benchmark for aggregation

lahsivjar:
Related to https://github.com/elastic/opentelemetry-collector-components/issues/245,  https://github.com/elastic/opentelemetry-collector-components/issues/211, and #346

### DIFF
--- a/processor/lsmintervalprocessor/processor_test.go
+++ b/processor/lsmintervalprocessor/processor_test.go
@@ -20,6 +20,7 @@ package lsmintervalprocessor
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"path/filepath"
 	"testing"
 	"time"
@@ -292,7 +293,11 @@ func TestClientMetadata(t *testing.T) {
 }
 
 func BenchmarkAggregation(b *testing.B) {
-	benchmarkAggregation(b, nil)
+	benchmarkAggregation(b, nil, false)
+}
+
+func BenchmarkAggregationParallel(b *testing.B) {
+	benchmarkAggregation(b, nil, true)
 }
 
 func BenchmarkAggregationWithOTTL(b *testing.B) {
@@ -301,10 +306,10 @@ func BenchmarkAggregationWithOTTL(b *testing.B) {
 		`set(instrumentation_scope.attributes["custom_scope_attr"], "scope")`,
 		`set(attributes["custom_dp_attr"], "dp")`,
 		`set(resource.attributes["dependent_attr"], Concat([attributes["aaa"], "dependent"], "-"))`,
-	})
+	}, false)
 }
 
-func benchmarkAggregation(b *testing.B, ottlStatements []string) {
+func benchmarkAggregation(b *testing.B, ottlStatements []string, runParallel bool) {
 	testCases := []struct {
 		name        string
 		passThrough bool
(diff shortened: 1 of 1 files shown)


processor/lsmintervalprocessor/processor_test.go
```diff
@@ -352,19 +357,35 @@ func benchmarkAggregation(b *testing.B, ottlStatements []string) {
 			md, err := golden.ReadMetrics(filepath.Join(dir, "input.yaml"))
 			require.NoError(b, err)
 			md.MarkReadOnly()
-			b.ResetTimer()
-
-			err = mgp.Start(context.Background(), componenttest.NewNopHost())
-			require.NoError(b, err)
-			for i := 0; i < b.N; i++ {
+			benchFunc := func() {
+				b.StopTimer()
```
lahsivjar, line 361: [**For reviewers**] Not a fan of the copying the pmetric and then adding things to resource attribute but I didn't want to make a lot of change in the benchmark and I also didn't have any other ideas.

processor/lsmintervalprocessor/processor_test.go
```diff
@@ -352,19 +357,35 @@ func benchmarkAggregation(b *testing.B, ottlStatements []string) {
...
 			}
 
+			b.ResetTimer()
+			err = mgp.Start(context.Background(), componenttest.NewNopHost())
+			require.NoError(b, err)
+
+			if runParallel {
+				b.RunParallel(func(pb *testing.PB) {
+					for pb.Next() {
+						benchFunc()
```
carsonip, line 380: https://pkg.go.dev/testing#B.RunParallel
> The body function will be run in each goroutine. It should set up any goroutine-local state and then iterate until pb.Next returns false. It should not use the [B.StartTimer](https://pkg.go.dev/testing#B.StartTimer), [B.StopTimer](https://pkg.go.dev/testing#B.StopTimer), or [B.ResetTimer](https://pkg.go.dev/testing#B.ResetTimer) functions, because they have global effect. It should also not call [B.Run](https://pkg.go.dev/testing#B.Run).

I'm afraid calling b.StopTimer in a PB won't work

carsonip: might have to bite the bullet and not stop the timer for PB

processor/lsmintervalprocessor/processor_test.go
```diff
@@ -352,19 +357,35 @@ func benchmarkAggregation(b *testing.B, ottlStatements []string) {
...
-			require.NoError(b, err)
-			for i := 0; i < b.N; i++ {
+			benchFunc := func() {
+				b.StopTimer()
+				// Copy the metric as the ConsumeMetric call is mutating
 				mdCopy := pmetric.NewMetrics()
 				md.CopyTo(mdCopy)
 				// Overwrites the asdf attribute in metrics such that it becomes high cardinality
-				mdCopy.ResourceMetrics().At(0).Resource().Attributes().PutStr("asdf", fmt.Sprintf("%d", i))
+				mdCopy.ResourceMetrics().At(0).Resource().Attributes().PutStr("asdf", fmt.Sprintf("rand_%d", rand.Int()))
```
axw, line 366: Should each goroutine get its own `*rand.Rand`? The global one is synchronised, so this may artificially limit parallelism.

Alternatively, perhaps generate a random "ID" for each goroutine and pass that in, and add to the iteration to get a unique resource ID.

processor/lsmintervalprocessor/processor_test.go
```diff
@@ -292,7 +293,11 @@ func TestClientMetadata(t *testing.T) {
 }
 
 func BenchmarkAggregation(b *testing.B) {
-	benchmarkAggregation(b, nil)
+	benchmarkAggregation(b, nil, false)
+}
+
+func BenchmarkAggregationParallel(b *testing.B) {
+	benchmarkAggregation(b, nil, true)
 }
```
axw, line 301: Do we really need two benchmarks? If you want to test without parallelism, you can run with `-cpu=1`

processor/lsmintervalprocessor/processor_test.go
```diff
@@ -352,19 +357,35 @@ func benchmarkAggregation(b *testing.B, ottlStatements []string) {
...
+				b.StopTimer()
+				// Copy the metric as the ConsumeMetric call is mutating
 				mdCopy := pmetric.NewMetrics()
 				md.CopyTo(mdCopy)
 				// Overwrites the asdf attribute in metrics such that it becomes high cardinality
-				mdCopy.ResourceMetrics().At(0).Resource().Attributes().PutStr("asdf", fmt.Sprintf("%d", i))
+				mdCopy.ResourceMetrics().At(0).Resource().Attributes().PutStr("asdf", fmt.Sprintf("rand_%d", rand.Int()))
+				b.StartTimer()
+
 				err = mgp.ConsumeMetrics(ctx, mdCopy)
```
axw, line 369: this now updates an error shared by all the goroutines (try running with -race)

axw:
@lahsivjar in the interest of time, I've pushed some changes - feel free to revert though

axw (approved):

processor/lsmintervalprocessor/processor_test.go
lahsivjar: 🤦

PR #352, opened 2025-02-07
lsmintervalprocessor: make processor non-mutating

axw:
The processor is no longer mutating, and will instead create a new pmetric.Metrics with just the metrics that will not be aggregated. This should be more efficient in the case where no or few metrics passed through.

This also allows us to remove some memory allocations from the benchmark code since it no longer needs to copy the input for each iteration.


carsonip (approved):

axw:
govulncheck stuff is unrelated

PR #355, opened 2025-02-13
test: add testutil internal package

rogercoll:
Exposes and internal function to get a local endpoint: `GetAvailableLocalAddress`. Ipv4 code from https://github.com/open-telemetry/opentelemetry-collector/tree/main/internal/testutil

Used at:
- https://github.com/elastic/opentelemetry-collector-components/pull/324
- https://github.com/elastic/opentelemetry-collector-components/pull/39

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -169,7 +169,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: ["1.22.12", "1.23.1"] # 1.20 is interpreted as 1.2 without quotes
+        go-version: ["1.22.12", "1.23.6"] # 1.20 is interpreted as 1.2 without quotes
         runner: [ubuntu-latest]
         group:
           - all
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.22.11"
+          go-version: "1.22.12"
           cache: false
       - name: Cache Go
         id: go-cache
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.22.11"
+          go-version: "1.22.12"
           cache: false
       - name: Cache Go
         id: go-cache
@@ -107,7 +107,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.22.11"
+          go-version: "1.22.12"
           cache: false
       - name: Cache Go
         id: go-cache
@@ -130,7 +130,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.22.11"
+          go-version: "1.22.12"
           cache: false
       - name: Cache Go
         id: go-cache
@@ -169,7 +169,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: ["1.22.11", "1.23.1"] # 1.20 is interpreted as 1.2 without quotes
+        go-version: ["1.22.12", "1.23.1"] # 1.20 is interpreted as 1.2 without quotes
         runner: [ubuntu-latest]
         group:
           - all
--- a/internal/testutil/go.mod
+++ b/internal/testutil/go.mod
@@ -1,6 +1,6 @@
 module github.com/elastic/opentelemetry-collector-components/internal/testutil
 
-go 1.23.4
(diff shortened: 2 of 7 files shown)


.github/workflows/checks.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.22.11"
+          go-version: "1.22.12"
```
florianl, line 24: Instead of hard coding a Go version and bumping it with every patch release, would it possible to just take the Go version from `go.mod`?
```suggestion
         go-version-file: /path/to/go.mod
```

internal/testutil/testutil_test.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetAvailableLocalAddress(t *testing.T) {
+	endpoint := GetAvailableLocalAddress(t)
```
florianl, line 29: This looks like a race condition to me.
`GetAvailableLocalAddress()` calls `findAvailable()`, which calls `findAvailableAddress()`. `findAvailableAddress()` opens a generic TCP listener and closes it. The host and port part of this then closed listener is returned and in the next line of `TestGetAvailableLocalAddress()` a new listener is opened on this host/port combination.

A possible race condition can happen between the closing of the listener in `findAvailableAddress()` and before `GetAvailableLocalAddress()` returns and a new listener is opened on the returned endpoint.

internal/testutil/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/elastic/opentelemetry-collector-components/internal/testutil
+
+go 1.22.3
```
florianl, line 3: Use the latest patched version of this minor version:

```suggestion
go 1.22.12
```

.github/workflows/checks.yml
```diff
@@ -169,7 +169,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: ["1.22.11", "1.23.1"] # 1.20 is interpreted as 1.2 without quotes
+        go-version: ["1.22.12", "1.23.1"] # 1.20 is interpreted as 1.2 without quotes
```
florianl, line 172: Use the latest patched version of this minor version:


```suggestion
        go-version: ["1.22.12", "1.23.6"] # 1.20 is interpreted as 1.2 without quotes
```

.github/workflows/checks.yml
rogercoll: That would be easier to maintain, but this repository is a Go monorepo without a central Go modules file. Maybe referencing one of the internal packages would be fine (e.g `internal/tools/go.mod`). wdyt?

florianl: It's just something I have noticed while reading this change. I'm not aware of the Go version bump policy of this project. If the Go version in all go.mods is bumped at the same time, any one of the `go.mod`s could be referenced. 
So overall, I don't have a personal preference.

internal/testutil/testutil_test.go
rogercoll: Nice catch, I just noticed this comment in the `findAvailableAddress` :

```
// There is a possible race if something else takes this same port before
// the test uses it, however, that is unlikely in practice.
```

Although we could add some sync protection, the `GetAvailableLocalAddress` function will be used across multiple packages and as it does not return a listener, there can be a race condition at the moment of its usage.

The upstream collector has been using this helper for a while (and many more usages) without a noticeable collision. Adding a reference to the upstream code in case it evolves.

.github/workflows/checks.yml
rogercoll: >  If the Go version in all go.mods is bumped at the same time, any one of the go.mods could be referenced.

At the moment, each submodule is bumped separately. I would prefer to wait until we have a `go.mod` file in the root directory that we can reference as base required Go version. Also, some CI checks run for two pinned Go versions to assert compatibility between the latest two minor versions.

rockdaboot (approved):
The upstream test code deserves a rewrite 😅 
Approving to unblock further work.

internal/testutil/testutil_test.go
gregkalapos: `GetAvailableLocalAddress` is only intended for tests, right? So the intended usage is just to get an available address for tests, if the race condition happens, the only consequence is that the test will fail, because the address may be taken when in the test we'd really use it. Or I'm I mistaken?

gregkalapos (approved):
Looks ok to me. The discussion in https://github.com/elastic/opentelemetry-collector-components/pull/355/files#r1954018530 is worth having, but since upstream works with this, I think this is ok to merge.

PR #356, opened 2025-02-18
Update to Go 1.23

axw:
Require Go 1.23, test with Go 1.24.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.22.12"
+          go-version: "1.23.6"
           cache: false
       - name: Cache Go
         id: go-cache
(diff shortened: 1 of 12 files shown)


axw:
This is necessary to update https://github.com/axiomhq/hyperloglog/ in lsmintervalprocessor. We could bump up just that module and the tools module, but I figured we may as well keep them all in sync.

marclop (approved):

.github/workflows/checks.yml
```diff
@@ -169,7 +169,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: ["1.22.12", "1.23.6"] # 1.20 is interpreted as 1.2 without quotes
+        go-version: ["1.23.6", "1.24.0"] # 1.20 is interpreted as 1.2 without quotes
```
florianl, line 172: Use `x` for the patch version to always use the latest one.
```suggestion
        go-version: ["1.23.x", "1.24.x"] # 1.20 is interpreted as 1.2 without quotes
```

florianl (approved):

.github/workflows/checks.yml
axw: Good call, thanks!

lahsivjar (approved):

Issue #336, opened 2025-01-29
Create a new connector component that encapsulates APM-related signaltometrics config


axw:
We should create a new connector that encapsulates this config: https://github.com/elastic/opentelemetry/blob/ccb569542dc3a381461dca58769f8c1d2149c474/resources/kubernetes/operator/helm/values.yaml#L259-L492

This will avoid users having to copy & paste the config, and encounter drift. We may provide some extension points in the config, e.g. to add more resource/signal attributes to include in aggregations.

lahsivjar:
I have discussed this with @simitt a bit and I think that while encapsulating complexity would be a good idea, we should also think about what we want to expose to the users as configuration. Wrapping configuration in a component would get quite involved based on what we expect from the customers to modify, for example, if we want the customer to be able to modify existing resource attributes used for generating transaction metrics then we would have to build a configuration to modify the existing configuration. OTOH, if we want the users to be able to add new signal to metrics configuration then we would want to expose the configuration offered by the signal to metrics component.

Basically, my point here is that abstracting complexity in itself sounds good but based on what knobs we want to offer to the customer for customizing the config this might get into the territory of more involved accidental complexity and defeat our purpose of a simpler interface for getting started.

axw:
@lahsivjar fair points. How would you suggest we proceed? I had in mind that we would start with a component that encapsulates the configuration, and provides no configuration settings at all. Then we would add configuration after we know more about what needs to be configurable.

lahsivjar:
If we want to start with no configuration then should we just have some form of templating to download the configuration files from somewhere while building the collector rather than wrapping it into a single component? - I am not against the component but just wondering if we should do it when we have the requirements of what we would want to expose to the customers a bit more clearer.

lahsivjar:
@axw and I talked on zoom and I am good with starting with a zero configuration component wrapper wrapping the signal to metrics connector.

felixbarny:
Generally, the component should offer a similar layer of abstraction for configuration as APM Server. The fact that this is using signaltometrics, batch, and lsminterval under the hood should be an implementation detail.

lahsivjar:
@felixbarny Do you mean we should wrap the whole pipeline for apm aggregations? While I love the idea, I am not sure if we can wrap more than one component in another component.

lahsivjar:
[...]

PR #367, opened 2025-02-19
Introduce elasticapmconnector 

axw:
This is a new connector that encapsulates:
 - signaltometricsconnector with an Elastic APM-specific configuration
 - lsmintervalprocessor with an Elastic APM-specific configuration

The intention behind this is to simplify EDOT Collector configuration so users have less configuration to deal with, and less opportunity for configuration drift and breaking assumptions.

Closes https://github.com/elastic/opentelemetry-collector-components/issues/336


connector/elasticapmconnector/config.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+func (cfg Config) signaltometricsConfig() *signaltometricsconfig.Config {
+	// serviceResourceAttributes is the resource attributes included in
+	// service-level aggregated metrics.
+	serviceResourceAttributes := []signaltometricsconfig.Attribute{
+		{Key: "service.name"},
+		{Key: "deployment.environment"}, // service.environment
+		{Key: "telemetry.sdk.language"}, // service.language.name
+		{Key: "agent.name"},             // set via elastictraceprocessor
```
felixbarny, line 99: We should set a default value like `unknown` as logs and metrics don't go through the `elastictraceprocessor`. See also the discussion we had in the context of the EDOT k8s gateway mode: https://github.com/elastic/elastic-agent/pull/6444#discussion_r1925699196.

I'm so glad that we can now centralize the configuration to make sure it's consistent everywhere.

lahsivjar (approved):
LGTM! Just some minor comments for updating the configuration to be directly compatible with the upstream one.

connector/elasticapmconnector/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+[Exporter Pipeline Type]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/connector/README.md#exporter-pipeline-type
+[Receiver Pipeline Type]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/connector/README.md#receiver-pipeline-type
+[Stability Level]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#stability-levels
+<!-- end autogenerated section -->
+
+## Configuration
+
+By design, this component has minimal configuration. An empty configuration will be functional
+though we would not recommend this for production workloads.
```
simitt, line 30: Could this be more clear about what the recommended configuration is?

felixbarny:
Regarding the discussion we had about exponential histograms vs explicit bucket histograms I'm wondering if we should expose this as a config option. Users would be able to set their preference on exponential/explicit as well as the explicit buckets (we'd ship with a default). That's something we can add later as a config option as well. Also, we'll probably only want to allow for exponential histograms to be configured once we've stabilized the performance for them. But even when we do, users may prefer an explicit bounds histogram for performance reasons. Another option we could add for exponential histograms is the max bucket size (and default to 160).

connector/elasticapmconnector/config.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	// transactionResourceAttributes is the resource attributes included
+	// in transaction group-level aggregationed metrics.
```
christos68k, line 103:
```suggestion
	// in transaction group-level aggregated metrics.
```

connector/elasticapmconnector/connector.go
```diff
@@ -0,0 +1,81 @@
...
+	lsminterval processor.Metrics
+}
+
+func newElasticAPMConnector(
+	ctx context.Context,
+	cfg *Config,
+	set connector.Settings,
+	nextConsumer consumer.Metrics,
+) (*elasticapmConnector, error) {
+	c := &elasticapmConnector{cfg: cfg, set: set}
```
christos68k, line 49: Nit: Can move this after `lsminterval` is created to make the error case more obvious

connector/elasticapmconnector/doc.go
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//go:generate mdatagen metadata.yaml
+
+// Package elasticapmconnector provides a connector for performing Elastic APM-specific enrichment preaggregation.
```
christos68k, line 20:
```suggestion
// Package elasticapmconnector provides a connector for performing Elastic APM-specific enrichment pre-aggregation.
```

christos68k (approved):

connector/elasticapmconnector/config.go
```diff
@@ -0,0 +1,282 @@
...
+		}, {
+			Name:                      "transaction.duration.histogram",
+			Description:               "APM service transaction aggregated metrics as histogram",
+			IncludeResourceAttributes: serviceResourceAttributes,
+			Attributes: append(serviceTransactionAttributes[:], signaltometricsconfig.Attribute{
+				Key:          "elasticsearch.mapping.hints",
+				DefaultValue: []any{"_doc_count"},
+			}),
+			Unit: "us",
+			ExponentialHistogram: &signaltometricsconfig.ExponentialHistogram{
```
vigneshshanmugam, line 184: Lets default to histogram based buckets for now - Results looks promising and we need this in time for EDOT GA https://github.com/elastic/hosted-otel-collector/issues/406#issuecomment-2669574786 Thoughts?

connector/elasticapmconnector/config.go
axw: Hmmm so I will set it, but I think just defaulting to `unknown` will be a problem too: that means metrics and traces coming from the same client will result in different aggregated service metrics. We might need to move the resource attribute enrichment logic out of `elastictrace` to a separate processor.

axw: Alternatively (preferably) we update the UI to stop expecting `agent.name`, and have it work with either `agent.name` or the `telemetry.sdk.*` SemConv attributes instead.

connector/elasticapmconnector/config.go
```diff
@@ -0,0 +1,282 @@
...
+
+func (cfg Config) lsmConfig() *lsmconfig.Config {
+	lsmConfig := &lsmconfig.Config{
+		Intervals: []lsmconfig.IntervalConfig{{
+			Duration: time.Minute,
+			Statements: []string{
+				// TODO(axw) confirm that it's ok to change metricset.interval
+				// from a resource attribute to datapoint attribute. Resource is
+				// meant to be about the producer of the signal, I don't think
+				// this fits.
```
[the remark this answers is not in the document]
axw: Thanks for confirming!

connector/elasticapmconnector/config.go
```diff
@@ -0,0 +1,282 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package elasticapmconnector // import "github.com/elastic/opentelemetry-collector-components/connector/elasticapmconnector"
+
+import (
+	"time"
+
+	signaltometricsconfig "github.com/elastic/opentelemetry-collector-components/connector/signaltometricsconnector/config"
```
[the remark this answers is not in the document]
axw: Created an issue: https://github.com/elastic/opentelemetry-collector-components/issues/369

connector/elasticapmconnector/config.go
axw: :man_facepalming: :joy: thanks

connector/elasticapmconnector/README.md
axw: :+1: will add a sentence recommending storing on disk

axw:
@felixbarny I'd rather defer on adding configuration for now. I'll switch to explicit bucket now, and we can enable exponential once we've optimised it. If it's inherently very expensive to merge or convert, then we could consider configuration.

connector/elasticapmconnector/config.go
axw: Done.

PR #388, opened 2025-02-25
IntakeV2 receiver - Fix [span|transaction].duration.us and timestamp.us attributes

gregkalapos:
Adding missing conversion to attributes:
- `span.duration.us`
- `transaction..duration.us`
- timestamp.us

Noticed that UI uses these values and it seem to expect microsecond values (as the attribute name suggests). Prior to this PR, the UI looked like this:

Screenshot 2025-02-25 at 14 43 45

The PR fixes this.

Note that we in `receiver.go` we call `SetStartTimestamp` and SetEndTimestamp` on the span - that was fine before, just these attributes had the wrong value (1000x more due to the missing conversion).

### DIFF
--- a/receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
+++ b/receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
@@ -41,14 +41,14 @@ func SetDerivedFieldsForTransaction(event *modelpb.APMEvent, attributes pcommon.
 	attributes.PutBool("transaction.root", event.ParentId == "")
 	attributes.PutStr("transaction.type", event.Transaction.Type)
 	attributes.PutStr("transaction.result", event.Transaction.Result)
-	attributes.PutInt("transaction.duration.us", int64(event.Event.Duration))
+	attributes.PutInt("transaction.duration.us", int64(event.Event.Duration/1_000))
 }
 
 // Sets fields that are NOT part of OTel for spans. These fields are derived by the Enrichment lib in case of OTLP input
 func SetDerivedFieldsForSpan(event *modelpb.APMEvent, attributes pcommon.Map) {
 
 	attributes.PutStr("processor.event", "span")
-	attributes.PutInt("span.duration.us", int64(event.Event.Duration))
+	attributes.PutInt("span.duration.us", int64(event.Event.Duration/1_000))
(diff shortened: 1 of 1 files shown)


receiver/elasticapmreceiver/testdata/invalid_ids_expected.yaml
```diff
@@ -55,7 +55,7 @@ resourceSpans:
                   stringValue: ""
               - key: transaction.duration.us
                 value:
-                  intValue: "32592981"
+                  intValue: "32592"
```
gregkalapos, line 58: Note there is no real data loss here.

`startTimeUnixNano` and `endTimeUnixNano` are unchanged - this is only for this attribute to satisfy Kibana requirements. Same for other instances in the `yaml` files.

receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
```diff
@@ -41,14 +41,14 @@ func SetDerivedFieldsForTransaction(event *modelpb.APMEvent, attributes pcommon.
 	attributes.PutBool("transaction.root", event.ParentId == "")
 	attributes.PutStr("transaction.type", event.Transaction.Type)
 	attributes.PutStr("transaction.result", event.Transaction.Result)
-	attributes.PutInt("transaction.duration.us", int64(event.Event.Duration))
+	attributes.PutInt("transaction.duration.us", int64(event.Event.Duration/1_000))
```
dmathieu, line 44: As you mention in the PR description, the attribute name is misleading here, since it should be `.ms`?
Is it possible kibana would accept `.ms`?

gregkalapos: Unfortunately Kibana [expects `us`](https://github.com/search?q=repo%3Aelastic%2Fkibana+transaction.duration.us&type=code); [`ms`](https://github.com/search?q=repo%3Aelastic%2Fkibana+transaction.duration.ms&type=code) seems to be unknown to Kibana. 

From what I know `us` is a common abbreviation for `μs` in case of microseconds. [Wikipedia is](https://en.wikipedia.org/wiki/Microsecond) also aligned with that.

dmathieu (approved):

PR #394, opened 2025-02-27
[loadgen/otelbench] Add changelog workflow

constanca-m:
Relates to https://github.com/elastic/hosted-otel-collector/issues/486.

The goal is to have `otelbench` as a docker image. This way, when we run the daily benchmarks for the MOtel projects, we can use the docker image in elastic container registry, instead of having to build it on the workflow, and maybe potentially cloning the code, like it is already happening (inefficiently) for the otelsoak.

To have a workflow that triggers a docker release we need:

1. A workflow that first checks there is a new changelog entry for otelbench. This is what this PR is doing. It validates that everything is as expected.
2. A new workflow that will get triggered by an update on version:
   - Still need to research what is the best way to update version: manual (by checking for example the VERSION field in the MAKEFILE), or follow some other process (on a schedule, for example).
   - Update the CHANGELOG.md, and delete the changelog fragments.
   - Build the otelbench image and publish it to the docker elastic registry. It will use as a tag the version that we have in the Makefile.

If there is any change to any `*.go` files in the `otelbench` directory, then it is mandatory to add a changelog entry (unless the change is done by dependabot or renovate). 

The changelog tool is the same as the one used in `open-telemetry-contrib`. Find the [source](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/chloggen) here.


loadgen/cmd/otelbench/README.md
```diff
@@ -129,3 +129,19 @@ It is possible to run with a customized config to avoid passing in command line
...
+1. Create a changelog entry by running `make chlog-new`.
+2. Validate your changelog entry by running `make chlog-validate`.
+3. If you want to preview the future `CHANGELOG.md` you can also run `make chlog-preview`.
+
+**Second step**: TODO
+<!-----
+Have a workflow that gets triggered by the changelog_otelbench,
+updates the CHANGELOG.md and pushes the image to elastic container
+registry.
+---->
```
constanca-m, line 146: I'm leaving this step to a different PR so we can have a first run immediately to see if the changelogs work.

loadgen/cmd/otelbench/README.md
```diff
@@ -129,3 +129,20 @@ It is possible to run with a customized config to avoid passing in command line
...
+2. Validate your changelog entry by running `make chlog-validate`.
+3. If you want to preview the future `CHANGELOG.md` you can also run `make chlog-preview`.
+
+## Create new release
+
+TODO
+<!-----
+Have a workflow that gets triggered by a version update (maybe the version from the MAKEFILE, maybe another option), and runs `$(CHLOGGEN) update --version $(VERSION)`, updating the CHANGELOG.md and deleting all changelog fragments + it pushes a new image to elastic container
+registry.
+---->
```
constanca-m, line 147: I am leaving this part to a new PR.

vigneshshanmugam (commented):
Changes LGTM, But I don't have as strong opinion on changelog based workflow for publishing new images, we could control the new image generation via manual GH releases given this is a internal tool. 

I ll leave it up to you and @carsonip to decide.

constanca-m:
I agree @vigneshshanmugam, that's why I mention that maybe the workflow to trigger the release will have maybe to be manualy triggered (or by updating the version field). This PR now is only for the CHANGELOG actions. The CHANGELOG is needed because if we end up having a few versions on the elastic registry, we need to have a way to know what is different on them from the previous versions. Otherwise, who is to say what version we need to pick? That's why I think it is a good idea to enforce the changelog, before starting with the releases.

gizas:
As long as this PR does not handle the actual image creation, also lgtm! 
Maybe alternatively, we can use releases and tags (see relevant [github action](https://github.com/marketplace/actions/create-a-release-in-a-github-action)). So to handle on each tag the notes and from now on to use the tags as reference if makes sense.

constanca-m:
> Maybe alternatively, we can use releases and tags

I don't think this would be a good idea, only for the fact that this repository contains other components. And maybe we want to release a new image right after `otelbench` changes, regardless of whatever else is in the repository.

I think I will have an upcoming workflow that will be triggered either: manually, or by changing the version on makefile:
```
on:
  workflow_dispatch:
    inputs:
      version:
        description: Version of the otelbench image to release.
        required: true
  push:
    branches:
      - main
    paths:
      - ./loadgen/cmd/otelbench/Makefile
```

Then we can either make a release by updating the version field, or manually. Of course, it will have checks to know if it makes sense to release.

gizas:
> only for the fact that this repository contains other components.

This is true. I was not thinking for the other components, this makes things more complex

carsonip (approved):
I understand the motivation behind using the changelog generation tool, as this is a mega-repo that makes it hard to identify commits that are only related to loadgen/otelbench. Change lgtm

PR #395, opened 2025-02-27
IntakeV2 receiver - Handle service.language.[name|version] from APM Data model

gregkalapos:
The Elastic APM Data model has:
- `service.language.name`
- `service.language.version`

OTel SemConv defines:
- `telemetry.sdk.language`
- (No corresponding version)

OTel also defines a [well known list of `telemetry.sdk.language` values](https://opentelemetry.io/docs/specs/semconv/attributes-registry/telemetry/) - which is very close to the `service.language.name`, but not 100% the same.
So far what I noticed:
- `C#` in Elastic vs. `dotnet` in OTel
- OTel is always lowercase, ElasticAPM sometimes starts with a capital latter.

This PR cleans this up:
- We keep the two attributes from the Elastic APM Data model and send those unchanged 
- Added logic to map to `telemetry.sdk.language` to satisfy the [well known list from OTel](https://opentelemetry.io/docs/specs/semconv/attributes-registry/telemetry/)


dmathieu (approved):

trentm:
Or perhaps `service.language.{name,version}` map to OTel's `process.runtime.{name,version}` ?
https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/process.md#process-runtimes

Issue #332, opened 2025-01-22
[otelbench] Connect to observability clusters for monitoring metrics


carsonip:
Feature request: make otelbench reach out to an observability cluster to query the target collector (the collector that receives events from otelbench) internal telemetry or apm-server monitoring metrics, and add that to benchmark results.

PR #400, opened 2025-03-01
[otelbench] Update otelbench to report remote internal telemetry metrics

1pkg:
Resolves https://github.com/elastic/opentelemetry-collector-components/issues/332 by updating loadgen otelbench cmd with an ability to optionally pull the remote aggregated otelcollector [basic level metrics](https://opentelemetry.io/docs/collector/internal-telemetry/#basic-level-metrics) from Elasticsearch and print them directly to bench output.

image

These changes were tested manually against a remote APM project and a remote Elasticsearch cluster.


loadgen/cmd/otelbench/remote_stats.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	return elasticsearchStatsFetcher{elasticsearchTelemetryConfig: cfg, client: client}, nil
+}
+
+func (esf elasticsearchStatsFetcher) FetchStats(ctx context.Context, from, to time.Time) (map[string]float64, error) {
+	var filters []types.Query
+	if esf.FilterCluster != "" {
+		filters = append(filters, types.Query{
+			Term: map[string]types.TermQuery{
+				"labels.cluster": {
```
carsonip, line 184:
```suggestion
				"labels.orchestrator_cluster_name": {
```

Q: shouldn't it be `labels.orchestrator_cluster_name`?

loadgen/cmd/otelbench/remote_stats.go
```diff
@@ -0,0 +1,236 @@
...
+					Value: esf.FilterCluster,
+				},
+			},
+		})
+	}
+	if esf.FilterProject != "" {
+		filters = append(filters, types.Query{
+			Term: map[string]types.TermQuery{
+				"labels.project": {
```
carsonip, line 193:
```suggestion
				"labels.project_id": {
```

Q: shouldn't it be `labels.project_id`?

loadgen/cmd/otelbench/README.md
```diff
@@ -34,6 +34,22 @@ Usage of ./otelbench:
...
+        optional remote Elasticsearch telemetry API key
+  -telemetry-elasticsearch-index string
+        optional remote Elasticsearch telemetry metrics index
+  -telemetry-elasticsearch-password string
+        optional remote Elasticsearch telemetry password
+  -telemetry-elasticsearch-url list
+        optional comma-separated list of remote Elasticsearch telemetry hosts
+  -telemetry-elasticsearch-username string
+        optional remote Elasticsearch telemetry username
+  -telemetry-filter-cluster string
```
carsonip, line 47:
```suggestion
  -telemetry-filter-cluster-name string
```

nit: align with dashboard naming

loadgen/cmd/otelbench/README.md
```diff
@@ -34,6 +34,22 @@ Usage of ./otelbench:
...
+        optional remote Elasticsearch telemetry cluster metrics filter
+  -telemetry-filter-project string
```
carsonip, line 49:
```suggestion
  -telemetry-filter-project-id string
```
nit: align with dashboard naming

loadgen/cmd/otelbench/README.md
```diff
@@ -129,3 +145,9 @@ It is possible to run with a customized config to avoid passing in command line
 ```shell
 ./otelbench -config=./my-custom-config.yaml
 ```
+
+Optional remote stats will be reported when additional telemetry flags are provided
+
+```shell
+./otelbench -config=./config.yaml -endpoint-otlp=localhost:4317 -endpoint-otlphttp=https://localhost:4318/prefix -api-key some_api_key -telemetry-elasticsearch-url=localhost:9200 -telemetry-elasticsearch-api-key telemetry_api_key -telemetry-elasticsearch-index "metrics*" -telemetry-filter-cluster cluster_name
```
carsonip, line 152:
```suggestion
./otelbench -config=./config.yaml -endpoint-otlp=localhost:4317 -endpoint-otlphttp=https://localhost:4318/prefix -api-key some_api_key -telemetry-elasticsearch-url=localhost:9200 -telemetry-elasticsearch-api-key telemetry_api_key -telemetry-elasticsearch-index "metrics*" -telemetry-filter-cluster-name cluster_name
```

loadgen/cmd/otelbench/main.go
```diff
@@ -123,11 +137,16 @@ func main() {
 		}
 	}
 
+	fetcher, err := newElasticsearchStatsFetcher(elasticsearchTelemetryConfig(Config.Telemetry))
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
```
carsonip, line 142: non-blocker: I'm thinking if we should fail fast here instead of going ahead with the benchmarking. I see that it isn't possible with the current structure of the code, as an error is returned if the ES telemetry config isn't valid (no connection attempt is made before benchmarks). What I meant by fail fast is to tell between cases
1. telemetry config is not present (user doesn't want telemetry from ES)
2. telemetry config is invalid (e.g. missing credentials) -> fail fast
3. connection failure (e.g. connection error / auth error) -> fail fast
4. telemetry config is present and works ok

marking this as a non-blocker as it doesn't provide additional value if the user is configuring it properly

loadgen/cmd/otelbench/main.go
```diff
@@ -123,11 +137,16 @@ func main() {
...
+	}
+
 	for _, concurrency := range Config.ConcurrencyList {
 		for _, signal := range getSignals() {
 			for _, exporter := range getExporters() {
 				benchName := fullBenchmarkName(signal, exporter, concurrency)
-				result := runBench(signal, exporter, concurrency)
+				result := runBench(fetcher, signal, exporter, concurrency)
```
carsonip, line 149: now that there are these time-based telemetry, does it make sense to add a bit of wait before/after each bench? We can't completely eliminate the effect of one benchmark to another (think about 1m,10m,60m aggregations), but we can try to minimize it.

loadgen/cmd/otelbench/main.go
carsonip: also thank you for using stderr! Great attention to detail

loadgen/cmd/otelbench/README.md
```diff
@@ -129,3 +145,9 @@ It is possible to run with a customized config to avoid passing in command line
 ```shell
 ./otelbench -config=./my-custom-config.yaml
 ```
+
+Optional remote stats will be reported when additional telemetry flags are provided
```
carsonip, line 149: Expanding a bit on "remote stats" would be useful. Maybe give an example about how collector memory can be retrieved from ES?

loadgen/cmd/otelbench/main.go
```diff
@@ -56,8 +57,21 @@ func fullBenchmarkName(signal, exporter string, concurrency int) string {
...
-func runBench(signal, exporter string, concurrency int) testing.BenchmarkResult {
+func runBench(fetcher remoteStatsFetcher, signal, exporter string, concurrency int) testing.BenchmarkResult {
 	return testing.Benchmark(func(b *testing.B) {
+		reportRemoteStats := func(from, to time.Time) {
+			if fetcher == nil {
+				return
+			}
+			stats, err := fetcher.FetchStats(context.Background(), from, to)
+			if err != nil {
+				return
```
carsonip, line 68: possibly log an err to stderr?

loadgen/cmd/otelbench/remote_stats.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	if cfg.ElasticsearchAPIKey == "" && (cfg.ElasticsearchUserName == "" || cfg.ElasticsearchPassword == "") {
+		return errors.New("remote stats will be ignored because of empty telemetry Elasticsearch API key and username or password")
+	}
+	if cfg.ElasticsearchIndex == "" {
+		return errors.New("remote stats will be ignored because of empty telemetry Elasticsearch search index")
+	}
+	// Use the default list of metrics if empty list was provided.
+	if len(cfg.Metrics) == 0 {
+		cfg.Metrics = []string{
```
carsonip, line 143: what about flag.Set?

loadgen/cmd/otelbench/flags.go
```diff
@@ -131,6 +144,41 @@ func Init() error {
...
+				urls = append(urls, val)
+			}
+			Config.Telemetry.ElasticsearchURL = urls
+			return nil
+		},
+	)
+	flag.StringVar(&Config.Telemetry.ElasticsearchUserName, "telemetry-elasticsearch-username", "", "optional remote Elasticsearch telemetry username")
+	flag.StringVar(&Config.Telemetry.ElasticsearchPassword, "telemetry-elasticsearch-password", "", "optional remote Elasticsearch telemetry password")
+	flag.StringVar(&Config.Telemetry.ElasticsearchAPIKey, "telemetry-elasticsearch-api-key", "", "optional remote Elasticsearch telemetry API key")
+	flag.StringVar(&Config.Telemetry.ElasticsearchIndex, "telemetry-elasticsearch-index", "", "optional remote Elasticsearch telemetry metrics index")
```
carsonip, line 164:
```suggestion
	flag.StringVar(&Config.Telemetry.ElasticsearchIndex, "telemetry-elasticsearch-index", "", "optional remote Elasticsearch telemetry metrics index pattern")
```

carsonip (commented):
looks great! a few nits and questions

loadgen/cmd/otelbench/remote_stats.go
1pkg: updated

loadgen/cmd/otelbench/remote_stats.go
1pkg: updated

loadgen/cmd/otelbench/README.md
1pkg: updated

loadgen/cmd/otelbench/README.md
1pkg: updated

loadgen/cmd/otelbench/README.md
1pkg: updated

loadgen/cmd/otelbench/main.go
1pkg: Yes, a good idea, I updated ES fetcher to run a quick ping to fail fast if configs are provided.

loadgen/cmd/otelbench/README.md
1pkg: updated

loadgen/cmd/otelbench/main.go
1pkg: updated

loadgen/cmd/otelbench/flags.go
1pkg: updated

loadgen/cmd/otelbench/remote_stats.go
1pkg: I was in between of using validation to set the metrics defaults vs using flag. Both options have their pros and cons. But I agree using `flag.Set` is more natural way. Updated.

loadgen/cmd/otelbench/main.go
1pkg: I added a flat 1m wait after each run, I'm not sure if this will be "good enough" wdyt?

carsonip (commented):
i missed a point about counter metrics correctness in the 1st review. Not sure how much work is needed to fix that. If it is too much, we might want to fix it in a follow up.

loadgen/cmd/otelbench/main.go
1pkg: I made some updates in the last commit https://github.com/elastic/opentelemetry-collector-components/pull/400/commits/64373254d1fc08baf156f5d3f20125ccf8a4e4ae to manage the timestamp offset better. Essentially, moving from/to timestamp logic out of `runBench` and to the main loop that schedules the benchmark runs.

Now the following logic applies: we wait for collector to finish, then we wait for a small duration to pad a bit for late metric arrivals and scraping interval, then we collect remote metrics in the time interval, finally we advance lower `from` interval with a bit of padding again to prevent mixing results. It's not bullet proof against large delays. But IMO it should be good enough to try it.

As for `wait for ES for the existence of a metric timestamp > bench end time` it might be event better but it adds complexity and I don't think this will be way more reliably neither, if there is a large delay in metric reporting we cannot reliably tell if remote metrics will include results from a previous run or not.

loadgen/cmd/otelbench/README.md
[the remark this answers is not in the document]
1pkg: resolved

loadgen/cmd/otelbench/remote_stats.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/elastic/go-elasticsearch/v8/typedapi/core/search"
+	"github.com/elastic/go-elasticsearch/v8/typedapi/some"
+	"github.com/elastic/go-elasticsearch/v8/typedapi/types"
+)
+
+// knownMetricsAggregations corresponds to know basic-level metrics from
+// https://opentelemetry.io/docs/collector/internal-telemetry/#basic-level-metrics.
+var knownMetricsAggregations map[string]types.Aggregations = map[string]types.Aggregations{
+	"otelcol_exporter_enqueue_failed_log_records": {
+		Sum: &types.SumAggregation{Field: some.String("otelcol_exporter_enqueue_failed_log_records")},
```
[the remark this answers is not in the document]
1pkg: Oh that's a great point on cumulative temporality for counter metrics. I updated the fetcher to use max-min aggregation now to get the aggregated sum for counters.

For the `What makes it harder is that we need to sum across collectors but do Max-Min per collector` as of now looking at the metrics we don't really include collector identity into metrics attributes as far as I can tell. So currently the metrics are already reported as a blended sum of all collectors. Nevertheless, in the future when we need to aggregate per collector we should be able to update the current query to use nested aggregations to additionally group max-min per collector and then manually sum them.

axw (commented):
Generally looks good, just minor comments

loadgen/cmd/otelbench/main.go
```diff
@@ -71,8 +73,7 @@ func runBench(signal, exporter string, concurrency int) testing.BenchmarkResult
 		if signal != "traces" {
 			close(tracesDone)
 		}
-
-		stop := make(chan struct{}) // close channel to stop the loadgen collector
+		stop := make(chan struct{}, 2) // close channel to stop the benchmark run
```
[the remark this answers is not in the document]
1pkg: Oh here I just changed a logic a bit to separate: 
- stopping collector first
- and then after collector is stopped fetching remote metrics

So I'm reusing channel to do this in sequence. Let me see if I can simplify this logic a bit so we it's more linear.

loadgen/cmd/otelbench/flags.go
```diff
@@ -131,6 +147,45 @@ func Init() error {
...
+				if val == "" {
+					continue
+				}
+				m = append(m, val)
+			}
+			Config.Telemetry.Metrics = m
+			return nil
+		},
+	)
+	if err := flag.Set("telemetry-metrics", "otelcol_process_cpu_seconds,otelcol_process_memory_rss,otelcol_process_runtime_total_alloc_bytes,otelcol_process_runtime_total_sys_memory_bytes,otelcol_process_uptime"); err != nil {
```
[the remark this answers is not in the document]
1pkg: updated

loadgen/cmd/otelbench/flags.go
```diff
@@ -131,6 +147,45 @@ func Init() error {
...
+		return errors.New(`error setting default value for flag "telemetry-metrics"`)
```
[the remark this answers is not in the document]
1pkg: replaced with `flag.DefaultValue`

loadgen/cmd/otelbench/README.md
```diff
@@ -34,6 +34,24 @@ Usage of ./otelbench:
...
+        optional remote Elasticsearch telemetry request timeout (default 1m0s)
+  -telemetry-elasticsearch-url list
+        optional comma-separated list of remote Elasticsearch telemetry hosts
+  -telemetry-elasticsearch-username string
+        optional remote Elasticsearch telemetry username
+  -telemetry-filter-cluster-name string
+        optional remote Elasticsearch telemetry cluster name metrics filter
+  -telemetry-filter-project-id string
+        optional remote Elasticsearch telemetry project id metrics filter
+  -telemetry-metrics list
```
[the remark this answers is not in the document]
1pkg: updated

loadgen/cmd/otelbench/README.md
```diff
@@ -34,6 +34,24 @@ Usage of ./otelbench:
         benchmark metrics (default true)
   -secret-token string
         secret token for target server
+  -telemetry-elasticsearch-api-key string
+        optional remote Elasticsearch telemetry API key
+  -telemetry-elasticsearch-index string
```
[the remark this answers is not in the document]
1pkg: updated

loadgen/cmd/otelbench/main.go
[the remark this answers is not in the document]
1pkg: We need this separate goroutine to wait for signal of the completion and stop the collector, there is no easy way to make the logic more linear there. Although, I made a small refactoring to replace this buffered multipurpose `stop` channel with 2 separate channels to `stop` collector and `done` for the entire `runBench`.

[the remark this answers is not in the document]
axw: Thanks, looks clearer now!

loadgen/cmd/otelbench/remote_stats.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (minMax) Value(metric string, aggs map[string]types.Aggregate) *float64 {
+	min, minOk := aggs[metric+"_min"].(*types.MinAggregate)
+	max, maxOk := aggs[metric+"_max"].(*types.MaxAggregate)
+	if minOk && maxOk && min.Value != nil && max.Value != nil {
+		v := float64(*max.Value) - float64(*min.Value)
+		return &v
+	}
+	return nil
+}
```
axw, line 68: The `max-min` logic could do with some comments to explain. IIUC we're effectively doing cumulative-to-delta conversion, and it won't work across multiple collectors. I'd suggest adding a TODO as well to call that out, and mention we'll need to do a nested agg/group-by to calculate per collector instance.

loadgen/cmd/otelbench/main.go
```diff
@@ -147,15 +138,32 @@ func main() {
...
+					stats, err := fetcher.FetchStats(ctx, from, to)
+					if err != nil {
+						fmt.Fprintf(os.Stderr, "error while fetching remote stats %s", err)
+						return
+					}
+					for unit, n := range stats {
+						b.ReportMetric(n, unit)
+					}
+					// advance the timestamp a bit to minimize the metrics influence of the previous run over the next run
+					from = to.Add(statsFetcherDurationPadding)
```
carsonip, line 163: does it also mean we'll miss the first `padding` of metrics from next bench run? I don't have a strong opinion here, just commenting as I'm not sure what value we get from adding 2nd padding here as we've already done `time.Sleep`.

carsonip (approved):
lgtm, very impressive!
nit: ++ on a comment on why we are doing max-min

loadgen/cmd/otelbench/remote_stats.go
1pkg: Added a comment to the ES aggregation used for OTel metrics, and TODO for Max-Min aggregation when collector instance dimension will be added.

loadgen/cmd/otelbench/remote_stats.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+}
+
+func (minMax) Value(metric string, aggs map[string]types.Aggregate) *float64 {
+	min, minOk := aggs[metric+"_min"].(*types.MinAggregate)
+	max, maxOk := aggs[metric+"_max"].(*types.MaxAggregate)
+	if minOk && maxOk && min.Value != nil && max.Value != nil {
+		v := float64(*max.Value) - float64(*min.Value)
+		return &v
+	}
```
axw, line 77: Do we care about the case where there's just a single datapoint? In which case min and max will be the same, and the delta will apparently be zero, but that would be wrong.

axw (approved):
LGTM! One question which is nonblocking because I don't think we care, but wanted to doublecheck

loadgen/cmd/otelbench/main.go
```diff
@@ -124,11 +130,37 @@ func main() {
...
 
+	fetcher, ignore, err := newElasticsearchStatsFetcher(elasticsearchTelemetryConfig(Config.Telemetry))
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		if !ignore {
+			os.Exit(2)
+		}
+	}
+
+	from := time.Now().UTC()
```
carsonip, line 141:
```suggestion
```

loadgen/cmd/otelbench/main.go
```diff
@@ -124,11 +130,37 @@ func main() {
...
+					stats, err := fetcher.FetchStats(ctx, from, to)
+					if err != nil {
+						fmt.Fprintf(os.Stderr, "error while fetching remote stats %s", err)
+						return
+					}
+					for unit, n := range stats {
+						b.ReportMetric(n, unit)
+					}
+					// advance the timestamp again for the next run
+					from = time.Now().UTC()
```
carsonip, line 162:
```suggestion
```

loadgen/cmd/otelbench/main.go
```diff
@@ -124,11 +130,37 @@ func main() {
...
+		}
+	}
+
+	from := time.Now().UTC()
 	for _, concurrency := range Config.ConcurrencyList {
 		for _, signal := range getSignals() {
 			for _, exporter := range getExporters() {
 				benchName := fullBenchmarkName(signal, exporter, concurrency)
-				result := runBench(signal, exporter, concurrency)
+				result := runBench(ctx, signal, exporter, concurrency, func(b *testing.B) {
```
carsonip, line 146:
```suggestion
                                from := time.Now().UTC()
				result := runBench(ctx, signal, exporter, concurrency, func(b *testing.B) {
```

carsonip (commented):
nit: seems we can simplify `from`

loadgen/cmd/otelbench/remote_stats.go
1pkg: updated to handle this case properly

1pkg:
I discovered a cool artifact when testing this against a local collector instance, basically when running locally the benchmark pass very fast. And if the benchmarking time is less than duration of metrics reporting the remote metrics will be returned empty thus will be skipped from the benchmarks too.

image

I don't think it's a bit deal because on a remote MOTEL the speed is different and if we ever need to address this, we could just extend the from/to timestamp padding to be in sync with metric reporting interval from collector.

PR #407, opened 2025-03-04
[loadgen/otelbench] Add release otelbench workflow

constanca-m:
Add a release workflow that can be triggered manually, or by updating the version inside the MAKEFILE in otelbench directory.

The workflow will update the changelog and release the docker image in elastic container registry.

### DIFF
--- a/.github/workflows/release_otelbench.yaml
+++ b/.github/workflows/release_otelbench.yaml
@@ -1,0 +1,106 @@
+run-name: release-otelbench
+name: Release otelbench docker image on elastic container library
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version of the otelbench image to release, should follow format "vN.N.N".
+        required: true
+  push:
+    branches:
+      - main
+    paths:
+      - ./loadgen/cmd/otelbench/Makefile
+
+jobs:
+  check-release:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get-new-version.outputs.new_version }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: '0'
+
+      - name: Checks if release is required.
+        id: get-new-version
+        run: ./scripts/check_release.sh "${{ github.event_name }}" "${{ github.event.inputs.version }}"
+
+  release-otelbench:
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    needs:
+      - check-release
+    if: needs.check-release.result == 'success'
+    permissions:
+      contents: write
+    env:
+      ELASTIC_REGISTRY: "docker.elastic.co/observability-ci"
+      IMAGE_NAME: "otelbench"
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: '0'
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: 'false'
+
+      - name: Cache go tools
+        id: go-tools-cache
+        uses: actions/cache@v4
+        with:
+          path: ./.tools
+          key: go-tools-${{ runner.os }}-${{ hashFiles('internal/tools/go.sum') }}
+
+      # If there is no cache with the go tools,
+      # install them. We need the changelog
+      # tool for this workflow to work.
+      - name: Install tools
+        if: steps.go-tools-cache.outputs.cache-hit != 'true'
+        run: |
+          make install-tools
+
+      - name: Update otelbench changelog and delete changelog fragments
+        working-directory: ./loadgen/cmd/otelbench
+        run: ../../../.tools/chloggen update --version ${{ needs.check-release.outputs.version }}
+
+      - name: Update Makefile version if dispatched
+        if: github.event_name == 'workflow_dispatch'
+        working-directory: ./loadgen/cmd/otelbench
+        run: sed -i "s/^VERSION := .*/VERSION := ${{ github.event.inputs.version }}/" Makefile
+
+      - name: Commit changes
+        run: |
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
+          
+          git checkout main
+          git pull --rebase origin main
+          git add ./loadgen/cmd/otelbench
+          if ! git diff --quiet ./loadgen/cmd/otelbench; then
+            git commit -m "Update otelbench changelog and delete changelog fragments."
+            git push origin main
+          fi
+
+      - name: Login to Elastic container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.ELASTIC_REGISTRY }}
+          username: ${{ secrets.DOCKER_AUTH_USERNAME }}
+          password: ${{ secrets.DOCKER_AUTH_PASSWORD }}
+
+      - uses: ko-build/setup-ko@v0.8
+        env:
+          KO_DOCKER_REPO: ${{ env.ELASTIC_REGISTRY }}
+
+      - name: Build container image and push to Elastic registry
+        working-directory: ./loadgen/cmd/otelbench
+        run: |
+          export KO_DOCKER_REPO=${{ env.ELASTIC_REGISTRY }}
+          local_platform="linux/$(go env GOARCH)"
+          KO_IMAGE=$(ko build --platform=$local_platform)
+          docker tag "$KO_IMAGE" "${{ env.ELASTIC_REGISTRY }}/${{ env.IMAGE_NAME }}:${{ needs.check-release.outputs.version }}"
--- a/.github/workflows/scripts/check_release.sh
+++ b/.github/workflows/scripts/check_release.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+set -eo pipefail
+
+# This script gets the previous version from the
+# previous commit, by running make get-version.
+# Then it compares this previous version with the
+# new given version. If there was an increase
+# in version, then a release is required.
+
+GITHUB_EVENT_NAME=$1
+GITHUB_INPUT_VERSION=$2
+
+WORKING_DIR="../../../loadgen/cmd/otelbench"
+
+# Get new version
+if [ "$GITHUB_EVENT_NAME" = "workflow_dispatch" ]; then
+  NEW_VERSION="$GITHUB_INPUT_VERSION"
+else
+  NEW_VERSION=$(make -C "$WORKING_DIR" get-version 2>/dev/null || echo "v0.0.0")
+fi
+
+# Get previous version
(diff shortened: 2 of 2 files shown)


.github/workflows/release_otelbench.yaml
```diff
@@ -0,0 +1,106 @@
...
+            git commit -m "Update otelbench changelog and delete changelog fragments."
+            git push origin main
+          fi
+
+      - name: Login to Elastic container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.ELASTIC_REGISTRY }}
+          username: ${{ secrets.DOCKER_AUTH_USERNAME }}
+          password: ${{ secrets.DOCKER_AUTH_PASSWORD }}
```
constanca-m, line 94: @v1v I once again need your help.

I want to push an image to `docker.elastic.co/observability-ci` registry. Is this the proper way to login to the registry? I have no access to the secrets variables in this repository, so which values should be here?

v1v: 👋 , I can create those GH secrets for you, the values are:
- ELASTIC_DOCKER_USERNAME
- ELASTIC_DOCKER_PASSWORD
- ELASTIC_DOCKER_REGISTRY

constanca-m: Thank you!

v1v: Secrets should be available in about 2/3 minutes or so

.github/workflows/release_otelbench.yaml
```diff
@@ -0,0 +1,106 @@
...
+    outputs:
+      version: ${{ steps.get-new-version.outputs.new_version }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: '0'
+
+      - name: Checks if release is required.
+        id: get-new-version
+        run: ./scripts/check_release.sh "${{ github.event_name }}" "${{ github.event.inputs.version }}"
```
v1v, line 29: maybe you need `continue-on-error: true`? Otherwise, this will fail and might confuse what's the rationale.

OTOH, I wonder whether this approach can be changed to do something like:
- https://github.com/elastic/apm-aggregation-service/blob/52b0be77db5049433fb9ef783cbc4bf0cff1d75d/.github/workflows/create-release-tag.yml#L26-L45

and then you use https://github.com/elastic/apm-aggregation-service/blob/52b0be77db5049433fb9ef783cbc4bf0cff1d75d/.github/workflows/create-release-tag.yml#L57-L65 in the `release-otelbench:` job.


The idea is to rely on the GH releases to create new releases if no release was done.

.github/workflows/release_otelbench.yaml
```diff
@@ -0,0 +1,106 @@
...
+      # install them. We need the changelog
+      # tool for this workflow to work.
+      - name: Install tools
+        if: steps.go-tools-cache.outputs.cache-hit != 'true'
+        run: |
+          make install-tools
+
+      - name: Update otelbench changelog and delete changelog fragments
+        working-directory: ./loadgen/cmd/otelbench
+        run: ../../../.tools/chloggen update --version ${{ needs.check-release.outputs.version }}
```
v1v, line 69: Do you think `/.tools` could be added to the PATH so the relative path can be simplified?

```yaml
      - name: Add .tools to PATH
        run: echo "$GITHUB_WORKSPACE/.tools" >> $GITHUB_PATH

      - name: Update otelbench changelog and delete changelog fragments
        working-directory: ./loadgen/cmd/otelbench
        run: ./chloggen update --version ${{ needs.check-release.outputs.version }}
```

.github/workflows/release_otelbench.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Update Makefile version if dispatched
+        if: github.event_name == 'workflow_dispatch'
+        working-directory: ./loadgen/cmd/otelbench
+        run: sed -i "s/^VERSION := .*/VERSION := ${{ github.event.inputs.version }}/" Makefile
+
+      - name: Commit changes
+        run: |
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
```
v1v, line 79: You won't be able to push changes, but you can reuse one of our reusable gh actions to help with the config settings.

See https://github.com/elastic/oblt-actions/tree/main/git/setup

Regarding pushing changes, a few questions:

1) If this project uses GH actions for the CI, then it should not use the GitHub actions GH Secret but a GH app.
2) Otherwise, GH workflows won't run for events created by the GH actions bot - to avoid recurrence.
3) I cannot see the section `permissions` with the required scopes `contents: write`

.github/workflows/release_otelbench.yaml
v1v: Sure, but it will be shown as a failure in https://github.com/elastic/opentelemetry-collector-components/actions, and the failure can be for anything.

.github/workflows/release_otelbench.yaml
constanca-m: > I cannot see the section permissions with the required scopes contents: write

It's in the job itself:

https://github.com/elastic/opentelemetry-collector-components/blob/620724e0c765078fa59b8bddef4550826f2bdee7/.github/workflows/release_otelbench.yaml#L37-L38

.github/workflows/release_otelbench.yaml
constanca-m: Thanks, I didn't know this existed. Fixed it.

.github/workflows/release_otelbench.yaml
constanca-m: > If this project uses GH actions for the CI, then it should not use the GitHub actions GH Secret but a GH app.
Otherwise, GH workflows won't run for events created by the GH actions bot - to avoid recurrence.

I am a bit confused over this. This part of the workflow needs to delete files in the repository, and update a file. I understand now that it needs to create a PR, right? 

After the changelog is updated, then it creates a docker image and it pushes the image to elastic private registry.

v1v: Aha, cool, thanks for this. Do you mind adding the section `contents: read` at the top?

```yaml

permissions:
  contents: read

jobs:
```

We use the least-permissive access by default.

v1v:
@constanca-m , if it helps we can sync-up and clarify some of my questions, in case I made things a little bit complicated to follow-up 🙇

.github/workflows/release_otelbench.yaml
v1v: Fair enough

.github/workflows/release_otelbench.yaml
constanca-m: I have changed it to this:

https://github.com/elastic/opentelemetry-collector-components/blob/b579ae97bb8087e8aa0c6763e4f5e6c8745d7044/.github/workflows/release_otelbench.yaml#L82-L115

I don't know what should be inside `github-token: ${{ secrets.GITHUB_TOKEN }} `.

I am open to follow a different approach too, if you think it is best.

constanca-m: I have sent you an invite to sync @v1v . Feel free to adjust the time if it does not fit you.

constanca-m: I have updated the PR now @v1v so that:

- It fetches the github token
- Uses a script to create the PR, similarly to your reference

constanca-m: @v1v Following our discussion, I have pushed a new commit:

- Uses the commit instead of the tag
- Splits the jobs between two workflows:
  - One to open a PR to update the version/changelog
  - Another one that gets triggered once the CHANGELOG.md is updated, and publishes a new docker image

Does it look good now?

.github/workflows/scripts/create_pr.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -eo pipefail
+
+VERSION=$1
+GITHUB_EVENT_NAME=$2
+BRANCH_NAME="update-changelog-$VERSION"
```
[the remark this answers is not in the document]
constanca-m: Done.

v1v (approved):
LGTM, just one left over.

Thanks for working on this!! 👏

.github/workflows/scripts/create_pr.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+set -eo pipefail
+
+VERSION=$1
+GITHUB_EVENT_NAME=$2
+BRANCH_NAME="update-changelog-$VERSION-$(date +%F_%H-%M-%S)"
+
+(
+  cd ./loadgen/cmd/otelbench
```
constanca-m, line 9: Actually I think I made a mistake before. So this script is called from the GH workflow, so we are at the root. Then we use `cd` here, so when the scope ends, we are back to the root directory. So I removed pushd and popd. Is this correct @v1v ?

v1v (approved):

loadgen/cmd/otelbench/README.md
```diff
@@ -140,9 +140,15 @@ If you want to contribute to any go files, you need to create a changelog entry:
 
 ## Create new release
 
-TODO
-<!-----
-Have a workflow that gets triggered by a version update (maybe the version from the MAKEFILE, maybe another option), and runs `$(CHLOGGEN) update --version $(VERSION)`, updating the CHANGELOG.md and deleting all changelog fragments + it pushes a new image to elastic container
-registry.
----->
+There are two ways you can trigger a new release:
```
v1v, line 143: 🥇  that's brilliant! Thanks!

loadgen/cmd/otelbench/README.md
```diff
@@ -140,9 +140,15 @@ If you want to contribute to any go files, you need to create a changelog entry:
...
+- Manually, by triggering the GH actions workflow `bump-otelbench`.
+- Automatically, by updating the VERSION field in the Makefile.
```
carsonip, line 145: Q:
- does updating the VERSION field also trigger `bump-otelbench`?
- automatic as in the workflow is run automatically? The trigger is still manual, by doing a code change on the VERSION field. Is the benefit that we can perform a release in the same code change PR, and avoiding to trigger bump-otelbench manually?

carsonip (commented):
Only reviewing the README from an end user PoV. Left a question

loadgen/cmd/otelbench/README.md
constanca-m: > does updating the VERSION field also trigger bump-otelbench?

Yes. If the version increasing, then CHANGELOG is updated, a PR is opened, and then the docker image is released (once the PR is merged).

> automatic as in the workflow is run automatically? 

Yes.

> The trigger is still manual, by doing a code change on the VERSION field. Is the benefit that we can perform a release in the same code change PR, and avoiding to trigger bump-otelbench manually?

Yes.

PR #440, opened 2025-03-06
[workflow/otelbench] Use git diff to detect changes before creating PR

constanca-m:
image

Even though the workflow runs "successfully", it does not open any PR like expected. This happens because changes were not detected. I am changing the command to use `diff-index`. Example:

```
c@c:~/Documents/Git-Forks/Mine/opentelemetry-collector-components$ git diff-index --cached HEAD
:000000 100644 0000000000000000000000000000000000000000 e69de29bb2d1d6434b8b29ae775ad8c2e48c5391 A      loadgen/cmd/otelbench/test
```


gizas:
This is what we do currently in k8s flow: https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/Makefile#L79C2-L79C27

gizas (approved):

PR #453, opened 2025-03-12
Add helpers for integrations components

jsoriano:
Add an internal package with common helpers and infrastructure for components related to collector integrations.

It includes mainly the logic to find and resolve integrations.

This is part of https://github.com/elastic/opentelemetry-collector-components/pull/96, more details there.


internal/integrations/integrations.go
```diff
@@ -0,0 +1,123 @@
...
+	if !anyExtension {
+		return nil, errors.New("no integration finder extension found")
+	}
+
+	return nil, ErrNotFound
+}
+
+// Config contains an structured integration.
+type Config struct {
+	Receivers  map[component.ID]map[string]any `mapstructure:"receivers"`
```
florianl, line 79: If possible, map in map should be avoided. Can there be a situation, where a particular `component.ID` is in `Receivers` but not in `Processors`?

florianl, line 79: Can you think of alternatives to `map[string]any`? What is the reason to use such a construct?

internal/integrations/rawtemplate.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// Resolve resolves the template using a confmap resolver.
+func (t *RawTemplate) Resolve(ctx context.Context, params map[string]any, pipelines []pipeline.ID) (*Config, error) {
+	selectedPipelines := t.source.Pipelines
+	if len(pipelines) > 0 {
+		var err error
+		selectedPipelines, err = selectComponents(t.source.Pipelines, pipelines)
+		if err != nil {
+			return nil, fmt.Errorf("selecting pipelines: %w", err)
```
florianl, line 62: No error of `selectComponents` is part of an API so that one could use it in any way.
```suggestion
			return nil, fmt.Errorf("selecting pipelines: %v", err)
```

internal/integrations/rawtemplate.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+	}
+	selectedReceivers, err := selectComponents(t.source.Receivers, listReceivers(selectedPipelines))
+	if err != nil {
+		return nil, fmt.Errorf("selecting receivers: %w", err)
```
florianl, line 67: No error of selectComponents is part of an API so that one could use it in any way.


```suggestion
		return nil, fmt.Errorf("selecting receivers: %v", err)
```

internal/integrations/rawtemplate.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	selectedProcessors, err := selectComponents(t.source.Processors, listProcessors(selectedPipelines))
+	if err != nil {
+		return nil, fmt.Errorf("selecting processors: %w", err)
```
florianl, line 71: No error of selectComponents is part of an API so that one could use it in any way.


```suggestion
		return nil, fmt.Errorf("selecting processors: %v", err)
```

internal/integrations/rawtemplate.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	return confmap.NewResolver(settings)
+}
+
+type selectableID interface {
+	comparable
+	fmt.Stringer
+}
+
+func selectComponents[C any, ID selectableID](from map[ID]C, selection []ID) (map[ID]C, error) {
```
florianl, line 111: all callers of `selectComponents()` just check if the returned error is nil. but if the returned `map[ID]C` is nil is never checked and might cause issues.

internal/integrations/rawtemplate.go
```diff
@@ -0,0 +1,240 @@
...
+		component, found := from[id]
+		if !found {
+			return nil, fmt.Errorf("component %q not found", id)
+		}
+		selected[id] = component
+	}
+	return selected, nil
+}
+
+func listReceivers(pipelines map[pipeline.ID]PipelineConfig) []component.ID {
```
florianl, line 126: What is the difference between `listReceivers()` and `listProcessors()`? Can these two functions be merged?

internal/integrations/rawtemplate.go
```diff
@@ -0,0 +1,240 @@
...
+		return confmap.NewRetrievedFromYAML(d)
+	case "pipelines":
+		d, err := yaml.Marshal(p.main.Pipelines)
+		if err != nil {
+			return nil, err
+		}
+		return confmap.NewRetrievedFromYAML(d)
+	}
+
+	return nil, fmt.Errorf("%s not found", uri)
```
florianl, line 217:
```suggestion
        default:
            return nil, fmt.Errorf("%s from %s not found", key, uri)
	}

```

internal/integrations/integrations.go
```diff
@@ -0,0 +1,123 @@
...
+	// pipelines.
+	Resolve(ctx context.Context, params map[string]any, pipelines []pipeline.ID) (*Config, error)
+}
+
+// Finder is the interface for extensions that can be used to find integrations by their names.
+type Finder interface {
+	FindIntegration(ctx context.Context, name string) (Integration, error)
+}
+
+// Find looks for integrations in extensions of the host that implement the IntegrationFinder interface.
```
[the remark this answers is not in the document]
jsoriano: Some tests added.

If there are more than one extension that implement the Finder interface all of them are used, and the first one found is returned.

But as we are doing it now there is no guarantee on the order used to evaluate the extensions as we just iterate over the map of extensions.
I think we should guarantee some kind of order, so users can provide multiple finders with different priorities, though not sure how to order them. At this point we don't have the collector config, so we cannot rely on the order of definition. Should we use alphabetical order by `eid.Name()`? Or add an explicit "priority" setting?

internal/integrations/integrations.go
```diff
@@ -0,0 +1,123 @@
...
+			logger.Error("integration finder failed",
+				zap.String("component", eid.String()),
+				zap.Error(err))
+			return nil, err
+		}
+
+		return integration, nil
+	}
+	if !anyExtension {
+		return nil, errors.New("no integration finder extension found")
```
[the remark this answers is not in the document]
jsoriano: > What difference does it make for the user of this API, when they receive errors.New("no integration finder extension found") vs ErrNotFound?

No actual difference. I will return an error wrapping `ErrNotFound` also here. We need to return an error here so the caller knows if it 

> In both cases no Integration is returned. Would it be fine just to log cases, where the is no finder extension?

We need to return an error in both cases, so the caller knows if it can use the integration.

internal/integrations/integrations.go
jsoriano: Right, I think this was a leftover of some previous version. Now we only care about the component IDs at this point. I think the `map[string]any` is not actually needed, removing it by now.

internal/integrations/rawtemplate.go
jsoriano: The value returned here is only used in `buildMainYML`, that checks if the map is empty using `len`.
Do you think I should return an initialized empty map instead of `nil`?

internal/integrations/rawtemplate.go
jsoriano: They list the receivers and the processors in the map of pipeline configs. Both sets of ids come from different fields.

Refactored to reuse the iteration, sort and compaction.

internal/integrations/rawtemplate.go
jsoriano: Do you think it can be a problem to wrap the error even if not useful now? I feel safer to wrap it just in case.

internal/integrations/integrations.go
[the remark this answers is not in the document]
constanca-m: > If there are more than one extension that implement the Finder interface all of them are used, and the first one found is returned.

What do you mean by all of them are used?

I was thinking that maybe it is useful if Integration has function `Name()` so it is known which integration is being used.

internal/integrations/rawtemplate.go
jsoriano: Changed to don't return `nil`.

internal/integrations/integrations.go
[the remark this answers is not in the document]
constanca-m: For the unit tests, you have:

```
	integration1 := mustNewIntegration(rawTemplate1)
	integration2 := mustNewIntegration(rawTemplate2)
```

And when I was testing it, I wanted to know which was being returned, that's all.

But if you think there is no use at all for this, then never mind.

internal/integrations/rawtemplate_test.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	for _, c := range cases {
+		t.Run(c.title, func(t *testing.T) {
+			var template Integration
+			var err error
+			template, err = NewRawTemplate(readTemplateFile(c.file))
+			if c.fileErr != "" {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), c.fileErr)
```
constanca-m, line 230: You can use [assert.ErrorContains](https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorContains) instead if you prefer:

```suggestion
				assert.ErrorContains(t, err, c.fileErr)
```

internal/integrations/rawtemplate_test.go
```diff
@@ -0,0 +1,262 @@
...
+				return
+			}
+			require.NoError(t, err)
+
+			config, err := template.Resolve(context.Background(), c.params, c.pipelines)
+			if c.expectedErr != "" {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), c.expectedErr)
```
constanca-m, line 238: The same here as https://github.com/elastic/opentelemetry-collector-components/pull/453/files#r1998990835.

internal/integrations/confmap.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (p *variablesProvider) Retrieve(ctx context.Context, uri string, _ confmap.WatcherFunc) (*confmap.Retrieved, error) {
+	key := strings.TrimPrefix(uri, varProviderScheme+":")
+	value, found := p.variables[key]
+	if !found {
+		return nil, fmt.Errorf("variable %q not found", key)
+	}
+
+	return confmap.NewRetrieved(value)
+}
```
constanca-m, line 54: Any chance you could use unit variables here? I don't want to block this PR on this, but unit tests also help a lot with reading the code :)

internal/integrations/rawtemplate_test.go
jsoriano: Nice one, changed, thanks.

internal/integrations/rawtemplate_test.go
jsoriano: Changed, thanks.

internal/integrations/confmap.go
jsoriano: Unit tests added for this provider, let me know if any other clarification is needed.

constanca-m (approved):
Thanks @jsoriano for addressing all my comments. I don't have the full picture on this helper, but the code that is here seems good.

jsoriano:
Thanks @constanca-m for your comments, once merged I will follow with other components in https://github.com/elastic/opentelemetry-collector-components/pull/96, what hopefully will give a more complete image.

@florianl would you like to take another look?

Before merging it looks like I will also need a review by @tetianakravchenko or @LikeTheSalad, could you please take a look?

constanca-m:
@tetianakravchenko is on leave

LikeTheSalad:
> Before merging it looks like I will also need a review by @tetianakravchenko or @LikeTheSalad, could you please take a look?

I'm happy to help, although I'm not familiar with the code involved in this PR. May I ask which part do you think I should take a look at?

jsoriano:
@LikeTheSalad sorry, I pinged you because you were automatically assigned as code owner, and I don't seem to be allowed to merge :innocent:
image

But I see that @constanca-m also belongs to the otel-devs team, so maybe I just don't have permissions to merge? :thinking:

constanca-m:
> so maybe I just don't have permissions to merge? 🤔

That is the case. If you want to merge it, you can ping me and I merge it for you.

internal/integrations/rawtemplate.go
florianl: If an error is wrapped it becomes part of the API. Unless some expected errors are documented or dedicated errors are exposed and part of the API, there should be no reason to wrap it.

internal/integrations/integrations.go
```diff
@@ -0,0 +1,138 @@
...
+// Host is the interface a host must implement to be used to locate finders along extensions.
+type Host interface {
+	// GetExtensions returns the map of extensions. Only enabled and created extensions will be returned.
+	// GetExtensions can be called by the component anytime after Component.Start() begins and
+	// until Component.Shutdown() ends.
+	GetExtensions() map[component.ID]component.Component
+}
+
+// Find looks for integrations in extensions of the host that implement the IntegrationFinder interface.
+func Find(ctx context.Context, logger *zap.Logger, host Host, name string) (Integration, error) {
```
florianl, line 56: Should the user be able to differentiate between no integration matches the given `name` vs there are no integrations at all? So far all return paths return `ErrNotFound` which let the user not differentiate in an automated way.

internal/integrations/integrations.go
```diff
@@ -0,0 +1,138 @@
...
+
+		for _, processor := range pipeline.Processors {
+			if _, found := c.Processors[processor]; !found {
+				return fmt.Errorf("processor %q not defined", processor.String())
+			}
+
+		}
+	}
+
+	return nil
```
florianl, line 118: In the case of an empty `Pipelines` map, `nil` is returned. Should we instead return something like `missing Pipelines`?

internal/integrations/integrations.go
```diff
@@ -0,0 +1,138 @@
...
+	Receiver *component.ID `mapstructure:"receiver"`
+
+	// Processors is the chain of processors of the pipeline, to be used as part of the receiver in receiver
+	// components, or as a combined processor when the pipeline is used as processor.
+	Processors []component.ID `mapstructure:"processors"`
+}
+
+// Validate validates the PipelineConfig.
+func (p *PipelineConfig) Validate() error {
+	return nil
```
florianl, line 134: Does this intentionally return always `nil`?

internal/integrations/integrations.go
```diff
@@ -0,0 +1,138 @@
...
+			if _, found := c.Processors[processor]; !found {
+				return fmt.Errorf("processor %q not defined", processor.String())
+			}
+
+		}
+	}
+
+	return nil
+}
+
```
florianl, line 120: If `PipelineConfig` implements some interface - which interface does it implement?
Otherwise, can we drop `(p *PipelineConfig) Validate() error{}`?

internal/integrations/rawtemplate.go
```diff
@@ -0,0 +1,241 @@
...
+	"gopkg.in/yaml.v3"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/pipeline"
+)
+
+// RawTemplate implements the Template interface for raw YAML content.
+// Unused components are removed after resolving it, so the variables they contain are not required.
+type RawTemplate struct {
```
florianl, line 36: Similar to above: Does RawTemplate implement an interface? Or can `(t *RawTemplate) Resolve(..)` be dropped?

internal/integrations/rawtemplate.go
```diff
@@ -0,0 +1,241 @@
...
+	return confmap.NewResolver(settings)
+}
+
+type selectableID interface {
+	comparable
+	fmt.Stringer
+}
+
+func selectComponents[C any, ID selectableID](from map[ID]C, selection []ID) (map[ID]C, error) {
+	selected := make(map[ID]C)
```
florianl, line 112:
```suggestion
	selected := make(map[ID]C, len(selection))
```

For the `!found` case, the allocated map is dropped anyway. So we can assume, that `len(selected)` is always `len(selection)`.

internal/integrations/rawtemplate.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+
+		for _, processor := range pipeline.Processors {
+			if _, found := c.Processors[processor]; !found {
+				return fmt.Errorf("processor %q not defined", processor.String())
+			}
+
+		}
+	}
+
```
florianl, line 171: Should there be an error, if `len(c.Pipelines)` is `0`?

internal/integrations/confmap.go
```diff
@@ -0,0 +1,68 @@
...
+func createVariablesProvider(variables map[string]any) confmap.CreateProviderFunc {
+	return func(_ confmap.ProviderSettings) confmap.Provider {
+		return &variablesProvider{variables: variables}
+	}
+}
+
+type variablesProvider struct {
+	variables map[string]any
+}
+
```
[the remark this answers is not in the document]
jsoriano: Checks added.

internal/integrations/integrations.go
```diff
@@ -0,0 +1,138 @@
...
+	Resolve(ctx context.Context, params map[string]any, pipelines []pipeline.ID) (*Config, error)
+}
+
+// Finder is the interface for extensions that can be used to find integrations by their names.
+type Finder interface {
+	FindIntegration(ctx context.Context, name string) (Integration, error)
+}
+
+// Host is the interface a host must implement to be used to locate finders along extensions.
+type Host interface {
```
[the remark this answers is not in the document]
jsoriano: `Host` is a subset of the `component.Host` interface, limited to the only method we use here, `GetExtensions`.

Renamed to `ExtensionGetter`, and added compile-time check to ensure that is a subset of `component.Host`.

internal/integrations/integrations.go
jsoriano: Yes, for users the error mentions that no extension is found.

I have changed it to a named `ErrNotConfigured` error, but still wrapping `ErrNotFound`, so components can just check if the integration is not found, while they propagate the error.

internal/integrations/integrations.go
jsoriano: Makes sense, yes, added.

internal/integrations/integrations.go
jsoriano: No interface implemented that I know, dropped as private methods.

I was using it like this because I saw it in other places, but no idea if they implement some interface.

internal/integrations/integrations.go
jsoriano: Yes, I think this was a leftover, removed.

internal/integrations/rawtemplate.go
jsoriano: Ok, changed.

internal/integrations/rawtemplate.go
jsoriano: :+1: done.

internal/integrations/rawtemplate.go
jsoriano: Good idea, done.

internal/integrations/rawtemplate.go
```diff
@@ -0,0 +1,243 @@
...
+	default:
+		return nil, fmt.Errorf("%s not found", uri)
+	}
+
+}
+
+func (p *rawConfigProvider) buildMainYML() []byte {
+	var buf bytes.Buffer
+	if len(p.main.Receivers) > 0 {
+		fmt.Fprintf(&buf, "receivers: ${%s:receivers}\n", configProviderScheme)
```
florianl, line 226: When looking at this output, I'm a bit confused.
```suggestion
		fmt.Fprintf(&buf, "receivers: ${config:receivers}\n")
```

The expected output is not defined, so I'm just confused. Would it make sense to print the number of receivers here instead of hard coding `config`?

internal/integrations/integrations.go
```diff
@@ -0,0 +1,142 @@
...
+	// GetExtensions returns the map of extensions. Only enabled and created extensions will be returned.
+	// GetExtensions can be called by the component anytime after Component.Start() begins and
+	// until Component.Shutdown() ends.
+	GetExtensions() map[component.ID]component.Component
+}
+
+// Compile-time check to ensure that ExtensionGetter is a subset of the component.Host interface.
+var _ ExtensionGetter = (component.Host)(nil)
+
+// Find looks for integrations in extensions of the host that implement the IntegrationFinder interface.
```
florianl, line 63:
```suggestion
// Find looks for integrations in extensions of the host that implement the IntegrationFinder interface.
// If multiple integrations match, only the first one is returned.
```

internal/integrations/integrations.go
```diff
@@ -0,0 +1,142 @@
...
+	// Receiver is the receiver to be used in the pipeline. It is optional, if a pipeline does not define a
+	// receiver it cannot be used to instantiate receivers.
+	Receiver *component.ID `mapstructure:"receiver"`
+
+	// Processors is the chain of processors of the pipeline, to be used as part of the receiver in receiver
+	// components, or as a combined processor when the pipeline is used as processor.
+	Processors []component.ID `mapstructure:"processors"`
+}
+
+// ComponentConfig contains the configuration of components. It is mainly irrelevant at this point.
```
florianl, line 141: This comment needs adjustment.

If it is irrelevant, then it should not be used in other places like this:

```
type Config struct {
	Receivers  map[component.ID]ComponentConfig `mapstructure:"receivers"`
	Processors map[component.ID]ComponentConfig `mapstructure:"processors"`
	Pipelines  map[pipeline.ID]PipelineConfig   `mapstructure:"pipelines"`
}
```

florianl (approved):

internal/integrations/rawtemplate.go
jsoriano: Actually `p.main` can be directly encoded, not sure what I was thinking about when writing this. Done in 093451c5cc23ead6a36fa595d26896444de0fd5d.

internal/integrations/integrations.go
jsoriano: Comments adjusted in 4d97f7e5e73d99b06d736ed64214c1781978b634, thanks.

jsoriano:
If there are no more change requests I think this could be merged. I don't have permissions, could you @constanca-m or @florianl click the button? Thanks!

Once merged I will continue opening PRs for the other components in https://github.com/elastic/opentelemetry-collector-components/pull/96

PR #470, opened 2025-03-21
Add extensions to find integrations

jsoriano:
_This is part of https://github.com/elastic/opentelemetry-collector-components/pull/96, more details there._

Add two new extensions:
- `config_integrations` can be used to embed integration templates in the collector configuration.
- `file_integrations` look for integrations in files in a given path.

These extensions implement the `integrations.Finder` interface, and can be used by components that make use of integrations.


extension/configintegrationextension/config.go
```diff
@@ -0,0 +1,47 @@
...
+
+package configintegrationextension // import "github.com/elastic/opentelemetry-collector-components/extension/configintegrationextension"
+
+import (
+	"fmt"
+	"strings"
+
+	"go.opentelemetry.io/collector/confmap/xconfmap"
+
+	"github.com/elastic/opentelemetry-collector-components/internal/integrations"
```
dmathieu, line 26: Depending on an internal package of a different module is usually a rather bad idea (see https://github.com/open-telemetry/opentelemetry-go/issues/3846).

jsoriano: Happy to change the approach to share this code.
Would it make sense to move it to a public package under this repository?
Something like `github.com/elastic/opentelemetry-collector-components/pkg/integrations`?

I guess another alternative also mentioned in the linked issue would be to copy the code with gogenerate, but having it in a different [public] package would sound better to me.

dmathieu: I think a different public package seems better.

jsoriano: Ok, moving it in https://github.com/elastic/opentelemetry-collector-components/pull/484.

jsoriano: Updated.

extension/configintegrationextension/config.go
```diff
@@ -0,0 +1,47 @@
...
+	Integrations map[string]string `mapstructure:"integrations"`
+}
+
+var _ xconfmap.Validator = &Config{}
+
+// Validate checks that the embedded integrations are syntactically valid.
+func (c *Config) Validate() error {
+	for name, integration := range c.Integrations {
+		raw := []byte(strings.ReplaceAll(integration, `$$`, `$`))
+		if err := integrations.ValidateRawTemplate(raw); err != nil {
```
dmathieu, line 41: I wonder if this util method is useful. You can do the same with the existing API:

```suggestion
		if _, err := integrations.NewRawTemplate(raw); err != nil {
```

dmathieu (approved):
LGTM overall.

dmathieu (approved):

PR #472, opened 2025-03-21
Add integration receiver

jsoriano:
_This is part of https://github.com/elastic/opentelemetry-collector-components/pull/96, more details there._

Integrations can be used to create new components based on configuration, by composing other components.

With this integration receiver, a receiver can be created by composing receivers and chains of processors for each receiver.


pkg/integrations/rawtemplate.go
```diff
@@ -135,11 +135,14 @@ func listComponents(pipelines map[pipeline.ID]PipelineConfig, listIDs func(Pipel
 // listReceivers lists the IDs of all the receivers found in pipelines
 func listReceivers(pipelines map[pipeline.ID]PipelineConfig) []component.ID {
 	return listComponents(pipelines, func(c PipelineConfig) []component.ID {
+		if c.Receiver == nil {
```
dmathieu, line 138: This behavior doesn't seem tested. Should it be? (I know this was also in your other PR. Just thinking about it now).

dmathieu (approved):

pkg/integrations/rawtemplate.go
jsoriano: This is covered by the tests with pipelines without receivers. These tests panic without this check.

PR #486, opened 2025-03-26
Group all dependabot directories

lahsivjar:
There have been multiple breaking changes on tools like `mdatagen` which has made it incredibly difficult to keep the versions updated as a breaking change in mdatagen requires all components to update together but our dependabot settings doesn't allow that. So, I am bringing back the proposal in https://github.com/elastic/opentelemetry-collector-components/pull/267/files.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,15 +1,9 @@
 ---
 version: 2
 updates:
-  # TODO: Replace go directories with wildcard https://github.com/dependabot/dependabot-core/issues/2178
-  # All the directories are grouped by ownership
-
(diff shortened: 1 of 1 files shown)


lahsivjar:
@rogercoll Sorry, I did not get that. We already have the otel group dependencies, I think we still require the `directories` to allow for multiple modules, right?

rogercoll:
> @rogercoll Sorry, I did not get that. We already have the otel group dependencies, I think we still require the `directories` to allow for multiple modules, right?

Ah, you are right, I missed on my first review.

rogercoll (approved):

Issue #459, opened 2025-03-17
[processor/lsminterval] Optimize merging for exponential histogram


lahsivjar:
Merging for exponential histograms is quite costly with the current implementation due to which we have updated the `elasticapmconnector` to use explicit bound histogram: https://github.com/elastic/opentelemetry-collector-components/pull/367. However, explicit bounds histogram provide poor resolution, especially for values that are outside the range for which they are designed.

There are a few ideas for optimizing exponential histograms:

1. Use https://github.com/lightstep/go-expohisto and introduce methods to marshal/unmarshal it to/from binary. With this, we can encode the exponential histograms as go-expohisto models and utilize the max bucket to limit and optimize exponential histogram merges. Max buckets will be a configuration exposed in the LSM interval processor.
2. Build our own version of `pdata` marshalling internally to the LSM interval processor that is faithful to OTLP but more optimised for our needs. Then if there's an opportunity to generalize some of it, we could propose enhancements to the main `pdata` package.

lahsivjar:
Getting in a bit of a mental block on this issue so I am putting it on hold for a bit as I work through some other tasks.

PR #497, opened 2025-03-28
Update expo from upstream, add logic for bucket limits

lahsivjar:
The PR updates the expohisto logic from upstream delta to cummultive processor - adds the logic to limit the max number of buckets. It also adds a config to the lsm interval to take the max buckets as a configuration.

Note that this is still quite far away from moving to exponential histogram as it still has poor performance.

Related to: https://github.com/elastic/opentelemetry-collector-components/issues/459


processor/lsmintervalprocessor/config/config.go
```diff
@@ -27,6 +27,10 @@ import (
 
 var _ component.Config = (*Config)(nil)
 
+const (
+	defaultMaxExponentialHistogramBuckets = 160
```
vigneshshanmugam, line 31: Was it just to match with the spec - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#base2-exponential-bucket-histogram-aggregation ? 

Can we add a link to call that out explicitly?

axw (approved):
Looks great!

PR #498, opened 2025-03-28
Add config `metrics::add_counter_attr` (default=true) to workaround TSDB version conflict by adding loadgenreceiver_counter resource attribute

carsonip:
AddCounterAttr, if true, adds a loadgenreceiver_counter resource attribute containing increasing counter value to the generated metrics.
It can be used to workaround timestamp precision and duplication detection of backends,
e.g. Elasticsearch TSDB version_conflict_engine_exception with millisecond-precision timestamp mapping,
which will be triggered when loadgenreceiver generates metrics too quickly such that
there exists data points of the same metric with the same dimensions within a millisecond but different nanoseconds.
Defaults to true.

Alternative to #493 but with little to no perf overhead.


vigneshshanmugam (approved):
LGTM

PR #499, opened 2025-03-28
Remove Abs struct and let compiler and runtime optimizations work

lahsivjar:
Optimize exponential histograms, below is the CPU profile before this PR:

Screenshot 2025-03-28 at 3 25 04 PM

The `Abs` struct is wasteful and casues a lot of CPU usage to do simple operations.

To be merged after: https://github.com/elastic/opentelemetry-collector-components/pull/497

Closes: https://github.com/elastic/opentelemetry-collector-components/issues/459


lahsivjar:
Based on benchmarks after https://github.com/elastic/opentelemetry-collector-components/pull/497/files, the current PR is not required as the numbers are similar even without this PR.

PR #500, opened 2025-03-31
Make elasticinframetricsprocessor idempotent

axw:
elasticinframetricsprocessor now sets "otel_remapped" as a scope attribute to prevent duplicate metrics if the processor is used twice in a pipeline.

This PR also simplifies the drop_original path. Instead of copying the remapped metrics to new slice after remapping, we now add the remapped metrics to a new slice while remapping, and then replace the original metrics slice. This leads to a small functional change: we now retain the scope schema URL, and we update the scope name & version to the elasticinframetricsprocessor's scope & version -- but only in the case where we drop original metrics, otherwise we continue to append to the original metric slice and leave the scope name/version alone.


simitt (approved):
Changes LGTM

lahsivjar (commented):
Code LGTM! But I do have a concern with doing this in otel-lib, added it to the tracking issue: https://github.com/elastic/opentelemetry-lib/issues/19#issuecomment-2766457809

PR #515, opened 2025-04-03
otelbench: fix benchmark name format to be compliant with x/tools/benchmark

1pkg:
Fix loadgen/otelbench benchmark name format so it's compliant with [x/tools/benchmark](https://cs.opensource.google/go/x/tools/+/refs/tags/v0.31.0:benchmark/parse/parse.go;l=48).


vigneshshanmugam (approved):

carsonip (commented):
nice catch, thanks

PR #522, opened 2025-04-10
[IntakeV2 receiver] Mapping APM Errors to LogRecord

gregkalapos:
Mapping error events from the IntakeV2 protocol to OTel LogRecords representing and error.

These are from a classic Elastic APM Agent via intakeV2 receiver + OTel collector in an OTel data stream:

Screenshot 2025-04-10 at 11 24 49

### DIFF
--- a/receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
+++ b/receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
@@ -133,11 +133,14 @@ func SetDerivedFieldsForError(event *modelpb.APMEvent, attributes pcommon.Map) {
 		if event.Error.Exception.Stacktrace != nil {
 			str := ""
 			for _, frame := range event.Error.Exception.Stacktrace {
-				if frame.Function != "" {
-					str += fmt.Sprintf("%s:%d %s \n", frame.Filename, *frame.Lineno, frame.Function)
+				f := frame.GetFunction()
+				l := frame.GetLineno()
+				if f != "" {
+					str += fmt.Sprintf("%s:%d %s \n", frame.GetFilename(), l, f)
 				} else {
-					if frame.Classname != "" && frame.Lineno != nil {
-						str += fmt.Sprintf("%s:%d \n", frame.Classname, *frame.Lineno)
+					c := frame.GetClassname()
+					if c != "" && l != 0 {
+						str += fmt.Sprintf("%s:%d \n", c, l)
 					}
 				}
 			}


receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
```diff
@@ -89,3 +90,69 @@ func SetDerivedFieldsCommon(event *modelpb.APMEvent, attributes pcommon.Map) {
...
+
+	if event.Error.Exception != nil {
+		if event.Error.Exception.Type != "" {
+			attributes.PutStr("exception.type", event.Error.Exception.Type)
+		}
+		if event.Error.Exception.Message != "" {
+			attributes.PutStr("exception.message", event.Error.Exception.Message)
+		}
+
+		if event.Error.Exception.Stacktrace != nil {
```
gregkalapos, line 133: We'll need to spend more time on this.

In the Elastic APM data model, the stacktrace is structured and at this point we only have the parsed stacktrace in the APM format. Kibana has a logic to show this in a native format of the given language. In OTel stacktrace is just a string.

I suggest we start with some meaningful default here for all stack traces regardless of the language. And then we can iterate on this in follow up PRs.

receiver/elasticapmreceiver/receiver.go
```diff
@@ -248,7 +249,18 @@ func (r *elasticAPMReceiver) processBatch(ctx context.Context, batch *modelpb.Ba
...
 			}
 		case modelpb.ErrorEventType:
-			// TODO
+			rl := ld.ResourceLogs().AppendEmpty()
+			gk := modelprocessor.SetGroupingKey{
+				NewHash: func() hash.Hash {
+					return xxhash.New()
+				},
+			}
+			if err := gk.ProcessBatch(ctx, batch); err == nil {
```
rogercoll, line 258: Could there be more than one error even type in a batch? Looks like we will be processing the batch multiple times. Maybe adding a `processedBatch` guard? Or exposing the [processError method](https://github.com/elastic/apm-data/blob/main/model/modelprocessor/groupingkey.go#L45)}

receiver/elasticapmreceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
```diff
@@ -0,0 +1,116 @@
...
+	"go.uber.org/zap"
+)
+
+type TopLevelFieldSetter interface {
+	SetTraceID(v pcommon.TraceID)
+	SetSpanID(v pcommon.SpanID)
+}
+
+// Shared across LogRecord and Span
+func SetTopLevelFieldsCommon(event *modelpb.APMEvent, timestamp time.Time, t TopLevelFieldSetter, logger *zap.Logger) {
```
rogercoll, line 38: Is the `timestamp` parameter needed here?

receiver/elasticapmreceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
```diff
@@ -0,0 +1,116 @@
...
+
+	if event.Trace != nil && event.Trace.Id != "" {
+		traceId, err := traceIDFromHex(event.Trace.Id)
+		if err == nil {
+			t.SetTraceID(traceId)
+		} else {
+			logger.Error("failed to parse trace ID", zap.String("trace_id", event.Trace.Id))
+		}
```
rogercoll, line 46: Should we continue processing the event in case of error? Should it be a `Warn` or lower log level?

In case of being an error, wdyt of fail fast and remove the logger parameter?


```suggestion
		traceId, err := traceIDFromHex(event.Trace.Id)
		if err != nil {
		     return fmt.Errorf("failed to parse trace ID: %w", err)
		} 
		t.SetTraceID(traceId)
```

receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
```diff
@@ -89,3 +90,69 @@ func SetDerivedFieldsCommon(event *modelpb.APMEvent, attributes pcommon.Map) {
 		attributes.PutStr(elasticattr.EventOutcome, "unknown")
 	}
 }
+
+// Sets fields that are NOT part of OTel for errors. These fields are derived by the Enrichment lib in case of OTLP input
+func SetDerivedFieldsForError(event *modelpb.APMEvent, attributes pcommon.Map) {
```
rogercoll, line 95: nit: an anonymous function could be used to wrap the string to nil comparison and attributes setter, similar to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/geoipprocessor/internal/provider/maxmindprovider/provider.go#L71

receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
rogercoll: > In OTel stacktrace is just a string.

Do you know if OTel defines any specific string format for stacktraces?

receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
```diff
@@ -89,3 +90,69 @@ func SetDerivedFieldsCommon(event *modelpb.APMEvent, attributes pcommon.Map) {
...
+		}
+
+		if event.Error.Exception.Stacktrace != nil {
+			str := ""
+			for _, frame := range event.Error.Exception.Stacktrace {
+				if frame.Function != "" {
+					str += fmt.Sprintf("%s:%d %s \n", frame.Filename, *frame.Lineno, frame.Function)
+				} else {
+					if frame.Classname != "" && frame.Lineno != nil {
+						str += fmt.Sprintf("%s:%d \n", frame.Classname, *frame.Lineno)
```
rogercoll, line 140: Are the trailing spaces needed?

receiver/elasticapmreceiver/receiver_test.go
```diff
@@ -332,6 +334,42 @@ func TestAgentCfgHandler(t *testing.T) {
 	}
 }
 
+var inputFiles_error = []struct {
```
rogercoll, line 337: nit: wdyt about moving this variable within the `TestErrors` func?

receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
rogercoll: Semconv seems to rely on languages SIG's: https://opentelemetry.io/docs/specs/semconv/attributes-registry/exception/

gregkalapos: Right - and the problem is, at this point we lost the language specific format, because it's already parsed in to the APM data format :(

receiver/elasticapmreceiver/receiver.go
gregkalapos: Ah, this is a good catch, this should not be in the for loop.

I moved this out before the loop - I think with that we are good.

What this does is that this generates the grouping key on errors - if this fails, the result is that there won't be a grouping key. But everything else would work. I think current check with printing an error is fine - this will cause issues by Kibana not being able to group errors on the APM UI - on the other hand, everything else will be still visible on the document in es.

receiver/elasticapmreceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
gregkalapos: No, you are right - removed 👍

receiver/elasticapmreceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
gregkalapos: I think this was discussed in a previous PR. So there is no enforcement for these ids after this point anywhere. So if parsing fails, there will be still error, transaction, and span documents stored in elasticsearch with all the remaining fields. Of course the value of such documents won't be very useful - correlating events won't work. That's why we went with an error level log.

On the other hand, I'd not stop the processing. Reason we ended up with this is that if we continue here, there will be still documents in elasticsearch - other fields will be set, plus we print an error. This is probably a better experience for users than potentially not storing anything in elasticsearch.

receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
gregkalapos: Yeah, this has a history :) 

Originally I did not have those, but then golden formatted the generated yaml files in a way that never matched the actual data in `plogtest.CompareLogs`.  It was super strange. So that's why I ended up with this.

receiver/elasticapmreceiver/receiver_test.go
gregkalapos: Ok

receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
felixbarny: There's some logic in Kibana to re-create the stack trace in a language specific way. We could do a similar thing here. But I agree that we should start with something simple that works generically for all languages and iterate if needed.

receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
rogercoll: Great, thanks for the context!

A follow-up suggestion would be to use the `protobufs` auto generated getters for non-checked pointer values (if they could be nil), something like:

```go
				if frame.Function != "" {
					str += fmt.Sprintf("%s:%d %s \n", frame.GetFilename(), frame.GetLineno(), frame.Function)
				} else {
					if frame.Classname != "" && frame.Lineno != nil {
						str += fmt.Sprintf("%s:%d \n", frame.Classname, *frame.Lineno)
					}
				}
```

receiver/elasticapmreceiver/internal/mappers/intakeV2ToDerivedFields.go
gregkalapos: Opened https://github.com/elastic/opentelemetry-collector-components/issues/526 as a follow up.

rogercoll (approved):
lgtm

seems Go version needs an upgrade to `1.23.8` in `./.github/workflows/checks.yml` for the security CI to pass

receiver/elasticapmreceiver/receiver.go
```diff
@@ -248,7 +260,8 @@ func (r *elasticAPMReceiver) processBatch(ctx context.Context, batch *modelpb.Ba
 				}
 			}
 		case modelpb.ErrorEventType:
-			// TODO
+			rl := ld.ResourceLogs().AppendEmpty()
```
jsvd, line 263: @gregkalapos quick question related to:
```go
	rl := ld.ResourceLogs().AppendEmpty()
        // [...]
	sl := rl.ScopeLogs().AppendEmpty()
	l := sl.LogRecords().AppendEmpty()
```

Given `processBatch` takes in `modelpb.Batch` which may contain many events of type ErrorEventType about the same Resource and even same Scope, did you consider building a plog.Logs that groups LogRecords under the same Scope and Resource?
I don;t think it matters much for the final destination, but could help processors that work on batches and iterate over the two sub "collections".

PR #523, opened 2025-04-11
elasticapmconnector: allow setting custom lsm intervals

1pkg:
This PR is exposing lsm intervals configuration to `elasticapmconnector` which makes it possible to configure custom aggregation intervals. This is useful in integration testing where the default intervals are not practical to use. If custom intervals were not provided the config will default to the old intervals values.

### DIFF
--- a/connector/elasticapmconnector/config.go
+++ b/connector/elasticapmconnector/config.go
@@ -18,6 +18,7 @@
 package elasticapmconnector // import "github.com/elastic/opentelemetry-collector-components/connector/elasticapmconnector"
 
 import (
+	"fmt"
 	"time"
 
 	lsmconfig "github.com/elastic/opentelemetry-collector-components/processor/lsmintervalprocessor/config"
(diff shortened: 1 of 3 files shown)


vigneshshanmugam (approved):
LGTM

connector/elasticapmconnector/config.go
```diff
@@ -46,6 +53,15 @@ type AggregationConfig struct {
...
 	// a validation error.
 	MetadataKeys []string `mapstructure:"metadata_keys"`
+
+	// Intervals holds an optional list of time intervals that the processor
+	// will aggregate over. The interval duration must be in increasing
+	// order and must be a factor of the smallest interval duration.
+	// The default aggregation intervals are 1m, 10m and 60m.
+	// These intervals should only be overridden for testing purposes when
+	// faster processor feedback is required, the default intervals
+	// should be always prefered otherwise.
```
axw, line 63:
```suggestion
	//
	// NOTE: these intervals should only be overridden for testing purposes when
	// faster processor feedback is required. The default intervals should be preferred
	// in all other cases -- using this configuration may lead to invalid behavior,
	// and will not be supported.
```

axw (approved):

carsonip (approved):

PR #527, opened 2025-04-15
[IntakeV2 receiver] Map Span links

gregkalapos:


trentm (approved):
LGTM, from the span links data perspective. However, I haven't reviewed any code in this repo before, so take with a grain of salt.

rogercoll (approved):

Issue #517, opened 2025-04-07
Adopt upstream collector release cycle


rogercoll:
- Add workflow to release all components at once
- Ensure all components use the same upstream core packages

# Open questions

- Should we tag components with the same upstream version? At the moment, the components do not follow upstream versions and are released independently.

lahsivjar:
> Should we tag components with the same upstream version? At the moment, the components do not follow upstream versions and are released independently.

+1 on having elastic components follow upstream versioning. Not having versioning with upstream means that we might not have all the elastic components tagged with a version corresponding to the upstream release version. For example, not all components in our internal repository have a version with v0.121.0 upstream version. I think this might lead to some issues down the line with breaking changes.

vigneshshanmugam:
Don't have a strong opinion on following the upstream versioning, But we should have the ability to release all and also individual components when required at our own pace, to help with Managed collector releases etc.

lahsivjar:
> But we should have the ability to release all and also individual components when required at our own pace, to help with Managed collector releases etc.

Good point! We would need to accommodate this in our version strategy, so there are 2 dimensions:

1. Upstream version to track which upstream version is used by the elastic components. 
2. MOTel specific versioning to track new features and stuff.

A bit of a brain dump, the simplest approach would be to align with upstream versioning schedule (I think it is biweekly) to release/tag elastic components, but that might be too restrictive for MOTel. Another solution would be to have a flat versioning from `main` in addition to following upstream versioning, something like:

1. Release a new version whenever upstream version is bumped up.
2. For MOTel and CD, have a versioning schema (for example: `<upstream_version>-<incremental_id>`) and create a new version ALWAYS from `main` based on the upstream version used in the main at the time of creation.

vigneshshanmugam:
> For MOTel and CD, have a versioning schema (for example: <upstream_version>-<incremental_id>) and create a new version ALWAYS from main based on the upstream version used in the main at the time of creation.

I like this idea.

rogercoll:
Sounds like a plan, I also like the idea of incremental_id version (very similar to what we do on the otel-demo fork). We will soon work on it.

rogercoll:
> For MOTel and CD, have a versioning schema (for example: <upstream_version>-<incremental_id>) and create a new version ALWAYS from main based on the upstream version used in the main at the time of creation.

[...]

PR #544, opened 2025-04-23
Multimod release

rogercoll:
Adopts a strategy similar to upstream (collector) to release multiple Go modules. New Make targets:

- `update-otel`: syncs otel dependencies for all modules
- `push-tags`: pushes a new tag (release) for each modulde defined in the `version.yaml` file

Docs: A release guide has been added as `docs/release.md`.

Fixes https://github.com/elastic/opentelemetry-collector-components/issues/517


versions.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - github.com/elastic/opentelemetry-collector-components/internal/tools
+  - github.com/elastic/opentelemetry-collector-components/internal/sharedcomponent
+  - github.com/elastic/opentelemetry-collector-components/internal/testutil
+  - github.com/elastic/opentelemetry-collector-components/pkg/integrations
+  - github.com/elastic/opentelemetry-collector-components/receiver/integrationreceiver
+  - github.com/elastic/opentelemetry-collector-components/extension/fileintegrationextension
+  - github.com/elastic/opentelemetry-collector-components/extension/configintegrationextension
+  - github.com/elastic/opentelemetry-collector-components/extension/beatsauthextension
+  - github.com/elastic/opentelemetry-collector-components/extension/apmconfigextension
+  - github.com/elastic/opentelemetry-collector-components/processor/integrationprocessor
```
lahsivjar, line 23: Curious, what is the reasoning behind excluding these modules?

docs/release.md
```diff
@@ -0,0 +1,44 @@
+# Releasing opentelemetry-collector-components
+
+0. Clone the [`https://github.com/open-telemetry/opentelemetry-collector`](https://github.com/open-telemetry/opentelemetry-collector)
+   repository (or ensure it is up-to-date) one directory above the components
+   repository (`../../opentelemetry-collector`).
```
lahsivjar, line 5: **nit**: Would it not be easier for the script to clone the upstream repository in a build folder or something based on the parameters passed for the upstream version?

versions.yaml
```diff
@@ -0,0 +1,25 @@
+module-sets:
+  edot-base:
+    version: v0.124.0
```
ChrsMark, line 3: Will we have to regularly release for every upstream version that is released? If so how this will be performed? Upstream Collector has a specific release schedule with round robin assignment, are we going to have specific teams/people handling this?

rogercoll: Yes, the plan is to be handled by the SDH "hero" of the `ingest-otel-data` team 

cc @sboomsma

versions.yaml
rogercoll: I am not aware of their usage outside this repository, nor have been released in the past https://github.com/elastic/opentelemetry-collector-components/tags

I would add them to the list as needed, would you prefer to release them all?

lahsivjar: I was not aware that they are not used outside. Sounds good to me to release as needed.

ChrsMark (approved):
LGTM

docs/release.md
rogercoll: I like the idea, let me know what you think of https://github.com/elastic/opentelemetry-collector-components/pull/544/commits/e195d3a3f8a3baaa4cafd3a8b111bbd3caaf822c

rogercoll:
Note: the command `make update-otel OTEL_VERSION=v0.124.0 OTEL_STABLE_VERSION=v1.30.0` will fail due to some upstream packages defined in the `versions.yaml` file but not release yet (e.g https://github.com/open-telemetry/opentelemetry-collector/commit/46101260f6ce692c082919d02d3620032c0d3f33)

lahsivjar (approved):
LGTM 🎉

Issue #398, opened 2025-02-28
[extension/apmconfig] Handle on AgentDisconnect message


rogercoll:
- Clean internal config catch for the corresponding service

PR #554, opened 2025-04-29
feat: deployment.environment.name instead of deprecated deployment.environment

rogercoll:
The goal of this PR was to remove entries from the central configuration fetcher map `uidToService map[string]agentcfg.Service`, but the agent's service catching logic was moved to the already existing sync.Map in the OpAMP callbacks → the fetcher does not need to deal with AgentDisconnect messages

- Agent's identifying attributes are now stored in the OpAMP callbacks structure.
- Unit and integration tests added for AgentDisconnect messages and new test file for the fetcher struct.
- Adopted the new `deployment.environment.name` instead of the deprecated `deployment.environment` attribute.

Fixes https://github.com/elastic/opentelemetry-collector-components/issues/398


rogercoll:
PTAL @elastic/ingest-otel-data

extension/apmconfigextension/elastic/centralconfig/fetcher.go
```diff
@@ -36,42 +35,28 @@ var _ apmconfig.RemoteConfigClient = (*fetcherAPMWatcher)(nil)
...
-		// only update the internal cache if service name is set
-		if serviceParams.Name != "" {
-			fw.uidToService[string(agentMsg.GetInstanceUid())] = serviceParams
+func (fw *fetcherAPMWatcher) RemoteConfig(ctx context.Context, agentUid apmconfig.InstanceUid, agentAttrs apmconfig.IdentifyingAttributes) (*protobufs.AgentRemoteConfig, error) {
+	var serviceParams agentcfg.Service
+	for _, attr := range agentAttrs {
+		switch attr.GetKey() {
+		case semconv.AttributeServiceName:
+			serviceParams.Name = attr.GetValue().GetStringValue()
+		case semconv.AttributeDeploymentEnvironment:
```
trentm, line 54: Is it (or should it be):
- `deployment.environment` (as used here),
- `service.environment` (as listed in the integration test below), or
- `deployment.environment.name` (the latest name in semconv, v1.27.0 deprecated `deployment.environment`: https://github.com/open-telemetry/semantic-conventions/blob/02681893953519164eefd086df9e20c7b15a159d/CHANGELOG.md?plain=1#L307)

rogercoll: Great question! The best option is to use the most up-to-date semantic convention (semconv) attribute that represents the deployment environment name. So, if all the OpenTelemetry SDKs we plan to support with this extension already use `deployment.environment.name`, then that’s the attribute we should go with.

I am going to change the expected value to the new `deployment.environment.name`, as well the integration tests and docs for clarity. Note that Kibana, either if the Otel SDK is providing `deployment.environment` or `deployment.environment.name`, it will store it in the apm config index with the `service.name` keyword (due classic SDKs). The following is an example of a Otel Java SDK config:

```json
{
  "took": 0,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 1,
      "relation": "eq"
    },
    "max_score": 1,
    "hits": [
      {
        "_index": ".apm-agent-configuration",
        "_id": "abc",
        "_score": 1,
        "_source": {
          "agent_name": "opentelemetry/java/elastic",
          "service": {
            "name": "dice-java-5",
            "environment": "prod-otel"
          },
          "settings": {
            "send_logs": "true",
            "send_traces": "true",
            "deactivate_instrumentations": "that-module-please"
          },
          "@timestamp": 1746175386213,
          "applied_by_agent": false,
          "etag": "abc"
        }
      }
    ]
  }
```

LikeTheSalad: We use `deployment.environment` for iOS and Android agents. The iOS one will switch from using Central Config to using OpAMP afaik. Should we add a case for `deployment.environment` to keep backwards compatibility? cc @bryce-b

rogercoll: > Should we add a case for deployment.environment to keep backwards compatibility?

If we plan to support OpAMP for SDKs that still use the deprecated `deployment.environment`, we should ensure backward compatibility. If this applies to iOS and/or Android, I will include the necessary support.

LikeTheSalad: Got it, thanks. I think for Android we can just start sending the new attribute when we enable OpAMP, since its current version doesn't support central config anyway. For iOS, I think we could send both attributes so that people can use the same build with both central config and OpAMP? Today we have the agent's weekly meeting, let me ask for some advice there in case I might me missing something.

rogercoll: Sounds good @LikeTheSalad. Because exposing an OpAMP endpoint, this extension should only be used for OTel SDKs (legacy SDKs can use the [elasticapmreceiver](https://github.com/elastic/opentelemetry-collector-components/tree/main/receiver/elasticapmreceiver) which exposes an intake v2 endpoint), does the iOS OTel SDK use central config already?

LikeTheSalad: > does the iOS OTel SDK use central config already?

It does.

> legacy SDKs can use the [elasticapmreceiver](https://github.com/elastic/opentelemetry-collector-components/tree/main/receiver/elasticapmreceiver) which exposes an intake v2 endpoint

Does this mean that elasticapmreceiver can act as an APM Server replacement, in that it provides central config too?

LikeTheSalad: Based on the weekly meeting, the new SDKs should use the new name when adding support for OpAMP config. And I think that if the `elasticapmreceiver` will always be available, then I don't see why we should make the OpAMP route backwards compatible. In terms of an agent that should support both, the APM Server and the OTel Collector routes, I'm only aware of the iOS one, so I'll loop @bryce-b in just in case, though it seems like relying on the `elasticapmreceiver` should do.

ChrsMark (approved):
I'm not quite familiar with this code-base but changes LGTM.
We don't have way to add changelog entries but this one would need one right? If so, can we ensure the PR tittle and description clearly reflect the changes and if they are breaking or not?

rogercoll:
> We don't have a ay to add changelog entries but this one would need one right?

Right, I will add a comment about these changes on the first release. This component is not used externally yet (not included in EDOT collector)

> If so, can we ensure the PR tittle and description clearly reflect the changes and if they are breaking or not?

Good point, from the user's perspective the PR is a breaking change because it will rely on the `deployment.environment.name` resource attribute instead of `deployment.environment`. I will update the commit/title accordingly.

PR #562, opened 2025-05-02
[Ratelimit Processor] Instrument the ratelimiter service with telemetry metrics

gizas:
This is the initial implementation for https://github.com/elastic/hosted-otel-collector/issues/513

The code adds a newMetrics reader in the [ratelimitprocessor](https://github.com/elastic/opentelemetry-collector-components/blob/main/processor/ratelimitprocessor/gubernator.go).

At the moment we have implemented only `otelcol_ratelimit.requests` metric.Int64Counter
Also the newMetricsReader makes use of the global otel processor that we assume that is initialized upon processor usage.

To be reviewed:

- if err != nil:  WithReason = "request_error" and WithDecision = "accepted"
- if n := len(responses); n != 1: WithReason ="request_error" and WithDecision = "accepted"
- if n := len(responses)==1 and resp.GetError() != "": WithReason = "limit_error" and WithDecision ="accepted"
- if gubernator.Status_OVER_LIMIT: WithReason = "over_limit" and WithDecision ="throttled"
- Default: "WithReason" = "under_limit" and "WithDecision" ="accepted"

## Sample document
After running `make smoketest` from the `hosted-otel-collector` repo, I can see this document in Kibana:
```
{
  "_index": ".ds-metrics-apm.app.hosted_otel_collector-default-2025.05.15-000001",
  "_id": "z24a75YBM8MU2AZ6gfwv",
  "_version": 1,
  "_source": {
    "@timestamp": "2025-05-20T19:10:12.799Z",
    "agent": {
      "name": "otlp",
      "version": "unknown"
    },
    "data_stream": {
      "dataset": "apm.app.hosted_otel_collector",
      "namespace": "default",
      "type": "metrics"
    },
    "event": {
      "ingested": "2025-05-20T19:10:13.000Z"
    },
    "labels": {
      "orchestrator_cluster_name": "default",
      "orchestrator_deploymentslice": "",
      "orchestrator_environment": "default",
      "ratelimit_decision": "accepted",
      "reason": "under_limit",
      "x-elastic-project-id": "local"
    },
    "metricset": {
      "name": "app"
    },
    "numeric_labels": {
      "limit_threshold": 0
    },
    "observer": {
      "hostname": "apm-server-apm-server-69868958cf-4rcgq",
      "type": "apm-server",
      "version": "8.18.1"
    },
    "otelcol_ratelimit": {
      "requests": 6
    },
    "service": {
      "framework": {
        "name": "github.com/elastic/opentelemetry-collector-components/processor/ratelimitprocessor"
      },
      "language": {
        "name": "unknown"
      },
      "name": "hosted-otel-collector",
      "node": {
        "name": "9f735ea3-a83e-4098-ad1a-8b289b8f0a9e"
      },
      "version": "git"
    }
  }
}
```


processor/ratelimitprocessor/internal/telemetry/attributes.go
```diff
@@ -0,0 +1,80 @@
...
+
+// WithProtocol returns a protocol attribute with key.
+func WithProtocol(protocol string) attribute.KeyValue {
+	return attribute.String(protocolKey, protocol)
+}
+
+// WithOutcome returns a project ID attribute with key.
+func WithOutcome(outcome string) attribute.KeyValue {
+	return attribute.String(outcomeKey, outcome)
+}
```
lahsivjar, line 63: Do we need these? Similarly, I think compression code, status code, signal type will also not be required for the rate limit processor. CMIIW.

Instead, we can add `WithDecision`.

processor/ratelimitprocessor/ratelimiter.go
```diff
@@ -75,3 +90,18 @@ func getUniqueKey(ctx context.Context, metadataKeys []string) string {
 	}
 	return uniqueKey.String()
 }
+
+func newMetrics(mp metric.MeterProvider) (metrics, error) {
+	meter := mp.Meter("internal/ratelimit")
```
lahsivjar, line 95: I think better to use a generic one (like `ratelimitprocessor`) for all of rate limit processor metrics.

processor/ratelimitprocessor/gubernator.go
```diff
@@ -102,17 +116,32 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	})
 	if err != nil {
 		r.set.Logger.Error("error executing gubernator rate limit request", zap.Error(err))
+		r.metrics.ratelimitRequests.Add(context.Background(), 1, metric.WithAttributeSet(attribute.NewSet(
+			telemetry.WithProjectID(projectID),
+			attribute.String("reason", "request_error"),
```
lahsivjar, line 121:
```suggestion
			telemetry.WithErrorReason("request_error"),
```

Or maybe have a generic one named `WithReason` and another one named `WithDecision`

processor/ratelimitprocessor/internal/telemetry/attributes.go
gizas: indeed we dont need them. I only use WithProtocol mainly. 
But I updated the reason and the decision as well

processor/ratelimitprocessor/ratelimiter.go
gizas: Sure changed it. No strong preferences.

processor/ratelimitprocessor/gubernator.go
gizas: I added `WithLimitReason`. Let me know if ok

processor/ratelimitprocessor/ratelimiter.go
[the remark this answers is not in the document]
gizas: Hmm no tbh @constanca-m ! Let me check how is applicable

processor/ratelimitprocessor/ratelimiter.go
```diff
@@ -51,6 +52,20 @@ type RateLimiter interface {
...
+}
+
+func getProjectIDFromMetadata(ctx context.Context, key string) string {
+	clientInfo := client.FromContext(ctx)
+	values := clientInfo.Metadata.Get(key)
+	if len(values) > 0 {
+		return values[0]
+	}
+	return "" // or a default value
+}
```
constanca-m, line 67:
```suggestion
func getProjectIDFromMetadata(ctx context.Context) (string, error) {
	clientInfo := client.FromContext(ctx)
	values := clientInfo.Metadata.Get(projectIDKey)
	if len(values) > 0 {
		return values[0], nil
	}
	return "", errors.New("project id not found") // or a default value
}
```

What do you think? Hardcode the key to be the elastic project id, and return error if not found

processor/ratelimitprocessor/ratelimiter.go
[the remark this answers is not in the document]
constanca-m: I tried locally, very shortly, like this:

- `metadata.yaml`:
```diff
diff --git a/processor/ratelimitprocessor/metadata.yaml b/processor/ratelimitprocessor/metadata.yaml
index 45e4a57..65f3b2a 100644
--- a/processor/ratelimitprocessor/metadata.yaml
+++ b/processor/ratelimitprocessor/metadata.yaml
@@ -10,3 +10,13 @@ tests:
   config:
     rate: 1
     burst: 1
+
+telemetry:
+  metrics:
+    ratelimit.requests:
+      enabled: true
+      description: Number of rate-limiting requests
+      unit: "{requests}"
+      sum:
+        value_type: int
+        monotonic: true
```

- `gubernator.go`:
```
diff --git a/processor/ratelimitprocessor/gubernator.go b/processor/ratelimitprocessor/gubernator.go
index 0d96ab0..bfbce83 100644
--- a/processor/ratelimitprocessor/gubernator.go
+++ b/processor/ratelimitprocessor/gubernator.go
@@ -21,12 +21,12 @@ import (
        "context"
        "errors"
        "fmt"
+       "github.com/elastic/opentelemetry-collector-components/processor/ratelimitprocessor/internal/metadata"
        "strings"
        "time"
 
        "go.opentelemetry.io/collector/component"
        "go.opentelemetry.io/collector/processor"
-       "go.opentelemetry.io/otel"
        "go.opentelemetry.io/otel/attribute"
        "go.opentelemetry.io/otel/metric"
        "go.uber.org/zap"
@@ -43,16 +43,15 @@ type gubernatorRateLimiter struct {
        set      processor.Settings
        behavior gubernator.Behavior
 
-       conn    *grpc.ClientConn
-       client  gubernator.V1Client
-       metrics metrics
+       conn             *grpc.ClientConn
+       client           gubernator.V1Client
+       telemetryBuilder *metadata.TelemetryBuilder
 }
 
 func newGubernatorRateLimiter(cfg *Config, set processor.Settings) (*gubernatorRateLimiter, error) {
        var behavior int32
 
-       globalmeterProvider := otel.GetMeterProvider()
-       m, err := newMetrics(globalmeterProvider)
+       telemetryBuilder, err := metadata.NewTelemetryBuilder(set.TelemetrySettings)
        if err != nil {
                return nil, err
        }
@@ -65,10 +64,10 @@ func newGubernatorRateLimiter(cfg *Config, set processor.Settings) (*gubernatorR
                behavior |= value
        }
        return &gubernatorRateLimiter{
-               cfg:      cfg,
-               set:      set,
-               behavior: gubernator.Behavior(behavior),
-               metrics:  m,
+               cfg:              cfg,
+               set:              set,
+               behavior:         gubernator.Behavior(behavior),
+               telemetryBuilder: telemetryBuilder,
        }, nil
 }
 
@@ -116,7 +115,7 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
        })
        if err != nil {
                r.set.Logger.Error("error executing gubernator rate limit request", zap.Error(err))
-               r.metrics.ratelimitRequests.Add(context.Background(), 1, metric.WithAttributeSet(attribute.NewSet(
+               r.telemetryBuilder.RatelimitRequests.Add(ctx, 1, metric.WithAttributeSet(attribute.NewSet(
                        telemetry.WithProjectID(projectID),
                        telemetry.WithLimitReason("request_error"),
                        telemetry.WithDecision("rejected"),
```

And then do the same for all other metrics. I think this should work.

Unfortunately, the attributes can't be set yet based on the `metadata.yaml`. I think this PR https://github.com/open-telemetry/opentelemetry-collector/pull/12182 includes that, I am not sure. But currently, I don't think it is possible, and we have to define like you were doing before.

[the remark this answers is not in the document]
constanca-m: Aah, you have to run `make generate` to generate the telemetry files that creater the meter and tracer.

processor/ratelimitprocessor/ratelimiter.go
gizas: Add it as well. I leave this comment open to review because now if we dont receive projectID then we return with error. I dont know if we want to be so strict. Maybe we can have the metrics but without enhancement?

gizas: Small update on this, I only set projectID "" when key not found. This was breaking other tests we had where we dont set the specific metadata key. I think we are ok by setting to empty

processor/ratelimitprocessor/gubernator.go
```diff
@@ -102,17 +115,32 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
...
+			telemetry.WithDecision("accepted"),
+		)))
 		return fmt.Errorf("expected 1 response from gubernator, got %d", n)
 	}
 	resp := responses[0]
 	if resp.GetError() != "" {
 		r.set.Logger.Error("failed to get response from gubernator", zap.Error(errors.New(resp.GetError())))
+		r.telemetryBuilder.RatelimitRequests.Add(context.Background(), 1, metric.WithAttributeSet(attribute.NewSet(
+			telemetry.WithProjectID(projectID),
+			telemetry.WithLimitReason("limit_error"),
```
vigneshshanmugam, line 141: How about using `WithErrorReason ` ?

processor/ratelimitprocessor/gubernator.go
```diff
@@ -102,17 +115,32 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
...
+			telemetry.WithProjectID(projectID),
+			telemetry.WithLimitReason("request_error"),
+			telemetry.WithDecision("accepted"),
+		)))
 		return fmt.Errorf("expected 1 response from gubernator, got %d", n)
 	}
 	resp := responses[0]
 	if resp.GetError() != "" {
 		r.set.Logger.Error("failed to get response from gubernator", zap.Error(errors.New(resp.GetError())))
+		r.telemetryBuilder.RatelimitRequests.Add(context.Background(), 1, metric.WithAttributeSet(attribute.NewSet(
```
vigneshshanmugam, line 139:
```suggestion
		r.telemetryBuilder.RatelimitRequests.Add(ctx, 1, metric.WithAttributeSet(attribute.NewSet(
```

processor/ratelimitprocessor/gubernator.go
```diff
@@ -102,17 +115,32 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	})
 	if err != nil {
 		r.set.Logger.Error("error executing gubernator rate limit request", zap.Error(err))
+		r.telemetryBuilder.RatelimitRequests.Add(ctx, 1, metric.WithAttributeSet(attribute.NewSet(
+			telemetry.WithProjectID(projectID),
+			telemetry.WithLimitReason("request_error"),
```
vigneshshanmugam, line 120: use constants declared from `attributes.go`

processor/ratelimitprocessor/ratelimiter.go
```diff
@@ -51,6 +51,20 @@ type RateLimiter interface {
 // high cardinality: tenant ID would be a good choice. For rate
 // limiting by IP (e.g. to avoid DDoS), consider running OpenTelemetry
 // Collector behind a WAF/API Gateway/proxy.
+
+const (
+	ProjectID = "x-elastic-project-id"
+)
+
+func getProjectIDFromMetadata(ctx context.Context) string {
```
vigneshshanmugam, line 59: We wanted to upstream this component at some point, so its better to abstract this out and add all keys passed via `metadata_keys` as attributes to the telemetry metrics.

processor/ratelimitprocessor/gubernator.go
```diff
@@ -39,12 +43,19 @@ type gubernatorRateLimiter struct {
...
-	client gubernator.V1Client
+	conn             *grpc.ClientConn
+	client           gubernator.V1Client
+	telemetryBuilder *metadata.TelemetryBuilder
 }
 
 func newGubernatorRateLimiter(cfg *Config, set processor.Settings) (*gubernatorRateLimiter, error) {
 	var behavior int32
+
+	telemetryBuilder, err := metadata.NewTelemetryBuilder(set.TelemetrySettings)
```
vigneshshanmugam, line 54: It would be good to add this also to the local rate limiter, to make sure we track all the incoming requests to this component. I would move this out.

vigneshshanmugam (commented):
The approach looks good to me. Some thoughts on additional metrics that we would need to introduce
- Request duration
- Can we measure the no of requests that are over the limits at a given time? I believe we can by using the decision dimension, just confirm if thats is the case. 
- Concurrent requests

@lahsivjar Could you think of additional metrics that were useful in MIS so we can add them here? Thanks.

processor/ratelimitprocessor/gubernator.go
gizas: Done!

processor/ratelimitprocessor/gubernator.go
gizas: Done although I used the errors per category a little on my own. Can I have any extra pair of eyes here to see if we all agree on the errors per category?

processor/ratelimitprocessor/ratelimiter.go
gizas: I added https://github.com/elastic/opentelemetry-collector-components/pull/562/files#diff-03ffdee6e5e78b42afe5d4f984f4bf495b5b10877d11593800c0a30abbd4e884R60 

WDYT?

processor/ratelimitprocessor/gubernator.go
```diff
@@ -102,17 +115,32 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	})
 	if err != nil {
 		r.set.Logger.Error("error executing gubernator rate limit request", zap.Error(err))
+		r.telemetryBuilder.RatelimitRequests.Add(ctx, 1, metric.WithAttributeSet(attribute.NewSet(
+			telemetry.WithProjectID(projectID),
+			telemetry.WithLimitReason("request_error"),
+			telemetry.WithDecision("rejected"),
```
kaiyan-sheng, line 120: @vigneshshanmugam I think `rejected` is the right one here but looking at apm code https://github.com/elastic/apm-ingest-service/blob/main/internal/ratelimit/ratelimit.go#L314 it's using `accepted` as the decision here.

processor/ratelimitprocessor/gubernator.go
kaiyan-sheng: @vigneshshanmugam Could you explain a bit more here please? Sorry I'm lost 😅

vigneshshanmugam: That was definitely without much context, apologeis. What I meant was to add it to https://github.com/elastic/opentelemetry-collector-components/blob/81b3c2413549f6e2d4257e708e7a6f5871031bc6/processor/ratelimitprocessor/local.go#L53 as well as there is way to disable gubernator and use local in-memory ratelimier instead.

processor/ratelimitprocessor/gubernator.go
vigneshshanmugam: To clarify, We should accept requests instead of rejecting the with the error when ratelimiter component was down for some reason, So MIS is doing the right thing here.

processor/ratelimitprocessor/gubernator.go
```diff
@@ -102,17 +115,35 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	})
 	if err != nil {
 		r.set.Logger.Error("error executing gubernator rate limit request", zap.Error(err))
+		attrs := []attribute.KeyValue{
+			telemetry.WithErrorReason(telemetry.BadRequest),
```
vigneshshanmugam, line 119: This is not a bad request, instead request/server error as the component was unavailable etc.

processor/ratelimitprocessor/gubernator.go
```diff
@@ -102,17 +115,35 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
...
+		attrs = attrsFromMetadata(ctx, r.cfg.MetadataKeys, attrs)
+		r.telemetryBuilder.RatelimitRequests.Add(ctx, 1, metric.WithAttributeSet(attribute.NewSet(attrs...)))
 		return fmt.Errorf("expected 1 response from gubernator, got %d", n)
 	}
 	resp := responses[0]
 	if resp.GetError() != "" {
 		r.set.Logger.Error("failed to get response from gubernator", zap.Error(errors.New(resp.GetError())))
+		attrs := []attribute.KeyValue{
+			telemetry.WithErrorReason(telemetry.ServerError),
+			telemetry.WithDecision("rejected"),
```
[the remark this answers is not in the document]
vigneshshanmugam: accepted.

processor/ratelimitprocessor/gubernator.go
```diff
@@ -102,17 +115,35 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	})
 	if err != nil {
 		r.set.Logger.Error("error executing gubernator rate limit request", zap.Error(err))
+		attrs := []attribute.KeyValue{
+			telemetry.WithErrorReason(telemetry.BadRequest),
+			telemetry.WithDecision("rejected"),
+		}
+		attrs = attrsFromMetadata(ctx, r.cfg.MetadataKeys, attrs)
```
vigneshshanmugam, line 122: nit: move this in to a common function.

processor/ratelimitprocessor/gubernator.go
```diff
@@ -126,17 +157,36 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
...
 			return errTooManyRequests
 		case ThrottleBehaviorDelay:
 			delay := time.Duration(resp.GetResetTime()-createdAt) * time.Millisecond
 			timer := time.NewTimer(delay)
 			defer timer.Stop()
 			select {
 			case <-ctx.Done():
+				attrs := []attribute.KeyValue{
+					telemetry.WithErrorReason(telemetry.StatusOverLimit),
+					telemetry.WithDecision("accepted"),
```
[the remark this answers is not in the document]
vigneshshanmugam: I dont have a strong opinion here, throttled or rejected both fits.

processor/ratelimitprocessor/gubernator.go
kaiyan-sheng: Agree, changing to `request_error`.

processor/ratelimitprocessor/gubernator.go
gizas: Done in [b478e1b](https://github.com/elastic/opentelemetry-collector-components/pull/562/commits/b478e1b668b2843d6c65f2f6a53742b8807a8c8a)

processor/ratelimitprocessor/gubernator.go
```diff
@@ -126,17 +157,36 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
...
+				r.telemetryBuilder.RatelimitRequests.Add(ctx, 1, metric.WithAttributeSet(attribute.NewSet(attrs...)))
 				return ctx.Err()
 			case <-timer.C:
 			}
 		}
 	}
+
+	attrs := []attribute.KeyValue{
+		telemetry.WithErrorReason(telemetry.StatusUnderLimit),
+		telemetry.WithDecision("accepted"),
```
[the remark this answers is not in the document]
gizas: @kaiyan-sheng in last commit I added telemetry.StatusUnderLimit ,which is mine. Feel free to change it

processor/ratelimitprocessor/gubernator.go
```diff
@@ -102,17 +115,35 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
...
+		attrs = attrsFromMetadata(ctx, r.cfg.MetadataKeys, attrs)
+		r.telemetryBuilder.RatelimitRequests.Add(ctx, 1, metric.WithAttributeSet(attribute.NewSet(attrs...)))
 		return errRateLimitInternalError
 	}
 
 	// Inside the gRPC response, we should have a single-item list of responses.
 	responses := getRateLimitsResp.GetResponses()
 	if n := len(responses); n != 1 {
+		attrs := []attribute.KeyValue{
+			telemetry.WithErrorReason(telemetry.Invalid),
```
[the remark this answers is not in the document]
gizas: Changed to request_error

processor/ratelimitprocessor/gubernator.go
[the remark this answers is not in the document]
kaiyan-sheng: I just added it in the latest commit, thanks!

processor/ratelimitprocessor/metadata.yaml
```diff
@@ -10,3 +10,31 @@ tests:
...
+    ratelimit.requests:
+      enabled: true
+      description: Number of rate-limiting requests
+      unit: "{requests}"
+      sum:
+        value_type: int
+        monotonic: true
+      attributes: ["project_id", "processor_id", "ratelimit_decision", "limit_threshold", "reason"]
+
+attributes:
```
vigneshshanmugam, line 25: Please see this comment https://github.com/elastic/opentelemetry-collector-components/pull/562#discussion_r2071880991, We dont want to expose the `project_id` etc details, instead keep this to attributes being derived from metadata_keys. Not sure if thats something possible with the metadata schema, but we can come back to later as well.

kaiyan-sheng: Ahh sorry I missed it. Will remove `project_id`. Yeah I dont think the current `mdatagen` supports dynamic attribute keys. Can look into it more later.
How about `processor_id` then? ProcessorID is from `telemetry.WithProcessorID(r.set.ID.String())`.

kaiyan-sheng:
(WIP) ToDo list in followup PR:
- [ ] Look into map type attribute for metadata_keys
- [ ] Add more telemetry metrics like [request duration ](https://github.com/elastic/opentelemetry-collector-components/pull/562#pullrequestreview-2812455687)and etc

processor/ratelimitprocessor/gubernator.go
lahsivjar: > Could you think of additional metrics that were useful in MIS so we can add them here? Thanks.

Currently in MIS we collect only 2 metrics from the rate limiter ([ref](https://docs.elastic.dev/mis/metrics-ingest-service#ratelimit)):

1. A metric to track the status of the ingest requests. This metric has dimensions with bounded cardinality to record the decision and the reason for the decision by the rate limiter service.
2. A metric to track the update requests sent between gubernator instances to update them. This metric has just one dimension of status.

While I like the idea of collecting request duration, however, I am not sure how useful concurrent requests would be. We should also be wary of adding too many metrics in hot path, as it does add some overhead.

kaiyan-sheng:
> Can we measure the no of requests that are over the limits at a given time? I believe we can by using the decision dimension, just confirm if thats is the case.

Yes we can use the `ratelimit_decision` attribute to filter `otelcol_ratelimit.requests` metric. 

> Request duration and concurrent requests

These two metrics are implemented in a separate [PR](https://github.com/elastic/opentelemetry-collector-components/pull/569).

gizas:
@kaiyan-sheng what is still pending on this PR? 
Have we decided to implement ratelimit.broadcasts based on the above https://github.com/elastic/opentelemetry-collector-components/pull/562#discussion_r2081464583?

kaiyan-sheng:
> @kaiyan-sheng what is still pending on this PR? Have we decided to implement ratelimit.broadcasts based on the above [#562 (comment)](https://github.com/elastic/opentelemetry-collector-components/pull/562#discussion_r2081464583)?

@gizas @vigneshshanmugam I probabaly don't understand `ratelimit.broadcasts` from APM fully. But I don't think this metric is applicable to our ratelimit processor. Please let me know if I'm wrong 🙂 I'm also trying to limit this PR's scope to only adding `otelcol_ratelimit.requests` metric and fix for CI to pass. I have a separate PR to add more metrics for ratelimiter.

processor/ratelimitprocessor/gubernator.go
```diff
@@ -126,6 +164,12 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 				zap.String("processor_id", r.set.ID.String()),
 				zap.Strings("metadata_keys", r.cfg.MetadataKeys),
 			)
+			r.requestTelemetry(ctx, []attribute.KeyValue{
+				telemetry.WithProcessorID(r.set.ID.String()),
+				telemetry.WithReason(telemetry.StatusOverLimit),
```
vigneshshanmugam, line 169: Isn't overlimit same as too many requests? Isnt throttled already covers that we are already over the limit? I feel we can simplify this and just remove the reason when we are throttled.

processor/ratelimitprocessor/local.go
```diff
@@ -59,20 +68,44 @@ func (r *localRateLimiter) RateLimit(ctx context.Context, hits int) error {
...
 		defer timer.Stop()
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 		case <-timer.C:
 		}
+		r.requestTelemetry(ctx, []attribute.KeyValue{
+			telemetry.WithReason(telemetry.StatusOverLimit),
+			telemetry.WithDecision("throttled"),
+			telemetry.WithThrottleBehavior("delay"),
```
vigneshshanmugam, line 96: I dont see the use of throttle behaviour and can be removed, WDYT?

processor/ratelimitprocessor/metadata.yaml
```diff
@@ -10,3 +10,28 @@ tests:
...
+      enabled: true
+      description: Number of rate-limiting requests
+      unit: "{requests}"
+      sum:
+        value_type: int
+        monotonic: true
+      attributes: ["processor_id", "ratelimit_decision", "limit_threshold", "reason"]
+
+attributes:
+  processor_id:
```
vigneshshanmugam, line 26: processor id can be removed.

vigneshshanmugam (commented):
Mostly LGTM

Makefile
```diff
@@ -65,6 +65,7 @@ gotidy:
 
 .PHONY: gogenerate
 gogenerate:
+	@command -v goimports >/dev/null 2>&1 || go install golang.org/x/tools/cmd/goimports@v0.33.0
```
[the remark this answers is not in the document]
kaiyan-sheng: OK good to know, I will move `goimports` into the `tools.go` file. But we still need to install it here though right? I made the change to use `install-tools` instead so it uses for the `tools.go` file. But then it installs all libraries listed in the file.

processor/ratelimitprocessor/gubernator.go
kaiyan-sheng: Yep totally will do. I only added it so all metrics have the same attributes. I don't think it's necessary!

processor/ratelimitprocessor/local.go
kaiyan-sheng: Agree! Also the over limit reason needs to be removed too.

gizas:
@kaiyan-sheng I think we need the `WithReason `  in https://github.com/elastic/opentelemetry-collector-components/pull/562/commits/fe67b2e08e12efd07a79206a0092737382e01289 for local.go
All the decision is throttled so how are we going to distinguish which case we have?

processor/ratelimitprocessor/gubernator_test.go
```diff
@@ -182,16 +196,205 @@ func TestGubernatorRateLimiter_RateLimit_MetadataKeys(t *testing.T) {
...
+
+				// Removed ignored dimensions from actualAttrs
+				filteredAttrs := make([]attribute.KeyValue, 0, len(actualAttrs))
+				for i := range actualAttrs {
+					if _, found := ignoredDimensions[string(actualAttrs[i].Key)]; !found {
+						filteredAttrs = append(filteredAttrs, actualAttrs[i])
+					}
+				}
+
+				// if !reflect.DeepEqual(expectedAttrs, filteredAttrs) {
```
gizas, line 385: @kaiyan-sheng I fixed the errors for test and lint.
In this test case I commented this part because this was creating the problem and now test passes. Why we needed this? I will let you delete or not this part in case missing sth

kaiyan-sheng:
> @kaiyan-sheng I think we need the `WithReason ` in [fe67b2e](https://github.com/elastic/opentelemetry-collector-components/commit/fe67b2e08e12efd07a79206a0092737382e01289) for local.go All the decision is throttled so how are we going to distinguish which case we have?

@gizas Oh good point! I forgot there are two diff reasons there for the local case. I will put them back!

processor/ratelimitprocessor/gubernator_test.go
kaiyan-sheng: I fixed the test in the latest commit, thanks!

processor/ratelimitprocessor/gubernator.go
constanca-m: > @vigneshshanmugam Could you explain a bit more here please? Sorry I'm lost 😅

I think I can explain this @kaiyan-sheng.

This rate limiter can be local (using a go library for the implementation) or from the gubernator. Basically, if you define a config from the gubernator, we expect it to be from there, otherwise it is local:

https://github.com/elastic/opentelemetry-collector-components/blob/a2d4b06050baadae4146c398d298b0ee90fcd7b2/processor/ratelimitprocessor/factory.go#L60-L64

So the rate limiter is later created this way:

https://github.com/elastic/opentelemetry-collector-components/blob/a2d4b06050baadae4146c398d298b0ee90fcd7b2/processor/ratelimitprocessor/processor.go#L67-L75

(I am using one signal just for example).

Since the telemetry builder for the local and gubernator is the same, I think @vigneshshanmugam is suggesting to only initiate it one time (so somewhere in the `processor.go`).

Then you can also check the functions:

https://github.com/elastic/opentelemetry-collector-components/blob/a2d4b06050baadae4146c398d298b0ee90fcd7b2/processor/ratelimitprocessor/processor.go#L146

Could the metric from the rate limiter be set here? Or does it really neeed to be in the `local.go` and `gubernator.go`? I don't know the answer to this now, I haven't checked the code in detail since the PR was first opened.


I hope it is clear. We can discuss this in more detail next week if you would like :)

processor/ratelimitprocessor/gubernator.go
```diff
@@ -102,21 +117,41 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	})
 	if err != nil {
 		r.set.Logger.Error("error executing gubernator rate limit request", zap.Error(err))
+		r.requestTelemetry(ctx, []attribute.KeyValue{
+			telemetry.WithReason(telemetry.RequestErr),
+			telemetry.WithDecision("accepted"),
+		})
```
constanca-m, line 123: So continuing my comment https://github.com/elastic/opentelemetry-collector-components/pull/562/files#r2098431295, maybe you could check `errors.Is(...,...)` and then set the metric in `processor.go`. This is so that `local.go` and `gubernator.go` won't need to have duplicated code.

processor/ratelimitprocessor/gubernator.go
kaiyan-sheng: @constanca-m Thanks for the comments! Yep make sense to move `telemetryBuilder, err := metadata.NewTelemetryBuilder(set.TelemetrySettings)` into `factory.go` so we don't create it separately for `local.go` and `gubernator.go`. But for metrics, I think it's still necessary to have in both places. I will keep them separate for now, move forward with this task and let's revisit it next week!

Makefile
```diff
@@ -65,6 +65,7 @@ gotidy:
 
 .PHONY: gogenerate
 gogenerate:
+	@command -v goimports >/dev/null 2>&1 || $(MAKE) -B install-tools
```
vigneshshanmugam, line 68: I dont have much context on the setup of this repo, I ll leave this to @lahsivjar or @carsonip who might be better to check this part.

processor/ratelimitprocessor/gubernator.go
vigneshshanmugam: ++, would be good to do as part of the followup.

processor/ratelimitprocessor/documentation.md
```diff
@@ -0,0 +1,23 @@
...
+
+| Unit | Metric Type | Value Type | Monotonic |
+| ---- | ----------- | ---------- | --------- |
+| {requests} | Sum | Int | true |
+
+#### Attributes
+
+| Name | Description | Values |
+| ---- | ----------- | ------ |
+| ratelimit_decision | rate limit decision | Any Str |
```
vigneshshanmugam, line 21:
```suggestion
| decision | rate limit decision | Any Str |
```

Feels implicit.

processor/ratelimitprocessor/internal/metadatatest/generated_telemetrytest.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+func NewSettings(tt *componenttest.Telemetry) processor.Settings {
+	set := processortest.NewNopSettings(processortest.NopType)
+	set.ID = component.NewID(component.MustNewType("ratelimit"))
+	set.TelemetrySettings = tt.NewTelemetrySettings()
+	return set
+}
+
+func AssertEqualRatelimitRequests(t *testing.T, tt *componenttest.Telemetry, dps []metricdata.DataPoint[int64], opts ...metricdatatest.Option) {
```
vigneshshanmugam, line 41: Cant we use this in testing instead of custom `assertMetrics` function?

processor/ratelimitprocessor/internal/telemetry/attributes.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"go.opentelemetry.io/otel/attribute"
+)
+
+// Reason represents a set of constant rate limit processing reasons to
+// include as attributes for telemetry.
+type Reason string
+
+const (
+	processorIDKey      = "processor_id"
```
vigneshshanmugam, line 29: nit: dead code.

processor/ratelimitprocessor/internal/telemetry/attributes.go
```diff
@@ -0,0 +1,61 @@
...
+	reasonKey           = "reason"
+	decisionKey         = "ratelimit_decision"
+	limitThresholdKey   = "limit_threshold"
+	throttleBehaviorKey = "throttle_behavior"
```
vigneshshanmugam, line 33: same as above.

processor/ratelimitprocessor/local.go
```diff
@@ -59,20 +64,44 @@ func (r *localRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	switch r.cfg.ThrottleBehavior {
 	case ThrottleBehaviorError:
 		if ok := limiter.AllowN(time.Now(), hits); !ok {
+			r.requestTelemetry(ctx, []attribute.KeyValue{
+				telemetry.WithReason(telemetry.TooManyRequests),
+				telemetry.WithDecision("throttled"),
+				telemetry.WithThrottleBehavior("error"),
```
vigneshshanmugam, line 70: I believe we agreed to remove these extra dimensions as they are conveyed by the decision already? Can we remove this. Also the over limit/under limits.

vigneshshanmugam (commented):
Few nits, mostly there.

processor/ratelimitprocessor/local.go
kaiyan-sheng: Yeah I added them back because if we want to tell the difference between throttled with delay or error, there won't be a way without these fields. Is that the level of details that does not matter?

kaiyan-sheng: Conclusion: 

- we accept when we can't reach gubernator
- If user is over the limit, we reject with throttled decision
- Under limits, we respond with accepted

Both cases we have the calculated threshold, delay or error does not matter here. Let's keep it simple for now and if we find useful later, we can add it then.

processor/ratelimitprocessor/internal/metadatatest/generated_telemetrytest.go
kaiyan-sheng: Yes I believe so! I will look into this tomorrow!

Makefile
carsonip: I've briefly looked at it. IMO it is an upstream mdatagen bug causing missing import line `"github.com/elastic/opentelemetry-collector-components/processor/ratelimitprocessor/internal/metadata"` in `processor/ratelimitprocessor/internal/metadatatest/generated_telemetrytest_test.go`. You may submit an issue upstream mdatagen. To avoid blocking your work, add a comment next to your goimports workaround and reference the upstream bug report. That'll be good enough

kaiyan-sheng: Done, thanks for looking into it!

PR #583, opened 2025-05-29
Add ratelimiter per-project settings override

zmoog:
WIP: draft a per-project ratelimiter settings override mechanism.

### DIFF
--- a/processor/ratelimitprocessor/config.go
+++ b/processor/ratelimitprocessor/config.go
@@ -34,20 +34,8 @@ const (
 	msgEmptyField = "%s field value is empty"
 )
 
-// Config holds configuration for the ratelimit processor.
-type Config struct {
-	// Gubernator holds configuration for Gubernator,
-	// to control distributed rate limiting.
-	//
-	// If Gubernator is nil, then rate limiting is performed
-	// locally to the collector.
-	Gubernator *GubernatorConfig `mapstructure:"gubernator"`
-
-	// MetadataKeys holds a list of client metadata keys for
-	// defining the rate limiting key, in addition to the
-	// processor ID.
-	MetadataKeys []string `mapstructure:"metadata_keys"`
-
+// RateLimitSettings holds the core rate limiting configuration.
+type RateLimitSettings struct {
 	// Strategy holds the rate limiting strategy.
 	//
 	// Defaults to "requests".
@@ -63,6 +51,29 @@ type Config struct {
 	//
 	// Defaults to "error"
 	ThrottleBehavior ThrottleBehavior `mapstructure:"throttle_behavior"`
+}
+
+// Config holds configuration for the ratelimit processor.
+type Config struct {
+	// Gubernator holds configuration for Gubernator,
+	// to control distributed rate limiting.
+	//
+	// If Gubernator is nil, then rate limiting is performed
+	// locally to the collector.
+	Gubernator *GubernatorConfig `mapstructure:"gubernator"`
+
+	// MetadataKeys holds a list of client metadata keys for
+	// defining the rate limiting key, in addition to the
+	// processor ID.
+	MetadataKeys []string `mapstructure:"metadata_keys"`
+
+	// Embed the rate limit settings
+	RateLimitSettings `mapstructure:",squash"`
+
+	// Overrides holds a list of overrides for the rate limiter.
+	//
+	// Defaults to empty
+	Overrides map[string]RateLimitSettings `mapstructure:"overrides"`
 }
 
 // Strategy identifies the rate-limiting strategy: requests, records, or bytes.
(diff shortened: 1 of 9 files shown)


zmoog:
@simitt, I'd like to double-check with you whether this is the override semantics we're looking for. 

Note: I also discussed the switch from a request-based to a bytes-based [^1] rate limiting with @vigneshshanmugam. So, in the following examples, I will use the `bytes` strategy to reflect this.

Let's start with a few examples.

(1) Baseline rate-limiting

In this example, the rate limiter applies the same baseline limits to all projects.

```yaml
ratelimit:
  metadata_keys:
   - x-elastic-project-id
  rate: 2046820
  burst: 5117050 # In these examples, I'm considering a 2.5x burst ratio 
  strategy: bytes
  throttle_behavior: error  
```


(2) Per-project rate override

In this example, we increase the rate by 50% over the baseline.

```yaml
ratelimit:
  metadata_keys:
   - x-elastic-project-id
  rate: 2046820
  burst: 5117050 # In these examples, I'm considering a 2.5x burst ratio 
  strategy: bytes
  throttle_behavior: error
  overrides:  
    x-elastic-project-id:local:
      rate: 3070230 # +50% over baseline
```


(3) Per-project rate and burst override

In this example, we increase the rate and burst by 50% over the baseline.

```yaml
ratelimit:
  metadata_keys:
   - x-elastic-project-id
  rate: 2046820
  burst: 5117050 # In these examples, I'm considering a 2.5x burst ratio 
  strategy: bytes
  throttle_behavior: error
  overrides:  
    x-elastic-project-id:local:
      rate: 3070230 # +50% over baseline
    x-elastic-project-id:whatever:
      rate: 3070230 # +50% over baseline
      burst: 7675575 # +50% over baseline
```

[^1]: I noticed the [comment](https://github.com/elastic/opentelemetry-collector-components/blob/5b8392b1b61a1d01b43ac1eadcb414ab6c921da3/processor/ratelimitprocessor/config.go#L84-L87) in the `bytes` strategy that highlights that the using this strategy is "much more expensive" than that the `requests` and `records` strategies. We should consider this aspect, and run some tests to measure this increase — if we don't already have some.

zmoog:
> General approach looks good to me. What would be worth adding is a time interval indicator, e.g. `rate_per_second` if the rate is applied on a per second interval. Gubernator also allows to configure the throttling interval. I don't think this is currently set anywhere, and is not the most urgent, but would be a nice add-on to also allow for it. See an example for this in https://github.com/elastic/serverless-gitops/blob/main/services/apm-ingest-service/values/production/default.yaml.

Adding a `throttle_interval` seems a low hanging fruit. Adding it!

constanca-m (commented):
Thank you @zmoog ! It looks great, I couldn't find issues with your approach.

processor/ratelimitprocessor/config.go
```diff
@@ -102,20 +134,83 @@ type GubernatorBehavior string
...
+		},
 	}
 }
 
-func (config *Config) Validate() error {
+// resolveRateLimitSettings returns the rate limit settings for the given unique key.
+// If no override is found, the default rate limit settings are returned.
+func resolveRateLimitSettings(cfg *Config, uniqueKey string) RateLimitSettings {
+	result := cfg.RateLimitSettings
+	if len(cfg.Overrides) > 0 {
```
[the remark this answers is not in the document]
zmoog: Not needed. This optimization isn't worth sacrificing readability for. I'm removing it.

processor/ratelimitprocessor/config.go
```diff
@@ -102,19 +134,82 @@ type GubernatorBehavior string
...
+		},
+	}
+}
+
+// resolveRateLimitSettings returns the rate limit settings for the given unique key.
+// If no override is found, the default rate limit settings are returned.
+func resolveRateLimitSettings(cfg *Config, uniqueKey string) RateLimitSettings {
+	// We start from the default settings
+	result := cfg.RateLimitSettings
+	if override, ok := cfg.Overrides[uniqueKey]; ok {
```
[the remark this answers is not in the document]
zmoog: I included the header because the typical complete settings for the rate limiter are something like this:

```yaml
ratelimit:
  metadata_keys:
   - x-elastic-project-id
  rate: 1000
  burst: 10000
  type: gubernator
  strategy: bytes
  throttle_behavior: error 
  throttle_interval: 10s
  overrides:  
    x-elastic-project-id:local:
      rate: 6140460
```

Users can set one or more entries for the `metadata_keys` they want to use for the rate limiting.

Currently, the current override implementation looks for an exact match of all the keys in the `metadata_keys`.

If we're using only one metadata keys like `x-elastic-project-id`, then the unique key value is `x-elastic-project-id:tenant-id`.

If we're using two or more metadata keys, then the unique key value is `x-elastic-project-id:tenant-id;source-ip:127.0.0.1`.

Would it be useful to add support for partial matches, such as the following example?

```yaml
ratelimit:
  metadata_keys:
   - x-elastic-project-id
   - source-ip # just an example, not sure it exists
  rate: 1000
  burst: 10000
  type: gubernator
  strategy: bytes
  throttle_behavior: error 
  throttle_interval: 10s
  overrides:  
    x-elastic-project-id:local:
      rate: 6140460
    source-ip:127.0.0.1:
      rate: 6140460
```

processor/ratelimitprocessor/README.md
```diff
@@ -42,4 +54,22 @@ processors:
...
\ No newline at end of file
+You can configure the Gubernator using `GUBER_*` [environment variables](https://github.com/gubernator-io/gubernator/blob/master/example.conf).
+
+Example when using as a distributed rate limiter (Gubernator) with overrides:
+
+```yaml
+processors:
+  ratelimiter:
+    metadata_keys:
+    - x-elastic-project-id
```
vigneshshanmugam, line 81: To keep it generic, lets use x-tenant-id or similar.

processor/ratelimitprocessor/README.md
```diff
@@ -14,7 +14,19 @@ a in-memory rate limiter, or makes use of a [gubernator](https://github.com/gube
...
 | `throttle_behavior` | Processor behavior for when the rate limit is exceeded. Options are `error`, return an error immediately on throttle and does not send the event, and `delay`, delay the sending until it is no longer throttled. | Yes      | `error`    |
+| `throttle_interval` | Time interval for throttling. Has no effect when `type` is `gubernator`.                                                                                                                                          | No       | `1s`       |
 | `type`              | Type of rate limiter. Options are `local` or `gubernator`.                                                                                                                                                        | No       | `local`    |
+| `overrides`         | Map of metadata key overrides for the rate limiter. See the possible overrides and examples below.                                                                                                                | No       |            |
+
+### Possible overrides:
```
vigneshshanmugam: nit: Overrides based on Metadata keys?

zmoog: Oh, sounds much better, thanks!

processor/ratelimitprocessor/README.md
```diff
@@ -14,15 +14,29 @@ a in-memory rate limiter, or makes use of a [gubernator](https://github.com/gube
...
+### Examples
 
 Example when using as a local rate limiter:
 
 ```yaml
 processors:
   ratelimiter:
     metadata_keys:
-    - x-elastic-project-id
+    - tenant-id
```
simitt, line 39: Can you expand on why the example is changed to use `tenant-id`? Do you expect this to be the primary the use case rather than a per project throttling? (I assumed we would mostly consider rate limit overrides per project).

processor/ratelimitprocessor/config.go
```diff
@@ -102,19 +134,82 @@ type GubernatorBehavior string
...
 func createDefaultConfig() component.Config {
 	return &Config{
-		Strategy:         StrategyRateLimitRequests,
-		ThrottleBehavior: ThrottleBehaviorError,
-		Type:             LocalRateLimiter,
+		Type: LocalRateLimiter,
+		RateLimitSettings: RateLimitSettings{
+			Strategy:         StrategyRateLimitRequests,
+			ThrottleBehavior: ThrottleBehaviorError,
+			ThrottleInterval: 1 * time.Second,
```
gizas, line 141: nit: Define it as constant as in here https://github.com/elastic/hosted-otel-collector/blob/accbdc7cacb29450ca61e0c052a4da5bbf624ca2/internal/extension/elasticcloudauthextension/config.go#L46

Just to have same style

simitt (approved):
LGTM!

As a follow up action item, it would be good to add a dedicated [runbook](https://docs.elastic.dev/hotel/runbooks/runbooks) describing how to set a per-project override.

PR #594, opened 2025-06-03
docs: add elasticapm receiver tls and apikey configs

rogercoll:
Adds a README which includes some sample configurations and links to enable TLS and API_KEY authentication in the elasticapm receiver.


ChrsMark (approved):
LGTM

One note here: reading these docs I wonder what `agent_config` part is and how ti should be used. I need to go into the code implementation to figure this out, right? Maybe it'd worth adding some docs about this option too.

rogercoll:
> One note here: reading these docs I wonder what `agent_config` part is and how ti should be used. I need to go into the code implementation to figure this out, right? Maybe it'd worth adding some docs about this option too.

 Totally right, the `agents_config` part is for retrieving the remote configurations, not for the data ingestion. I created this issue to document all receiver's functionality: https://github.com/elastic/opentelemetry-collector-components/issues/599

Issue #595, opened 2025-06-03
[extension/apmconfig] Enable TLS by using OpenTelemetry confighttp package


rogercoll:
The extension opens an HTTP port for the OpAMP endpoint, but it does not provide the ability to encrypt the connection. Instead of exporting all the available TLS configs and creating the corresponding server, we should leverage the [confighttp](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/README.md#http-configuration-settings) package and delegate the server creation to it. The same approach is used by the OTLP and the Elasticapm receiver.

PR #596, opened 2025-06-03
feat: add confighttp configuration with TLS support

rogercoll:
Fixes: https://github.com/elastic/opentelemetry-collector-components/issues/595

Breaking change in all configurations, from:

```yaml
apmconfig:
  opamp:
    server:
       endpoint: "http://..."
```

The new configuration uses a similar approach as the OTLP receiver, where multiple protocols can run simultaneously. Having in mind that Websockets protocol is supported by OpAMP and might be implemented in the future:
```yaml
apmconfig:
  opamp:
    protocols:
      http:
        endpoint: "http://..."
```

In addition, the [confighttp](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/README.md#http-configuration-settings) configuration is used to create the server, thus leveraging all its features (TLS, authentication, CORS, timeout, etc).

### DIFF
--- a/extension/apmconfigextension/README.md
+++ b/extension/apmconfigextension/README.md
@@ -89,7 +89,7 @@ extensions:
 
 ## Advanced configuration
 
-The apmconfig extension embeds the [confighttp.ServerConfig](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/README.md), which means it supports standard HTTP server configuration, including TLS/mTLS and authentication.
+The apmconfig extension embeds the [confighttp.ServerConfig](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.125.0/config/confighttp/README.md), which means it supports standard HTTP server configuration, including TLS/mTLS and authentication.
 
 ### TLS and mTLS settings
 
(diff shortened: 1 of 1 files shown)


trentm:
> I would like to rename the `agent_config` section as well.

(Take my feedback with a grain of salt. I have almost no experience with naming OTel Collector-y things.)

Do these config settings need to be namespaced under `agent_config` or `fetcher` at all? If this fetching is a fundamental function of this extension, then could it be somethign like:

```
extensions:
  apmconfig:
    cache_duration: 10s
    elasticsearch:
      endpoint: http://localhost:9200
      auth:
        authenticator: bearertokenauth
    opamp:
      server:
        endpoint: "localhost:4320"
```

?

---

On the subject of naming: Is there a reason this extension is named `apmconfig` and not `elasticapmconfig`?  I gather it doesn't really matter, because the component isn't going to be upstreamed every, I would imagine.

rogercoll:
> Do these config settings need to be namespaced under agent_config or fetcher at all?

It's not strictly necessary, but in my opinion, it provides a useful way to distinguish the two core responsibilities of the extension:

- Fetcher: from when to retrieve the remote configuration
- Server: how to communicate with agents

>  Is there a reason this extension is named apmconfig and not elasticapmconfig? I gather it doesn't really matter, because the component isn't going to be upstreamed every, I would imagine.

I actually believe it could be a candidate for upstreaming. The OpAMP server is designed to be agnostic with respect to the remote configuration fetcher/provider. It relies on a client interface that can be implemented by various fetchers—Elasticsearch being one example (see the Otel parameters here: https://github.com/elastic/opentelemetry-collector-components/blob/main/extension/apmconfigextension/apmconfig/client.go#L39) Indeed, since agent_config is a map, it could be extended to support other providers or fetchers in the future:

```yaml
extensions:
  apmconfig:
    agent_config:
      kibana:
        endpoint: http://localhost:9200
```

Or another vendor:


```yaml
extensions:
  apmconfig:
    agent_config:
      other_vendor:
        url: http://localhost:9200
```

extension/apmconfigextension/README.md
```diff
@@ -64,24 +64,80 @@ unique remote configuration.
...
+    opamp:
+      protocols:
+        http:
+          endpoint: ":4320"
+```
+
+
+## Advanced configuration
+
+The apmconfig extension embeds the [confighttp.ServerConfig](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/README.md), which means it supports standard HTTP server configuration, including TLS/mTLS and authentication.
```
ChrsMark, line 92: Can we pin those to specific versions instead of `main`?

ChrsMark (approved):
LGTM

extension/apmconfigextension/README.md
rogercoll: Definitely, https://github.com/elastic/opentelemetry-collector-components/pull/596/commits/2e623d3ff9697a35dddef3b43ca3a322d5f8d28b

PR #601, opened 2025-06-06
Add cumulativetodelta processor

anuraaga:
I noticed EDOT collector doesn't have `cumulativetodelta` processor. It seems useful to include it because currently APM only supports delta metrics FWIU. For example, if scraping prometheus endpoints with the collector, this processor can be used to get them into APM.

I also fixed cross compile on MacOS - `builder` was also being built with the provided `GOOS` causing it to be not executable.

/cc @codefromthecrypt

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ remove-toolchain:
 # Build a collector based on the Elastic components (generate Elastic collector)
 .PHONY: genelasticcol
 genelasticcol: $(BUILDER)
-	$(BUILDER) --config ./distributions/elastic-components/manifest.yaml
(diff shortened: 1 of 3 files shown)


axw (approved):

codefromthecrypt:
This is important for two reasons:

1. normal otel SDKs use cumulative by default, and even if EDOT sdks use delta, some may struggle to override the ENV just to use elastic dist of collector.
2. prometheus histograms are cumulative and there seems no way to change this, so the prometheus receiver has limited value until we also add this processor. this affects envoy for example

so this processor add allows a tactical way to migrate a brown field setup, or those without a path for delta at all.

PR #609, opened 2025-06-10
[extension/apmconfig] save RemoteConfigStatus.LastRemoteConfigHash regardless of status

trentm:
Before this change the OpAMP server would only store the LastRemoteConfigHash
provided in an AgentToServer message if the RemoteConfigStatus.Status
was "APPLIED".

An agent may want to (a) acknowledge that a particular config was
received (via LastRemoteConfigHash) *and* (b) provide some other status
(APPLYING or FAILED). In my particular case, for EDOT Node.js, I want to
be able to support the case where the server might be sending new config
names that an older EDOT Node.js might not know about. My
RemoteConfigStatus in this case is:

    LastRemoteConfigHash: ...
    Status=FAILED
    ErrorMessage="there were issues applying remote config: config name 'some_new_var' is unsupported"

The downside of the current behaviour is that the server will ignore the
provided LastRemoteConfigHash and continue to re-send the same remote
config in every ServerToAgent. Re-sending is not going to fix the issue
the agent had with applying the config.

My interpretation of the OpAMP spec and proto is that the OpAMP server
should use *only* the LastRemoteConfigHash field for determining if
RemoteConfig should be sent. The proto says:

    // The hash of the remote config that was last received by this Agent in the
    // AgentRemoteConfig.config_hash field.
    // The Server SHOULD compare this hash with the config hash
    // it has for the Agent and if the hashes are different the Server MUST include
    // the remote_config field in the response in the ServerToAgent message.


trentm:
Oops, I have test failures to deal with.

trentm:
Another argument in favour of the behaviour I'm proposing is that this is what the internal example OpAMP server from opamp-go.git does:
https://github.com/open-telemetry/opamp-go/blob/b9a8e1cdee7cf33db29607ffc049521e2d6a54d3/internal/examples/server/data/agent.go#L202

trentm:
It looks like this could impact EDOT Java: its code *can* send a "FAILED" RemoteConfigStatus.Status.  /cc @jackshirazi

extension/apmconfigextension/opamp_callbacks.go
```diff
@@ -132,11 +132,14 @@ func (rc *remoteConfigCallbacks) onMessage(ctx context.Context, conn types.Conne
 		return &serverToAgent
 	}
 
-	if message.GetRemoteConfigStatus() != nil && message.GetRemoteConfigStatus().Status == protobufs.RemoteConfigStatuses_RemoteConfigStatuses_APPLIED && bytes.Equal(remoteConfig.ConfigHash, message.RemoteConfigStatus.GetLastRemoteConfigHash()) {
-		rc.logger.Info("Remote config applied", zap.String("hash", hex.EncodeToString(remoteConfig.ConfigHash)), agentUidField)
-		agent.lastConfigHash = message.GetRemoteConfigStatus().GetLastRemoteConfigHash()
+	remoteConfigStatus := message.GetRemoteConfigStatus()
```
xrmx, line 135: While at it maybe add a comment on our updated policy for updating the lastConfigHash?

extension/apmconfigextension/opamp_callbacks_test.go
```diff
@@ -177,17 +177,6 @@ func TestOnMessage(t *testing.T) {
 					expectedServerToAgent: &protobufs.ServerToAgent{
 						InstanceUid:  []byte("test"),
 						Capabilities: uint64(protobufs.ServerCapabilities_ServerCapabilities_OffersRemoteConfig),
-						RemoteConfig: &protobufs.AgentRemoteConfig{
```
xrmx, line 180: Should we also add a test case with the failure case you shared?

extension/apmconfigextension/opamp_callbacks.go
trentm: @xrmx I'm missing context. What policy do you mean?

extension/apmconfigextension/opamp_callbacks.go
```diff
@@ -132,11 +132,14 @@ func (rc *remoteConfigCallbacks) onMessage(ctx context.Context, conn types.Conne
...
+	if remoteConfigStatus != nil {
```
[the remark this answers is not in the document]
trentm: This if-block is about having the OpAMP server store/remember the `LastRemoteConfigHash` from the agent.

It is the `if !bytes.Equal(agent.lastConfigHash, remoteConfig.ConfigHash) {` block below that decides whether the server should send remote config in the ServerToAgent response.

---

However, I see now that given the possible early-return on line 132, this block should be moved up higher. Basically I'm trying to change this `onMessage` function to:
- first, save any agent data from `AgentToServer` that the server should remember,
- *then*, decide what data should be added to `ServerToAgent`.

extension/apmconfigextension/opamp_callbacks.go
```diff
@@ -132,11 +132,14 @@ func (rc *remoteConfigCallbacks) onMessage(ctx context.Context, conn types.Conne
...
+		agent.lastConfigHash = remoteConfigStatus.GetLastRemoteConfigHash()
```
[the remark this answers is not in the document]
trentm: I may be misunderstanding what you are asking about. 

My understanding is that **not every AgentToServer message needs to include the `RemoteConfigStatus`**. After an agent has sent its `RemoteConfigStatus`, it doesn't need to include it again unless it changes (https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agent-status-compression).  So, the server needs to store that `RemoteConfigStatus.LastRemoteConfigHash` whenever it receives it.

extension/apmconfigextension/opamp_callbacks_test.go
trentm: I could do so, yes. Though this "APPLYING" test case happens to test a similar case. Adding one for `RemoteConfigStatuses_FAILED` is a good idea.

trentm: commit c277b97

extension/apmconfigextension/opamp_callbacks.go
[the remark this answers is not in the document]
trentm: > However, I see now that given the possible early-return on line 132, this block should be moved up higher.

commit f9e8b25830fc4e5a73d7b77a4f4562328af12bc0

extension/apmconfigextension/opamp_callbacks.go
xrmx: To resend the config only if the hash changed and not retry if the client failed to apply it

extension/apmconfigextension/opamp_callbacks.go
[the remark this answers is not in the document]
rogercoll: The proposal of moving up the if-block looks cleaner/simpler, thanks!

rogercoll (approved):

jackshirazi:
> It looks like this could impact EDOT Java: its code _can_ send a "FAILED" RemoteConfigStatus.Status. /cc @jackshirazi

Not specifically relevant but for anyone reading this (in case of searching/tracking after some maintenance change) regarding the Java agent, it has the capability to send FAILED, but can't actually send it for the current implementation (1.4.2 snapshot)

extension/apmconfigextension/opamp_callbacks.go
rogercoll: Fixed in https://github.com/elastic/opentelemetry-collector-components/pull/609/commits/c277b97fd60663a6e5d009868c5efdd0395d48fe

Merging as discussed internally

PR #610, opened 2025-06-11
feat: return empty config if hash equals remote config etag

rogercoll:


extension/apmconfigextension/opamp_callbacks.go
```diff
@@ -134,9 +128,9 @@ func (rc *remoteConfigCallbacks) onMessage(ctx context.Context, conn types.Conne
 
 	if message.GetRemoteConfigStatus() != nil && message.GetRemoteConfigStatus().Status == protobufs.RemoteConfigStatuses_RemoteConfigStatuses_APPLIED && bytes.Equal(remoteConfig.ConfigHash, message.RemoteConfigStatus.GetLastRemoteConfigHash()) {
 		rc.logger.Info("Remote config applied", zap.String("hash", hex.EncodeToString(remoteConfig.ConfigHash)), agentUidField)
-		agent.lastConfigHash = message.GetRemoteConfigStatus().GetLastRemoteConfigHash()
+		agent.LastConfigHash = message.GetRemoteConfigStatus().GetLastRemoteConfigHash()
 		rc.agentState.Store(agentUid, agent)
-	} else if !bytes.Equal(agent.lastConfigHash, remoteConfig.ConfigHash) {
+	} else if !bytes.Equal(agent.LastConfigHash, remoteConfig.ConfigHash) {
```
trentm, line 133: This bytes.Equal check could be dropped now, then right? This same check is now done in `fetcherAPMWatcher.RemoteConfig()`.

extension/apmconfigextension/opamp_callbacks.go
```diff
@@ -40,12 +40,6 @@ type remoteConfigCallbacks struct {
 	logger     *zap.Logger
 }
 
-type agentInfo struct {
-	agentUid              apmconfig.InstanceUid
-	identifyingAttributes apmconfig.IdentifyingAttributes
-	lastConfigHash        []byte
-}
```
trentm, line 47: (non blocking) Take this with a grain of salt, as I'm only casually in this repo.  I think `agentInfo` for the OpAMP server's cache of info about was/is more accurate about what it is doing, or could eventually be doing.

- This could eventually store the last time there was a heartbeat from a given agent `InstanceUid` to support having a set of "active" or "recent" agents. This could eventually store the [EffectiveConfig](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoservereffective_config) or the `status` and `error_message` fields from [RemoteConfigStatus](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#remoteconfigstatus-message) for use in the "Agent Explorer" table in Kibana:

Screenshot 2025-06-11 at 1 10 27 PM

- The `Query` struct being passed to `RemoteConfig()` doesn't actually use the `agentUid` field, so the `RemoteConfig()` signature *could* be `RemoteConfig(context.Context, IdentifyingAttributes, LastConfigHash)` -- i.e. still just 3-args in case that was your trigger to change to using a `Query` struct.

trentm:
The govulncheck failure should be fixed by https://github.com/elastic/opentelemetry-collector-components/pull/613

extension/apmconfigextension/opamp_callbacks.go
rogercoll: Good point, the idea of the `Query` type was to provide a more Elastic agnostic interface for remote config providers. But given that probably there won't be another implementation near time soon, I will roll back to use the specific signature: `RemoteConfig(context.Context, IdentifyingAttributes, LastConfigHash)`

> This could eventually store the EffectiveConfig or the status and error_message fields from RemoteConfigStatus for use in the "Agent Explorer" table in Kibana
sounds like a possible use case to me, switching back to `agentInfo` structure for caching agent's info

rogercoll: changed in https://github.com/elastic/opentelemetry-collector-components/pull/610/commits/d0f9068f076c8d169bc1e90af924de77004b8360

Issue #465, opened 2025-03-18
[connector/elasticapm] Configure overflow limits


lahsivjar:
Overflow limits should be configured to create an upper bound for the aggregation resource usage.

lahsivjar:
@isaacaflores2 You are on the right track. For the values, let’s take the limits from the apm-aggregation setup available for MIS.

isaacaflores2:
@lahsivjar @vigneshshanmugam Once these changes are merged, should we define the limits in the `motel-aggregation-collector`?

Since the aggregation collector is multi-tenant I guess these limits would apply to multiple projects 🤔 So maybe it would be worth testing with the current aggregation collector resources configuration to determine a new set of overflow limits?

isaacaflores2:
> Do you mean the MIS limits? (I believe in APM we increase them with increasing deployment size).

Yes currently the default values in the connector are setup to match the APM default values. I think we can configure the motel aggregation collector to increase the limits based on the deployment size. This can be covered by: https://github.com/elastic/hosted-otel-collector/issues/1110

PR #626, opened 2025-06-18
[connector/elasticapm] Add ability configure metric aggregation limit configs

isaacaflores2:
Exposed aggregation limit configs that are propagated to the `lsmintervalprocessor`. 

Note: I opted to keep the configuration names ( `resource_limit` ) the same as the `lsmintervalprocessor` mostly so they are consistent for both components. Another option would be to restructure everything to omit the `_limit` suffix so `elasticapm::aggregation::limit::resource` and `elasticapm::aggregation::limit::scope`. Let me know what you think. 

## Checklist
- [x] Update `connector/elasticapmconnector/README.md` to includes the limits in the example config.yml
- [x] Add test cases to assert overflow behavior
- [x] Run collector locally to verify overflow behavior
- [x] Double check the default values are appropriate and match MIS
   - I found the default values under `apm-managed-service/checkout/apm-aggregation-service/config/config.go`
   ```go
                         // Defaults to apm-server 8GB equivalent limits.
			MaxSpanGroups:                         40000,
			MaxSpanGroupsPerService:               4000,
			MaxTransactionGroups:                  40000,
			MaxTransactionGroupsPerService:        4000,
			MaxServiceTransactionGroups:           8000,
			MaxServiceTransactionGroupsPerService: 800,
			MaxServices:                           8000,
   ``` 
   
 ## Local Testing
 1. Build collector `make genelasticcol` 
 2. Run the collector with the below config
 ```yaml
 receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
        include_metadata: true
      http:
        endpoint: 0.0.0.0:4318
        include_metadata: true

exporters:
  otlp:
    endpoint: "${env:ELASTIC_APM_SERVER_URL:-localhost:8200}"
    headers:
      Authorization: "Bearer ${env:ELASTIC_APM_TOKEN}"
    sending_queue:
      enabled: false
    timeout: 60s
  debug:

connectors:
  elasticapm:
    aggregation:
      limit:
        resource:
          max_cardinality: 1
        scope:
          max_cardinality: 1
        metric:
          max_cardinality: 1
        datapoint:
          max_cardinality: 1

service:
  pipelines:
    logs:
      receivers: [otlp]
      exporters: [elasticapm, otlp]
    metrics:
      receivers: [otlp]
      exporters: [elasticapm, otlp]
    traces:
      receivers: [otlp]
      exporters: [elasticapm, otlp]
    metrics/apm:
      receivers: [elasticapm]
      exporters: [otlp, debug]

 ```
 3. Send data with different service names (`local_test_1`, `local_test_2`)
  -  `telemetrygen traces --duration 1s --otlp-insecure  --service "local_test_1" `
  4. Search for overflow metric documents with query `labels.overflow:true` to validate overflow behavior

      <details>
      <summary>Example Overflow metric</summary>
      
      ```json
         {
          "_index": ".ds-metrics-service_summary.1m.otel-default-2025.07.31-000001",
          "_id": "RO44WNIpbtQmrnRmAAABmGKSIzM",
          "_version": 1,
          "_source": {
            "@timestamp": "2025-07-31T22:20:01.203Z",
            "_metric_names_hash": "fd91fda6720b28bc",
            "attributes": {
              "metricset.interval": "1m",
              "metricset.name": "service_summary",
              "processor.event": "metric"
            },
            "data_stream": {
              "dataset": "service_summary.1m.otel",
              "namespace": "default",
              "type": "metrics"
            },
            "metrics": {
              "service_summary": 1
            },
            "resource": {
              "attributes": {
                "agent.name": "unknown",
                "service.name": "_other",
                "signaltometrics.service.instance.id": "11c21f6a-98d8-44a6-ab32-4574bca301c1",
                "signaltometrics.service.name": "motel-aggregation-collector"
              }
            },
            "scope": {
              "name": "github.com/open-telemetry/opentelemetry-collector-contrib/connector/signaltometricsconnector"
            }
          },
          "fields": {
            "signaltometrics.service.instance.id": [
              "11c21f6a-98d8-44a6-ab32-4574bca301c1"
            ],
            "scope.name": [
              "github.com/open-telemetry/opentelemetry-collector-contrib/connector/signaltometricsconnector"
            ],
            "service.name": [
              "_other"
            ],
            "data_stream.namespace": [
              "default"
            ],
            "resource.attributes.agent.name": [
              "unknown"
            ],
            "_metric_names_hash": [
              "fd91fda6720b28bc"
            ],
            "service_summary": [
              1
            ],
            "metricset.interval": [
              "1m"
            ],
            "data_stream.type": [
              "metrics"
            ],
            "resource.attributes.signaltometrics.service.name": [
              "motel-aggregation-collector"
            ],
            "metricset.name": [
              "service_summary"
            ],
            "signaltometrics.service.name": [
              "motel-aggregation-collector"
            ],
            "attributes.processor.event": [
              "metric"
            ],
            "@timestamp": [
              "2025-07-31T22:20:01.203Z"
            ],
            "data_stream.dataset": [
              "service_summary.1m.otel"
            ],
            "processor.event": [
              "metric"
            ],
            "metrics.service_summary": [
              1
            ],
            "agent.name": [
              "unknown"
            ],
            "attributes.metricset.interval": [
              "1m"
            ],
            "resource.attributes.signaltometrics.service.instance.id": [
              "11c21f6a-98d8-44a6-ab32-4574bca301c1"
            ],
            "attributes.metricset.name": [
              "service_summary"
            ],
            "event.dataset": [
              "service_summary.1m.otel"
            ],
            "resource.attributes.service.name": [
              "_other"
            ]
          }
        }
      ```
      </details>

#### service map overflow
service_map_other

## Reference
Closes : https://github.com/elastic/opentelemetry-collector-components/issues/465

### DIFF
--- a/connector/elasticapmconnector/config.go
+++ b/connector/elasticapmconnector/config.go
@@ -122,25 +122,34 @@ func (cfg Config) lsmConfig() *lsmconfig.Config {
 			MaxCardinality: cfg.Aggregation.Limit.ResourceLimit.MaxCardinality,
 			Overflow: lsmconfig.OverflowConfig{
 				Attributes: []lsmconfig.Attribute{
-					{
-						// Specific attribute required for APU UI compatibility
-						Key:   "service.name",
-						Value: "_other",
-					},
+					{Key: "service.name", Value: "_other"}, // Specific attribute required for APU UI compatibility
+					{Key: "overflow", Value: "resource"},
 				},
 			},
 		}
 		lsmConfig.ScopeLimit = lsmconfig.LimitConfig{
 			MaxCardinality: cfg.Aggregation.Limit.ScopeLimit.MaxCardinality,
-			Overflow:       lsmconfig.OverflowConfig{},
+			Overflow: lsmconfig.OverflowConfig{
+				Attributes: []lsmconfig.Attribute{
+					{Key: "overflow", Value: "scope"},
+				},
+			},
 		}
 		lsmConfig.MetricLimit = lsmconfig.LimitConfig{
 			MaxCardinality: cfg.Aggregation.Limit.MetricLimit.MaxCardinality,
-			Overflow:       lsmconfig.OverflowConfig{},
+			Overflow: lsmconfig.OverflowConfig{
+				Attributes: []lsmconfig.Attribute{
+					{Key: "overflow", Value: "metric"},
+				},
+			},
 		}
 		lsmConfig.DatapointLimit = lsmconfig.LimitConfig{
 			MaxCardinality: cfg.Aggregation.Limit.DatapointLimit.MaxCardinality,
-			Overflow:       lsmconfig.OverflowConfig{},
+			Overflow: lsmconfig.OverflowConfig{
+				Attributes: []lsmconfig.Attribute{
+					{Key: "overflow", Value: "datapoint"},
+				},
+			},
 		}
 	}
 	return lsmConfig
--- a/connector/elasticapmconnector/testdata/logs/service_summary_overflow/aggregated_metrics.yaml
+++ b/connector/elasticapmconnector/testdata/logs/service_summary_overflow/aggregated_metrics.yaml
@@ -76,6 +76,9 @@ resourceMetrics:
         - key: service.name
           value:
             stringValue: "_other"
+        - key: overflow
+          value:
+            stringValue: resource
     scopeMetrics:
       - scope:
           name: github.com/open-telemetry/opentelemetry-collector-contrib/connector/signaltometricsconnector
--- a/connector/elasticapmconnector/testdata/metrics/service_summary_overflow/aggregated_metrics.yaml
+++ b/connector/elasticapmconnector/testdata/metrics/service_summary_overflow/aggregated_metrics.yaml
@@ -79,6 +79,9 @@ resourceMetrics:
         - key: service.name
           value:
             stringValue: "_other"
+        - key: overflow
+          value:
+            stringValue: resource
     scopeMetrics:
       - metrics:
           - name: service_summary
--- a/connector/elasticapmconnector/testdata/traces/span_metrics_overflow/aggregated_metrics.yaml
+++ b/connector/elasticapmconnector/testdata/traces/span_metrics_overflow/aggregated_metrics.yaml
@@ -31,6 +31,9 @@ resourceMetrics:
                     - key: processor.event
                       value:
                         stringValue: metric
+                    - key: overflow
+                      value:
+                        stringValue: metric
           - description: Overflow metric count due to metric limit
             name: _overflow_metric
             sum:
@@ -47,6 +50,9 @@ resourceMetrics:
                     - key: processor.event
                       value:
                         stringValue: metric
+                    - key: overflow
+                      value:
+                        stringValue: metric
           - description: Overflow metric count due to metric limit
             name: _overflow_metric
             sum:
@@ -61,6 +67,9 @@ resourceMetrics:
                       value:
                         stringValue: 60m
                     - key: processor.event
+                      value:
+                        stringValue: metric
+                    - key: overflow
                       value:
                         stringValue: metric
           - name: service_summary


connector/elasticapmconnector/README.md
```diff
@@ -39,11 +39,43 @@ By default, aggregated metrics will be exported without any client metadata. It
...
 ```yaml
 elasticapm:
   aggregation:
     directory: /path/to/aggregation/directory
     metadata_keys: [list, of, metadata, keys]
+    resource_limit:
+      max_cardinality: 8000
+      overflow:
+        attributes:
+          - key: "overflow"
```
isaacaflores2, line 59: Note: I was thinking we could probably not expose these attributes for this component, so a user just defines the max cardinality for each limit. Or at least define some defaults such as: 
```yaml
overflow:
        attributes:
          - key: "overflow"
            value: true
          - key: "component"
             value: "connector/elasticapmconnector"
```

connector/elasticapmconnector/README.md
```diff
@@ -39,11 +39,43 @@ By default, aggregated metrics will be exported without any client metadata. It
 propagate client metadata from input to exported metrics by specifying a list of metadata keys
 in `elasticapm::aggregation::metadata_keys`.
 
+By default, cardinality for aggregated metrics will be limited.
+Each limit defines a `max_cardinality` and a list of `attributes` that will be added to the overflow metric.
+There are four limits that can be configured: 
+- `elasticapm::aggregation::resource_limit`: configures the max cardinality of resources
+- `elasticapm::aggregation::scope_limit`: configures the max cardinality of scopes within a resource
+- `elasticapm::aggregation::metric_limit`: configures the max cardinality of metrics within a scope
+- `elasticapm::aggregation::datapoint_limit`: configures the max cardinality of metrics within a scope
```
1pkg, line 48: should the description be different between `metric_limit` and `datapoint_limit`?

connector/elasticapmconnector/README.md
isaacaflores2: I added default overflow attribute values

connector/elasticapmconnector/README.md
```diff
@@ -39,11 +39,43 @@ By default, aggregated metrics will be exported without any client metadata. It
...
+      overflow:
+        attributes:
+          - key: "overflow"
+            value: true
+    metric_limit:
+      max_cardinality: 4000
+      overflow:
+        attributes:
+          - key: "overflow"
+            value: true
```
lahsivjar, line 72: I don't think this is right, could you double check the attributes that are published for apm-aggregation for overflow bucket: https://github.com/elastic/apm-aggregation/blob/main/aggregators/converter.go

We should keep them same as the UI relies on it. https://github.com/elastic/opentelemetry-collector-components/issues/141#issuecomment-2592951016 comment can probably aid in understanding how lsminterval overflow can be used for achieving parity with apm-aggregation

connector/elasticapmconnector/config.go
```diff
@@ -64,6 +65,18 @@ type AggregationConfig struct {
...
+	ResourceLimit lsmconfig.LimitConfig `mapstructure:"resource_limit"`
+
+	// ScopeLimit defines the max cardinality of scopes within a resource
+	ScopeLimit lsmconfig.LimitConfig `mapstructure:"scope_limit"`
+
+	// MetricLimit defines the max cardinality of metrics within a scope
+	MetricLimit lsmconfig.LimitConfig `mapstructure:"metric_limit"`
+
+	// DatapointLimit defines the max cardinality of datapoints within a metric
+	DatapointLimit lsmconfig.LimitConfig `mapstructure:"datapoint_limit"`
```
lahsivjar, line 79: These configs are too generic for `elasticapmconnector`. I would propose to have configurations for `service_summary`, `service_transaction`, `transaction`, and `service_destination` and using them to configure the correct lsm interval configurations rather than directly exposing lsminterval configs. Again, https://github.com/elastic/opentelemetry-collector-components/issues/141#issuecomment-2592951016 might help a bit in understanding how we can do this.

lahsivjar (commented):
A few comments on the approach for the configurations.

connector/elasticapmconnector/README.md
isaacaflores2: Ah okay I did not know the UI relies on certain overflow attribute values. I will add `_other` in the appropriate places

connector/elasticapmconnector/config.go
isaacaflores2: Hmm its not clear to me how we can map configurations for `service_summary`, `service_transaction`, `transaction`, and `service_destination` to lsminterval configs. To me it seems that these would apply to the `resource_limit` only. I am most likely missing something. I will spend more time understanding the  `elasticapmconnector` metrics and how they affect the different levels of cardinality in the `lsmintervalprocessor`.

connector/elasticapmconnector/README.md
isaacaflores2: I updated to define only the `service.name: _other` attribute. The apm-aggregator defines separate overflow attributes for `transaction.name` and `transaction.type`. We would have to always apply these attributes for every metric which I do not think we want. For example the `service_summary` metric should probably not have these transaction attributes. I am double checking with the kibana team if they still only rely on  `service.name: _other`

isaacaflores2: @lahsivjar  I updated the connector only have one default overflow attribute for now `service.name: _other`. 

I created https://github.com/elastic/opentelemetry-collector-components/issues/697 since I could not find a simple way to only apply `transaction.name` for span aggregation metrics.

lahsivjar: Sounds good, however take a look at the comment I posted on where the `service.name` should be overridden.

For adding overflow attriburtes to other entities, we had plans to do this via OTTL conditions but that was not done yet. For now, we can add a generic identifier to those so that we can debug overflows and in future (once #697 is addressed) we can update those configs.

connector/elasticapmconnector/config.go
```diff
@@ -87,13 +109,41 @@ func (cfg Config) lsmConfig() *lsmconfig.Config {
...
+			Overflow:       defaultOverflowConfig,
+		}
+		lsmConfig.MetricLimit = lsmconfig.LimitConfig{
+			MaxCardinality: cfg.Aggregation.Limit.MetricLimit.MaxCardinality,
+			Overflow:       defaultOverflowConfig,
+		}
+		lsmConfig.DatapointLimit = lsmconfig.LimitConfig{
+			MaxCardinality: cfg.Aggregation.Limit.DatapointLimit.MaxCardinality,
+			Overflow:       defaultOverflowConfig,
+		}
```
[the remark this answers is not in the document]
isaacaflores2: @lahsivjar Ah good point. I updated to only apply the `service.name` to the resource limit

lahsivjar (approved):
LGTM! Just one minor comment about putting configs in files in testadata directory which can be addressed in https://github.com/elastic/opentelemetry-collector-components/pull/698.

Thanks for sticking with this and apologies about the late review.

connector/elasticapmconnector/connector_test.go
```diff
@@ -36,25 +37,43 @@ import (
...
 		name string
+		cfg  *Config
 	}{
-		{name: "logs/service_summary"},
+		// output should remain the same for all provided configs
+		{name: "logs/service_summary", cfg: &Config{}},
+		{name: "logs/service_summary", cfg: &oneCardinalityAggregationConfig},
+
+		// output should show overflow behavior
+		{name: "logs/service_summary_overflow", cfg: &oneCardinalityAggregationConfig},
```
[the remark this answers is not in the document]
isaacaflores2: @lahsivjar I think that makes sense. It also helps continue to see configuration from the "users" experience when they use col-config.yaml

Issue #612, opened 2025-06-11
[extension/apmconfig] memleak in `remoteConfigCallbacks.agentState`?


trentm:
```go
type remoteConfigCallbacks struct {
	*types.Callbacks
	configClient apmconfig.RemoteConfigClient

	agentState sync.Map
	logger     *zap.Logger
}
```

Is this a possible memleak? The only code evicting entries in `agentState` is this:
https://github.com/elastic/opentelemetry-collector-components/blob/9c8f30c1c7ca4f9ac15e7d5602931d6db056cecb/extension/apmconfigextension/opamp_callbacks.go#L108-L112 

However, my (possibly naive) understanding is that OpAMP clients **using the HTTP transport** do *not* typically (ever?) send an AgentToServer message with `agent_disconnect` set.   The Node.js OpAMP client (which only supports the HTTP transport) does not ever use `AgentDisconnect`. I'm not sure about Java and Python implementations.

---

The only other mention of `AgentDisconnect` in the current OpAMP spec is in regards to *websocket* connection handling: https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#duplicate-websocket-connections

The opamp proto includes the following comment:

```
 * AgentDisconnect is the last message sent from the Agent to the Server. The Server
 * SHOULD forget the association of the Agent instance with the message stream.
 *
 * If the message stream is closed in the transport layer then the Server SHOULD
 * forget association of all Agent instances that were previously established for
 * this message stream using AgentConnect message, even if the corresponding
 * AgentDisconnect message were not explicitly received from the Agent.
```

My opinion is that this was all written for the initial websocket-only design of OpAMP. The HTTP transport support is somewhat bolted on, so the guidance when using an HTTP transport is somewhat loose.

---

FWIW, the OpAMP server I wrote (for dev and testing-only) uses a cache/mapping of "active agents" that evicts entries after a certain amount of time:
https://github.com/elastic/elastic-otel-node/blob/3a6d4212b25fee9a5361c45204a894ef9c60bd21/packages/mockopampserver/lib/mockopampserver.js#L193-L199

xrmx:
Thanks @trentm for submitting this. I agree with the concern there should be some other mechanisms to free resources on the server. In the meantine I've just implemented `AgentDisconnect` in EDOT Python before exiting since it's easy to implement (sending a message with instance_uid, sequence_num and an empty AgentDisconnect instance in agent_disconnect) and appears to work:

```
[...]

PR #629, opened 2025-06-19
feat: replace sync.Map with freelru

rogercoll:
Fixes https://github.com/elastic/opentelemetry-collector-components/issues/612

### DIFF
--- a/extension/apmconfigextension/extension.go
+++ b/extension/apmconfigextension/extension.go
@@ -63,7 +63,7 @@ func (op *apmConfigExtension) Start(ctx context.Context, host component.Host) er
 		return err
 	}
 
-	opampCallbacks, err := newRemoteConfigCallbacks(remoteConfigClient, op.extensionConfig.OpAMP.Cache, op.telemetrySettings.Logger)
+	opampCallbacks, err := newRemoteConfigCallbacks(ctx, remoteConfigClient, op.extensionConfig.OpAMP.Cache, op.telemetrySettings.Logger)
 	if err != nil {
 		return err
 	}
--- a/extension/apmconfigextension/opamp_callbacks.go
+++ b/extension/apmconfigextension/opamp_callbacks.go
@@ -49,7 +49,7 @@ type agentInfo struct {
 	lastConfigHash        apmconfig.LastConfigHash
 }
 
-func newRemoteConfigCallbacks(configClient apmconfig.RemoteConfigClient, ttlConfig CacheConfig, logger *zap.Logger) (*remoteConfigCallbacks, error) {
+func newRemoteConfigCallbacks(ctx context.Context, configClient apmconfig.RemoteConfigClient, ttlConfig CacheConfig, logger *zap.Logger) (*remoteConfigCallbacks, error) {
 	cache, err := freelru.NewSharded[string, *agentInfo](ttlConfig.Capacity, func(key string) uint32 {
 		return uint32(xxhash.Sum64String(key))
 	})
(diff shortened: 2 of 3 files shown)


extension/apmconfigextension/opamp_callbacks.go
```diff
@@ -22,20 +22,27 @@ import (
...
+	"github.com/elastic/go-freelru"
 	"github.com/elastic/opentelemetry-collector-components/extension/apmconfigextension/apmconfig"
 	"github.com/open-telemetry/opamp-go/protobufs"
 	"github.com/open-telemetry/opamp-go/server/types"
 	"go.uber.org/zap"
 )
 
+const (
+	cacheCapacity = 1000
+	cacheTTL      = 10 * time.Second
```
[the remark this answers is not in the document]
rogercoll: Ah, missed this comment.

Yes, changed to x4 the default heartbeat here: https://github.com/elastic/opentelemetry-collector-components/pull/629/commits/1f80fdf82650e42a2e3dc68a7dc15882bc968d80

wdyt?

xrmx (approved):

extension/apmconfigextension/opamp_callbacks.go
```diff
@@ -45,11 +58,22 @@ type agentInfo struct {
 	lastConfigHash        apmconfig.LastConfigHash
 }
 
-func newRemoteConfigCallbacks(configClient apmconfig.RemoteConfigClient, logger *zap.Logger) *remoteConfigCallbacks {
+func newRemoteConfigCallbacks(configClient apmconfig.RemoteConfigClient, ttlConfig cacheConfig, logger *zap.Logger) (*remoteConfigCallbacks, error) {
+	cache, err := freelru.NewSharded[string, *agentInfo](ttlConfig.capacity, func(key string) uint32 {
+		h := fnv.New32a()
```
rockdaboot, line 63: FNV1a is one of the worst algorithms for hashing in regards to performance and randomness. Though it doesn't matter if the hash/lru operations are part of a hot path.

extension/apmconfigextension/extension.go
```diff
@@ -36,6 +37,14 @@ const (
 	defaultOpAMPPath = "/v1/opamp"
 )
 
+var defaultCacheConfig = cacheConfig{
+	// Cache capacity for active agents
+	capacity: 1000,
```
rockdaboot, line 42: nit: use 1024 instead of 1000 because computers work best with 2^N :)

extension/apmconfigextension/opamp_callbacks.go
```diff
@@ -106,26 +130,26 @@ func (rc *remoteConfigCallbacks) onMessage(ctx context.Context, conn types.Conne
 	agentUidField := zap.String("instance_uid", agentUid)
 	if message.GetAgentDisconnect() != nil {
 		rc.logger.Info("Disconnecting the agent from the remote configuration service", agentUidField)
-		rc.agentState.Delete(agentUid)
+		_ = rc.agentState.Remove(agentUid)
```
rockdaboot, line 133: If i understand correctly, the fear is that disconnect messages may become lost and cache entries pile up (-> memory leak).

With the use of an LRU you limit the number of cache entries. So you limit the amount of leaked memory. That's a good step forward.
Ideally, you also have go routine that regularly calls PurgeExpired() to actively purge expired entries from the cache. A health check every 30s was mentioned - that might be a good place to call PurgeExpired(). The alternative is to start a small go routine dedicated for this task.

extension/apmconfigextension/extension.go
rogercoll: Good point, let's be friendly with computers then :) https://github.com/elastic/opentelemetry-collector-components/pull/629/commits/308f6995550b189eac6e2a6f958f0dd6de7914d0

extension/apmconfigextension/opamp_callbacks.go
rogercoll: Sounds like a good improvement to save some memory, purging cache based on the TTL https://github.com/elastic/opentelemetry-collector-components/pull/629/commits/13b77750303396e7b669fcb5da3030ea532e7384

wdyt?

ChrsMark (approved):
I'm not really familiar with `freelru` but the patch looks good to me in general. Please ensure that the `freelru` comments are also addressed :).

extension/apmconfigextension/opamp_callbacks.go
rockdaboot: LGTM

Issue #370, opened 2025-02-20
elasticapmconnector: switch back to exponential histograms


axw:
We should ideally be using exponential histograms for latency distributions to maintain high precision. We have temporarily switched to explicit bucket histograms due to poor merging performance for exponential.

PR #630, opened 2025-06-19
[connector/elasticapm]Use exponential histogram from txn metrics

lahsivjar:
With the fixes for exponential histogram, it should perform on-par with histograms ([slack thread](https://elastic.slack.com/archives/C0680QCJKMM/p1743179432353619)).

Closes: https://github.com/elastic/opentelemetry-collector-components/issues/370


axw (approved):
Thanks!

PR #621, opened 2025-06-17
lsmintervalprocessor: add internal telemetry

1pkg:
## Context

This PR adds additional telemetry for `lsmintervalprocessor` to expose some basic processing stats and internal metrics from pebble.

## Results

image

PR #631, opened 2025-06-21
lsmintervalprocessor: fix telemetry builder shutdown and add internal telemetry tests

1pkg:
A followup PR to https://github.com/elastic/opentelemetry-collector-components/pull/621 that adds forgotten unit tests to verify that the internal metrics are reported as they should. It additionally fixes `shutdown` call for the internal metrics telemetry builder.


vigneshshanmugam (approved):
LGTM, thanks you

PR #634, opened 2025-06-25
Rename elastictrace processor to elasticapm and extend to logs and metrics

gregkalapos:
Extends the elastictrace processor to also enrich metrics and logs in addition to metrics.

Also renames the processor to reflect this change.


extension/apmconfigextension/factory.go
```diff
@@ -54,7 +54,7 @@ func createDefaultConfig() component.Config {
 	defaultElasticSearchClient := configelasticsearch.NewDefaultClientConfig()
 	httpCfg := confighttp.NewDefaultServerConfig()
 	httpCfg.Endpoint = "localhost:4320"
-	httpCfg.TLSSetting = nil
+	// httpCfg.TLSSetting = nil
```
gregkalapos, line 57: This does not seem to exist -

```
# github.com/elastic/opentelemetry-collector-components/receiver/elasticapmreceiver
../receiver/elasticapmreceiver/factory.go:61:22: defaultServerConfig.TLSSetting undefined (type confighttp.ServerConfig has no field or method TLSSetting)
```

What's strange is that, from what I understand this comes from the `confighttp` package, which I don't changer here - it should be the same version as before the PR

receiver/elasticapmreceiver/factory.go
```diff
@@ -58,7 +58,7 @@ func createDefaultConfig() component.Config {
 	defaultESClientConfig.Endpoint = defaultESEndpoint
 
 	// TODO: Remove this once we have a proper way to configure TLS
-	defaultServerConfig.TLSSetting = nil
+	//defaultServerConfig.TLSSetting = nil
```
gregkalapos, line 61: Same as in `extension/apmconfigextension/factory.go`

receiver/elasticapmreceiver/agentcfg.go
```diff
@@ -124,9 +124,7 @@ func buildQuery(r *http.Request) (agentcfg.Query, error) {
 			},
 		}
 	default:
-		if err := fmt.Errorf("%s: %s", msgMethodUnsupported, r.Method); err != nil {
-			return query, err
-		}
+		return query, fmt.Errorf("%s: %s", msgMethodUnsupported, r.Method)
```
gregkalapos, line 127: This was a golint error:

```
INFO [runner] linters took 324.3555ms with stages: goanalysis_metalinter: 323.816583ms 
elasticapmreceiver/agentcfg.go:127:67: SA4023: this comparison is always true (staticcheck)
		if err := fmt.Errorf("%s: %s", msgMethodUnsupported, r.Method); err != nil {
		                                                                ^
elasticapmreceiver/$GOROOT/src/fmt/errors.go:22:1: SA4023(related information): fmt.Errorf never returns a nil interface value (staticcheck)
```

PR #639, opened 2025-06-30
Upgrade upstream open-telemetry components to v0.129.0

lahsivjar:


pkg/integrations/testdata/template-unknown-fields.yaml
```diff
@@ -17,7 +17,6 @@ extensions:
   someextension:
 
 pipelines:
-  extensions: [somextension]
```
lahsivjar, line 20: @jsoriano I had to change this to produce the same expected error (`field extensions not found`). I took a brief look at why this was happening and I think this is because the new version of upstream otel collector is failing the pipelines deocding from yaml with error like `invalid integration template format: in \"extensions\" id: unknown pipeline signal: \"extensions\"` and thus the expected error is not triggered. Let me know if it doesn't make sense or if I am understanding something incorrectly.

axw (approved):

rogercoll (approved):

Issue #590, opened 2025-05-30
Rename elasticapmreceiver and agree on the new name


gregkalapos:
### Suggestion for the new name: `elasticintakev2`receiver

@rogercoll noted in our weekly that we have two components both named `elasticapmre`:

- elasticapmreceiver
- elasticapmconnector

A connector and a receiver can't have the same id in the config, so in a default setup this causes an error. It can be workarounded by adding a specific id to those, then everything is good, but forcing adding an id by default when both components are used is not nice.

Therefore, we agreed to rename the receiver. Above is the suggested name. Let's try to reach a consensus here, once we agreed on the name, I'll rename the receiver.

rogercoll:
> Can you share the error message? Is it just those two component types that clash? I'm surprised, because for example there are exporters and receivers that have the same name, e.g. kafkareceiver/kafkaexporter.

That's right, the issue is that the `elasticapmconnector` is a connector, so a receiver and exporter at the same time. The collision happens with two "receivers" named equally:

```
Error: invalid configuration: connectors::elasticapm: ambiguous ID: Found both "elasticapm" receiver and "elasticapm" connector. Change one of the components' IDs to eliminate ambiguity (e.g. rename "elasticapm" connector to "elasticapm/connector")
2025/05/27 12:48:04 collector server run finished with error: invalid configuration: connectors::elasticapm: ambiguous ID: Found both "elasticapm" receiver and "elasticapm" connector. Change one of the components' IDs to eliminate ambiguity (e.g. rename "elasticapm" connector to "elasticapm/connector")
```

axw:
`elasticintakev2receiver` makes sense to me

> Why the version in the name?

@theletterf "intakev2" is the name of the protocol that Elastic APM classic agents speak. Not the greatest name, but that's what it is :)

gregkalapos:
Alternatively, we could drop `v2` - for backend agents that's the only intake version that is supported, right? 

So maybe just `elasticintakereceiver`? 

I'd still prefer `elasticintakev2receiver` because it's more precise, but no strong feelings on this.

trentm:
(Naive Q.)   Will this thing ever support intake v3 for RUM?

AlexanderWert:
IMHO, `elasticintakereceiver` might be misleading. So I would propose one of those:
- `elasticapmintakereceiver`
- `elasticintakev2receiver`  <- with a slight preference for that

gregkalapos:
[...]

PR #646, opened 2025-07-07
[receiver/elasticapmreceiver] Rename to `elasticapmintakereceiver`

girodav:
Rename `elasticapmreceiver` to `elasticapmintakereceiver`.

- closes https://github.com/elastic/opentelemetry-collector-components/issues/590


gregkalapos (approved):

gregkalapos:
> I don't have the full context for this, but would it make sense to deprecate the old receiver for one release (rather than deleting it right away), just to make it easier for folks currently using it?

This receiver has not yet been released - we have not added it to EDOT either. So this is a really good time to rename it, so we don't need to take care of those things.

PR #650, opened 2025-07-09
apikeyauthextension: propagate status code in errs

marclop:
Updates the extension to return a `status.Error` type (which wraps the `status.Status` type) instead of a plain `error` type. This allows propagating the status code and message from the extension to the collector.

This is similar to what the upstream OTLP receiver does:

https://github.com/open-telemetry/opentelemetry-collector/blob/v0.129.0/receiver/otlpreceiver/otlphttp.go#L183-L192

### DIFF
--- a/extension/apikeyauthextension/authenticator.go
+++ b/extension/apikeyauthextension/authenticator.go
@@ -224,6 +224,10 @@ func (a *authenticator) getCacheKey(id string, headers map[string][]string) (str
 
 // Authenticate validates an ApiKey scheme Authorization header,
 // passing it to Elasticsearch for checking privileges.
+//
+// Callers can use status.FromError(err) to get the status code
+// and message from the returned error. If no status.Status is returned,
+// the error should be considered an internal error.
 func (a *authenticator) Authenticate(ctx context.Context, headers map[string][]string) (context.Context, error) {
 	authHeaderValue, id, err := a.parseAuthorizationHeader(headers)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


extension/apikeyauthextension/authenticator.go
```diff
@@ -245,27 +247,29 @@ func (a *authenticator) Authenticate(ctx context.Context, headers map[string][]s
 			// Client has specified an API Key with a colliding ID,
 			// but whose secret component does not match the one in
 			// the cache.
-			return ctx, fmt.Errorf("API Key %q unauthorized", id)
+			return ctx, status.Errorf(codes.InvalidArgument,
```
axw, line 250: Why is this one InvalidArgument? I would expect it to be Unauthenticated too

extension/apikeyauthextension/authenticator.go
```diff
@@ -225,12 +227,12 @@ func (a *authenticator) getCacheKey(id string, headers map[string][]string) (str
...
 	authHeaderValue, id, err := a.parseAuthorizationHeader(headers)
 	if err != nil {
-		return ctx, err
+		return ctx, status.Error(codes.Unauthenticated, err.Error())
 	}
 
 	cacheKey, err := a.getCacheKey(id, headers)
 	if err != nil {
-		return ctx, err
+		return ctx, status.Error(codes.Internal, err.Error())
```
axw, line 235: Doesn't really hurt, but I'd be inclined to not use `status.Error` here or in the other Internal case, and expect callers to assume errors are internal unless told otherwise

axw (commented):
Looks good, main concern is the InvalidArgument case. It may leak information

extension/apikeyauthextension/authenticator.go
marclop: good point

Issue #619, opened 2025-06-17
ratelimitprocessor: ThrottleBehaviorDelay misinterprets reset time


axw:
The rate limit processor has blocking (delay) and non-blocking (error) modes. By default it is non-blocking.

The blocking mode is handled in this code:

https://github.com/elastic/opentelemetry-collector-components/blob/93132ef675a366fc986c39499db7e63fca7d4e9a/processor/ratelimitprocessor/gubernator.go#L169-L177

This code is essentially sleeping until the reset time, and then continuing. The reset time is when the rate limit resets - see https://github.com/gubernator-io/gubernator/blob/427194a6c593e6d9da7283fbaea49939052c8676/gubernator.proto#L204-L205. Imagine the rate limit is exhausted and >1 clients hit this code path: they will both be allowed through at the reset time, regardless of how many items they request.

I think there should be a loop here: sleep until the reset, then request again with the remaining items, and repeat as necessary.

PR #660, opened 2025-07-14
Fix delay behavior

constanca-m:
Issue is https://github.com/elastic/opentelemetry-collector-components/issues/619.


processor/ratelimitprocessor/gubernator.go
```diff
@@ -169,13 +203,16 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 			)
 			return status.Error(codes.ResourceExhausted, errTooManyRequests.Error())
 		case ThrottleBehaviorDelay:
-			delay := time.Duration(resp.GetResetTime()-createdAt) * time.Millisecond
+			current := r.addRequests(uniqueKey, hits)
+			delay := time.Duration(resp.GetResetTime()-createdAt+int64(current)*cfg.ThrottleInterval.Milliseconds()) * time.Millisecond
```
constanca-m, line 207: @vigneshshanmugam Can I ask your opinion on this logic before I do the same for the local rate limiter?

This bug was added to GA goals: https://github.com/elastic/opentelemetry-collector-components/issues/619

This implementation has some things that worry me, but I don't if it makes sense to be worried over this:
1. What if we receive tons of requests? The delay will be increasing for each request, and then we will have many processes on hold. I think it makes sense to maybe set the delay to a fixed time, and after that time, each request starts getting rejected, regardless of the strategy. WDYT?
2. What if we receive requests like this:
    - We get a request that takes 10 tokens
    - We then get a request that takes 4 tokens
    - Both are delayed. Should the request of 4 tokens take priority, thus less delay for it, or should the 10 tokens have priority since it arrived first?

vigneshshanmugam: I dont think you need to do this, The issue here is that we are allowing all the clients to retry and send data simultanously once we past the reset time which is wrong. Its assuming the reset time guarentees the availability of tokens - requests/bytes based on the strategy. 

The ideal way we could fix is basically looping through and retry after the delay, re checking the rate limits at each time. This respects all the configurations like strategy/algorithm along with keeping the race condition in check. 

Rough sudo code would be
```go
for {
  // check rate limit
 resp = getCurrentLimits()
 if resp.IsUnderLimit {
  // not limited
   return nil
 }
 
// we shouldn't use createAt here its flawed as its based on when the request is made and doesnt take retries in to account. 
  delay := time.Duration(resp.GetResetTime()-time.Now().UnixMilli()) * time.Millisecond
  // same code as before



} 

```

Hope this helps. Let me know if you want more details.

processor/ratelimitprocessor/gubernator.go
```diff
@@ -161,13 +167,24 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
...
+		retry:
+			for {
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case <-timer.C:
+					if err := makeRateLimitRequest(); err != nil {
+						return err
+					}
+					if resp.GetStatus() == gubernator.Status_UNDER_LIMIT {
```
vigneshshanmugam, line 182: this feels incorrect, we are using the old response, we need to get the new response after retry.

vigneshshanmugam (approved):

PR #665, opened 2025-07-16
Adding back elastictrace processor for smooth deprecation

gregkalapos:
The PR adds back `elastictrace` processor and upadtes it to OTel `v0.130.0`.

Discussed via Slack with @andrzej-stencel and @AlexanderWert 

### Reasoning:
- In order to enable a smooth transition from `elastictrace` processor to `elasticapm` processor, we still need the `elastictrace` processor in the cade-base with OTel `v0.130.0` in order to have 9.1 on `v0.130.0`.

### The plan
- In 9.1 we'll only have `elastictrace` processor and no `elasticapm` processor. This gives us more time to test `elasticapm` processor.
- In 9.2 we'll add the `elasticapm` processor and mark `elastictrace` processor as deprecated. 

### Notes
- So far I have not ran into any API braking changes when updating `elastictrace` processor to `v0.130.0` -  it compiles after updating `go.mod` as is 🤔 
- I do not mark `elastictrace` processor as deprecated yet, since in 9.1 we'll only have `elastictrace` processor - so we should mark it as deprecated after 9.1 is out.

Feel free to correct if anything is wrong in above statements.


processor/elastictraceprocessor/go.mod
```diff
@@ -0,0 +1,73 @@
+module github.com/elastic/opentelemetry-collector-components/processor/elastictraceprocessor
+
+go 1.23.8
+
+require (
+	github.com/elastic/opentelemetry-lib v0.18.0
```
gregkalapos, line 6: I suggest purposely leaving this one on this version - even if there are newer releases to be on the safe side. Most of the logic of this processor is actually in this lib - the purpose to keep the `elastictrace` processor around is to have a smooth transition. For this, I'd not pull in changes that significantly extended the responsibility of `opentelemetry-lib` - I'd rather just keep it as is and then use the newer versions in the `elasticapm` processor.

rogercoll (approved):

PR #687, opened 2025-07-29
profilingmetricsconnector: add (costum)? classification

florianl:
Allow more detailed aggregation and classification of metrics from profiling information.

Treemap visualization of the top 25 entries generated by classification:

### DIFF
--- a/connector/profilingmetricsconnector/config.go
+++ b/connector/profilingmetricsconnector/config.go
@@ -17,8 +17,25 @@
 
 package profilingmetricsconnector // import "github.com/elastic/opentelemetry-collector-components/connector/profilingmetricsconnector"
 
+// Aggregation applies Match as a regular expression on function strings
+// and generates a metric with Label if it matches.
+type Aggregation struct {
+	Match string `mapstructure:"match"`
+	Label string `mapstructure:"label"`
+}
+
 // Config represents the receiver config settings within the collector's config.yaml
 type Config struct {
 	// MetricsPrefix is the prefix to add to all generated metric names
 	MetricsPrefix string `mapstructure:"metrics_prefix"`
+
+	// Generate metrics based on frame type.
+	ByFrameType bool `mapstructure:"by_frametype"`
+
+	// Generate metrics based on functional classification.
+	// Currently only Java and Go are supported.
+	ByClassification bool `mapstructure:"by_classification"`
+
+	// CustomAggregations allows to generate costum metrics.
+	CustomAggregations []Aggregation `mapstructure:"aggregations"`
 }
--- a/connector/profilingmetricsconnector/config.go
+++ b/connector/profilingmetricsconnector/config.go
@@ -36,6 +36,6 @@ type Config struct {
 	// Currently only Java and Go are supported.
 	ByClassification bool `mapstructure:"by_classification"`
 
-	// CustomAggregations allows to generate costum metrics.
+	// CustomAggregations allows to generate custom metrics.
 	CustomAggregations []Aggregation `mapstructure:"aggregations"`
 }
--- a/connector/profilingmetricsconnector/connector.go
+++ b/connector/profilingmetricsconnector/connector.go
@@ -77,7 +77,7 @@ func (c *profilesToMetricsConnector) Start(ctx context.Context, host component.H
 		// Precompile all regular expressions at startup to do it only once.
 		re, err := regexp.Compile(agg.Match)
 		if err != nil {
-			return err
+			return fmt.Errorf("compiling rx for label %v: %v", agg.Label, err)
 		}
 		c.aggregations = append(c.aggregations, aggregation{
 			re:    re,
@@ -175,13 +175,13 @@ func (c *profilesToMetricsConnector) extractMetricsFromScopeProfiles(dictionary 
 		}
 
 		if len(c.aggregations) > 0 {
-			costumAggregationCounts := make(map[string]int64)
+			customAggregationCounts := make(map[string]int64)
 			for _, sample := range profile.Sample().All() {
-				c.collectCostumAggregationCounts(dictionary, locIndices, sample, costumAggregationCounts)
+				c.collectCustomAggregationCounts(dictionary, locIndices, sample, customAggregationCounts)
 			}
 
 			// Add metric for custom aggregations.
-			c.addMetrics(origin, costumAggregationCounts,
+			c.addMetrics(origin, customAggregationCounts,
 				"samples.custom_aggregation", "Number of profiling frames by custom aggregation", "aggregation",
 				scopeMetrics, profile.Time())
 		}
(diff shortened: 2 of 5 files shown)


connector/profilingmetricsconnector/config.go
```diff
@@ -17,8 +17,25 @@
...
 	MetricsPrefix string `mapstructure:"metrics_prefix"`
+
+	// Generate metrics based on frame type.
+	ByFrameType bool `mapstructure:"by_frametype"`
+
+	// Generate metrics based on functional classification.
+	// Currently only Java and Go are supported.
+	ByClassification bool `mapstructure:"by_classification"`
+
+	// CustomAggregations allows to generate costum metrics.
```
christos68k, line 39:
```suggestion
	// CustomAggregations allows to generate custom metrics.
```

connector/profilingmetricsconnector/connector.go
```diff
@@ -54,6 +66,25 @@ func (c *profilesToMetricsConnector) ConsumeProfiles(ctx context.Context, profil
...
+	for _, agg := range c.config.CustomAggregations {
+		if agg.Label == "" || agg.Match == "" {
+			// Both, label and match, need to be set.
+			return errInvalAgg
+		}
+
+		// Precompile all regular expressions at startup to do it only once.
+		re, err := regexp.Compile(agg.Match)
+		if err != nil {
+			return err
```
christos68k, line 80: Maybe we can return a better error here that gives more context:

```suggestion
			return fmt.Errorf("compiling rx for label %v: %v", agg.Label, err)
```

connector/profilingmetricsconnector/connector.go
```diff
@@ -118,14 +149,42 @@ func (c *profilesToMetricsConnector) extractMetricsFromScopeProfiles(dictionary
...
+		}
+
+		if len(c.aggregations) > 0 {
+			costumAggregationCounts := make(map[string]int64)
+			for _, sample := range profile.Sample().All() {
+				c.collectCostumAggregationCounts(dictionary, locIndices, sample, costumAggregationCounts)
+			}
+
+			// Add metric for custom aggregations.
+			c.addMetrics(origin, costumAggregationCounts,
```
christos68k, line 184:
```suggestion
			customAggregationCounts := make(map[string]int64)
			for _, sample := range profile.Sample().All() {
				c.collectCustomAggregationCounts(dictionary, locIndices, sample, customAggregationCounts)
			}

			// Add metric for custom aggregations.
			c.addMetrics(origin, customAggregationCounts,
```

connector/profilingmetricsconnector/connector.go
```diff
@@ -159,26 +218,123 @@ func (c *profilesToMetricsConnector) collectFrameTypeCounts(dictionary pprofile.
...
+					continue
+				}
+				classificationCounts[hotspotInfo.pack+"."+hotspotInfo.class] += 1
+			}
+		}
+	}
+}
+
+// collectCostumAggregationCounts walks all locations/frames of a sample and collects costum aggregation information.
+func (c *profilesToMetricsConnector) collectCostumAggregationCounts(dictionary pprofile.ProfilesDictionary, locationIndices pcommon.Int32Slice, sample pprofile.Sample, costumAggregationCounts map[string]int64) {
```
christos68k, line 308:
```suggestion
// collectCustomAggregationCounts walks all locations/frames of a sample and collects costum aggregation information.
func (c *profilesToMetricsConnector) collectCustomAggregationCounts(dictionary pprofile.ProfilesDictionary, locationIndices pcommon.Int32Slice, sample pprofile.Sample, customAggregationCounts map[string]int64) {
```

connector/profilingmetricsconnector/connector.go
```diff
@@ -159,26 +218,123 @@ func (c *profilesToMetricsConnector) collectFrameTypeCounts(dictionary pprofile.
...
+		loc := locationTable.At(int(li))
+
+		for _, line := range loc.Line().All() {
+			fnIdx := line.FunctionIndex
+			fnEntry := funcTable.At(int(fnIdx()))
+			fnStr := strTable.At(int(fnEntry.NameStrindex()))
+
+			for _, agg := range c.aggregations {
+				if agg.re.MatchString(fnStr) {
+					costumAggregationCounts[agg.label] += 1
```
christos68k, line 331:
```suggestion
					customAggregationCounts[agg.label] += 1
```

connector/profilingmetricsconnector/connector.go
```diff
@@ -159,26 +218,123 @@ func (c *profilesToMetricsConnector) collectFrameTypeCounts(dictionary pprofile.
...
 		dataPoint.SetIntValue(count)
 		dataPoint.SetTimestamp(ts)
-
+		dataPoint.Attributes().PutStr(dataPointAttribute, typ)
 		dataPoint.Attributes().PutStr("profile.type_unit", origin.typ+"_"+origin.unit)
 	}
 }
 
+// collectClassificationCounts walks all locations/frames of a sample and collects the classification information.
+func (c *profilesToMetricsConnector) collectClassificationCounts(dictionary pprofile.ProfilesDictionary, locationIndices pcommon.Int32Slice, sample pprofile.Sample, classificationCounts map[string]int64) {
```
christos68k, line 245: If we  use a `map[string]int64` for `classificationCounts` we're dropping the frame type information, meaning we are mixing packages from Java and Go in the same map. 

If we instead use a `map[string]map[string]int64` we can keep frame type around which will later allow us to easily split the counts by frame type for visualizations, for example by adding a frametype attribute to each metric.

connector/profilingmetricsconnector/general_aggregation.go
```diff
@@ -0,0 +1,142 @@
...
+	//   - Matches the literal dot that separates the class name from the method name.
+	//
+	// `([a-zA-Z_][a-zA-Z0-9_.$<>+]+)`
+	//   - Second capturing group, for the method name itself.
+	//   - `[a-zA-Z_][a-zA-Z0-9_.$<>+]+`: Matches a valid method name, which can also include
+	//     `$` for synthetic methods (e.g., `lambda$fillGapRunnable$13`).
+	//
+	// `\(`
+	//   - Matches the literal opening parenthesis that signifies the start of method arguments.
+	hotspotRegex = regexp.MustCompile(`(?:\S+\s+)?([a-zA-Z_][a-zA-Z0-9_.$<>+]+(?:\.[a-zA-Z_][a-zA-Z0-9_.$<>+]+)*)\.([a-zA-Z_][a-zA-Z0-9_.$<>+]+)\(`)
```
christos68k, line 65: Since we know the frame is Hotspot interpreted, instead of using a regex isn't it simpler to split the input string by '.' and then:

`result[:-1]` will be the `fullClassName`
`result[-1]` will be the method call

If we want to get rid of the arguments, we can use this:

`result := re.sub('\(.*\) ?.*$', '', input)`

christos68k (approved):

connector/profilingmetricsconnector/general_aggregation.go
christos68k: We should first:

1. Get rid of the return type (it's always followed by a space, so splitting by `" "` and taking the element at index 1 will do it)
2. Get rid of the arguments (e.g. via simpler regex `"\(.*\) ?.*$"`)

florianl: With https://github.com/elastic/opentelemetry-collector-components/pull/687/commits/9735400e1dacce69cf0ab1d434c0b13831c86c29 I have replaces the regex matching with functions only using the strings package. Quick benchmarks also show, that just using strings funtions performs better in terms of computing time and also reduces allocations.

PR #334, opened 2025-01-24
[beatsauthextension] Add support for beats related ssl parameters to be used with otel-components

khushijain21:
This extension supports beats' related client ssl parameters such as [verification_mode](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-ssl.html#client-verification-mode) , [ca_trusted_fingerprint](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-ssl.html#ca_trusted_fingerprint) and [ca_sha256](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-ssl.html#ca-sha256) to be used with otel components using `extensions`

### Local Testing done as follows: 

```
receivers:
  otlpjsonfile:
    include_file_name: true
    include:
      - "./otlp-all.json"

extensions:
  beatsauth: 
    tls:
      verification_mode: full
      ca_trusted_fingerprint: "XXX"

exporters:
  debug:
    verbosity: detailed
  elasticsearch: 
   endpoints: 
    -  https://my-deployment-XXX:443
   password: XXXX
   user: elastic
   tls:
    ca_file: ./ca.crt
   auth: 
    authenticator: beatsauth  

service:
  extensions: [beatsauth]
  pipelines:
    logs:
      receivers: [otlpjsonfile]
      processors: []
      exporters: [debug, elasticsearch]
```

The `ca.crt` file contains a a self signed certificate.  And when `ca_trusted_fingerprint` is not used, the indexing fails. The indexing is successfull on passing correct `ca_trusted_fingerprint`.

image

PR #690, opened 2025-07-31
[beatsauthextension] Fix bug in `configureTransport` method

khushijain21:
Before this change, when `ca_trusted_fingerprint` was set, it would add this fingerprint to the system CA - ignoring any user provided certificate authority.

What it should actually do is add the trusted fingerprint to the incoming `rootCA`. 

It also sets the host name on `BuildModuleClientConfig` to perform correct hostname verification

For Ref: https://www.elastic.co/docs/reference/beats/filebeat/configuration-ssl#ca_trusted_fingerprint

Testing:

Previously, it was tested with hosted elasticsearch instance which is why we did not catch this bug.
This was now done by bringing up elasticsearch instance locally using self signed certificates.

 Testing similar to how its defined here https://github.com/elastic/opentelemetry-collector-components/pull/334

Also added test cases which verify this end-to-end


extension/beatsauthextension/authenticator.go
```diff
@@ -82,15 +82,11 @@ func (a *authenticator) configureTransport(transport *http.Transport) error {
 	if a.tlsConfig != nil {
 		// injecting verifyConnection here, keeping all other fields on TLSClientConfig intact
 
-		// should inject incoming root CA into our tls config
-		// so that we can build our custom verifyConnection method
-		// if root CA is empty, it uses system CA
+		// inject incoming root CA into our tls config
+		// because ca_trusted_fingerprint will be appended to incoming CA
 		a.tlsConfig.RootCAs = transport.TLSClientConfig.RootCAs
```
florianl, line 87: Maybe I'm missing something, but here the *x509.CertPool is replace with `transport.TLSClientConfig.RootCAs` rather than appended or extended. Is this expected?

extension/beatsauthextension/authenticator.go
```diff
@@ -78,12 +85,18 @@ func (a *authenticator) RoundTripper(base http.RoundTripper) (http.RoundTripper,
...
+		// copy incoming root CA into our tls config
+		// because ca_trusted_fingerprint will be appended to root CA
+		a.tlsConfig.RootCAs = transport.TLSClientConfig.RootCAs
+
+		beatTLSConfig := a.tlsConfig.BuildModuleClientConfig(transport.TLSClientConfig.ServerName)
 
 		transport.TLSClientConfig.VerifyConnection = beatTLSConfig.VerifyConnection
 		transport.TLSClientConfig.InsecureSkipVerify = beatTLSConfig.InsecureSkipVerify
+		transport.TLSClientConfig.RootCAs = a.tlsConfig.RootCAs
+
```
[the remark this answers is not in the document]
khushijain21: If `ca_trusted_fingerprint` exists, it is appended to `a.tlsConfig.RootCA`, and then we copy the new values back into out Transport. The `transport.TLSClientConfig` here is the one actually performing the TLS handshake. 

`a.tls.Config` is only meant for the extension to store the incoming config.

extension/beatsauthextension/authenticator.go
```diff
@@ -78,12 +85,17 @@ func (a *authenticator) RoundTripper(base http.RoundTripper) (http.RoundTripper,
 
 func (a *authenticator) configureTransport(transport *http.Transport) error {
 	if a.tlsConfig != nil {
-		// injecting verifyConnection here, keeping all other fields on TLSClientConfig intact
-		beatTLSConfig := a.tlsConfig.BuildModuleClientConfig(a.cfg.TLS.ServerName)
+		// copy incoming CertPool into our tls config
+		// because ca_trusted_fingerprint will be appended to CertPool
+		a.tlsConfig.RootCAs = transport.TLSClientConfig.RootCAs
```
axw, line 90: I think there may be a data race here? IIANM, the same `authenticator` instance may be used for multiple concurrent HTTP clients.

khushijain21: ah yes you are right, fixed it

extension/beatsauthextension/authenticator.go
```diff
@@ -77,13 +86,22 @@ func (a *authenticator) RoundTripper(base http.RoundTripper) (http.RoundTripper,
...
+
 	if a.tlsConfig != nil {
-		// injecting verifyConnection here, keeping all other fields on TLSClientConfig intact
-		beatTLSConfig := a.tlsConfig.BuildModuleClientConfig(a.cfg.TLS.ServerName)
+		a.mx.Lock()
+		defer a.mx.Unlock()
+
+		// copy incoming CertPool into our tls config
+		// because ca_trusted_fingerprint will be appended to CertPool
+		a.tlsConfig.RootCAs = transport.TLSClientConfig.RootCAs
```
axw, line 96:
```suggestion
		// copy incoming CertPool into our tls config
		// because ca_trusted_fingerprint will be appended to CertPool
		tlsConfig := *a.tlsConfig // copy before updating, configureTransport may be called concurrently
		tlsConfig.RootCAs = transport.TLSClientConfig.RootCAs
		beatTLSConfig := tlsConfig.BuildModuleClientConfig(transport.TLSClientConfig.ServerName)
```

I _think_ what you did works but it seems unnecessarily complicated to use a mutex. Can we make a copy of `a.tlsConfig` instead?

axw (commented):
Looks better, thanks, just a suggestion to simplify.

extension/beatsauthextension/authenticator.go
```diff
@@ -37,19 +39,26 @@ type authenticator struct {
 	cfg       *Config
 	telemetry component.TelemetrySettings
 	tlsConfig *tlscommon.TLSConfig // set by Start
+	logger    *logp.Logger
+	mx        sync.Mutex
```
axw, line 43: This can now be removed.

extension/beatsauthextension/authenticator.go
```diff
@@ -77,13 +86,21 @@ func (a *authenticator) RoundTripper(base http.RoundTripper) (http.RoundTripper,
...
 	if a.tlsConfig != nil {
-		// injecting verifyConnection here, keeping all other fields on TLSClientConfig intact
-		beatTLSConfig := a.tlsConfig.BuildModuleClientConfig(a.cfg.TLS.ServerName)
+
+		// copy incoming CertPool into our tls config
+		// because ca_trusted_fingerprint will be appended to CertPool
+		tlsConfig := *a.tlsConfig // copy before updating, configureTransport may be called concurrently
+		tlsConfig.RootCAs = transport.TLSClientConfig.RootCAs
+
+		beatTLSConfig := a.tlsConfig.BuildModuleClientConfig(transport.TLSClientConfig.ServerName)
```
axw, line 97:
```suggestion
		beatTLSConfig := tlsConfig.BuildModuleClientConfig(transport.TLSClientConfig.ServerName)
```

PR #696, opened 2025-08-05
[processor/ratelimiter] Add ratelimit.request_size histogram metric

constanca-m:
### Results

In a document:

```json
"otelcol_ratelimit": {
      "request_size": {
        "values": [
          175
        ],
        "counts": [
          2
        ]
      }
    },
```

In a dashboard:

Screenshot From 2025-08-05 15-13-02

### DIFF
--- a/processor/ratelimitprocessor/metadata.yaml
+++ b/processor/ratelimitprocessor/metadata.yaml
@@ -35,7 +35,39 @@ telemetry:
       unit: "{bytes}"
       histogram:
         value_type: int
-        bucket_boundaries: [ 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2000, 3000, 4000, 5000, 6000, 7000, 8000, 9000, 10000, 20000, 30000, 50000, 100_000, 200_000, 300_000, 400_000, 500_000, 600_000, 700_000, 800_000, 900_000, 1000_000, 2000_000, 3000_000, 4000_000, 5000_000, 6000_000, 7000_000, 8000_000, 9000_000 ]
+        bucket_boundaries: [
+          # Small requests (<1MB)
+          512,        # 0.5 KB
+          1024,       # 1 KB
+          2048,       # 2 KB
+          4096,       # 4 KB
+          8192,       # 8 KB
+          16384,      # 16 KB
+          32768,      # 32 KB
+          65536,      # 64 KB
+          131072,     # 128 KB
+          262144,     # 256 KB
+          524288,     # 512 KB
+
+          # Medium requests (1-10MB)
+          1048576,    # 1 MB
+          2097152,    # 2 MB
+          3145728,    # 3 MB
+          4194304,    # 4 MB
+          5242880,    # 5 MB
+          6291456,    # 6 MB
+          7340032,    # 7 MB
+          8388608,    # 8 MB
+          9437184,    # 9 MB
+          10485760,   # 10 MB
+
+          # Large requests (up to OTLP HTTP limit)
+          12582912,   # 12 MB
+          14680064,   # 14 MB
+          16777216,   # 16 MB
+          18874368,   # 18 MB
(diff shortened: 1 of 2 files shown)


processor/ratelimitprocessor/metadata.yaml
```diff
@@ -29,6 +29,13 @@ telemetry:
         value_type: double
         monotonic: true
         bucket_boundaries: [ 0.0001, 0.0005, 0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.5, 1.0, 5.0, 10.0 ]
+    ratelimit.request_size:
```
vigneshshanmugam, line 32: Histogram is a wrong choice here, let's use a counter similar to how measure bytes in other components. We could do sum aggregation on top of it to get throughout per second etc.

Example - https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ee30200e7499ce220793f87850f9f98b4bb4b83b/exporter/kafkaexporter/internal/metadata/generated_telemetry.go#L33

constanca-m: Aah, so we want the total instead? Got it, I thought we wanted an idea of the size of the requests we were getting

vigneshshanmugam: I take that back, Thinking this through - given we already have the throttled request count per tenant, Lets stick with histogram as we can calculate the distribution clearly not just the volume.

processor/ratelimitprocessor/metadata.yaml
```diff
@@ -29,6 +29,13 @@ telemetry:
...
+      enabled: true
+      description: Number of bytes in received request. Only available if strategy is rate per bytes.
+      unit: "{bytes}"
+      histogram:
+        value_type: int
+        bucket_boundaries: [ 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2000, 3000, 4000, 5000, 6000, 7000, 8000, 9000, 10000, 20000, 30000, 50000, 100_000, 200_000, 300_000, 400_000, 500_000, 600_000, 700_000, 800_000, 900_000, 1000_000, 2000_000, 3000_000, 4000_000, 5000_000, 6000_000, 7000_000, 8000_000, 9000_000 ]
```
vigneshshanmugam, line 38: One suggestion would be to stick with something like this as its acting on the uncompressed data, so its easier to model. 

```
[ 
  512,         # 0.5 KB
  1024,        # 1 KB
  2048,        # 2 KB
  4096,        # 4 KB
  8192,        # 8 KB
  16384,       # 16 KB
  32768,       # 32 KB
  65536,       # 64 KB
  131072,      # 128 KB
  262144,      # 256 KB
  524288,      # 512 KB
  1048576,     # 1 MB
  2097152,     # 2 MB
  4194304,     # 4 MB
  8388608,     # 8 MB
  16777216,    # 16 MB
  33554432,    # 32 MB
  67108864,    # 64 MB
  100663296    # ~96 MB 
]
```

vigneshshanmugam (approved):
LGTM

processor/ratelimitprocessor/metadata.yaml
constanca-m: I stole the bucket boundaries from the batch processor: https://github.com/open-telemetry/opentelemetry-collector/blob/42a3ae04a72f69119f052ef9593dfc2ea537ebbf/processor/batchprocessor/metadata.yaml#L35-L41

I will update them tomorrow then

vigneshshanmugam: Not opposed to using that, But caps at 9MB - Given OTLP HTTP receiver limits at 20MB, we should allow cover them as well. I am good with adding 5000_000, 10000_000, 15000_000, 20000_000 to the list or using something like above till the 20 MB mark.

Issue #685, opened 2025-07-25
[connector/elasticapm] Allow configuration of custom dimensions for all aggregated metrics.


lahsivjar:
## Motivation

The dimensions are pre-configured for the `signaltometrics` component used within the `elasticapm` connector. This is a bit restrictive to some use-cases where users want to customize the dimensions produced by the aggregations. Customizing dimensions makes the UI for flexible as users can then search the APM UI with the custom aggregated dimensions.

## Implementation

Expose configuration in the `elasticapm` connector to customize the aggregated metrics. Note that the configs offered in the `elasticapm` connector should use the abstractions of `service_summary`, `service_trsansactoin`, `service_destination`, and `transaction` instead of the more flexible options offered by the signal to metrics component.

PR #698, opened 2025-08-06
[connector/elasticapm]Add support for adding custom attrs to aggregated metrics

lahsivjar:
The component exposes configuration for each aggregated APM metrics: `service_summary`, `service_transaction`, `transaction`, and `span_destination`. Each of these metrics can be customized to include an optional list of attributes in addition to the required attributes.

Fixes https://github.com/elastic/opentelemetry-collector-components/issues/685


connector/elasticapmconnector/config.go
```diff
@@ -66,6 +71,26 @@ type AggregationConfig struct {
...
+	// cardinality to be performant.
+	ResourceAttributes []string `mapstructure:"custom_resource_attributes"`
+	// Attributes define a list of resource attributes that will be added to the
+	// aggregated metrics as optional attributes i.e. the attribute will be added
+	// to the aggregated metrics if they are present in the incoming signal,
+	// otherwise, the attribute will be ignored.
+	//
+	// NOTE: any custom attributes should have a bounded and preferably low
+	// cardinality to be performant.
+	Attributes []string `mapstructure:"custom_attributes"`
```
lahsivjar, line 91: [**For reviewers**] (CC: @AlexanderWert @axw ) I have tried to keep the customization for the attributes minimal - the configuration only takes a list of attributes and considers them optional i.e. if the attribute is present in the incoming signal then it is added to the aggregated metric, otherwise it is ignored. Features like default values (which is present in signal to metrics component) are deliberately excluded. Let me know your thoughts on this. 

Note that adding the advanced features is a simple task, so, if we feel like we should have them I will make the changes.

lahsivjar: In addition, if we want to keep the configurations even simpler i.e. just take a list of resource attributes and other attributes and add them to all metrics then that can be done too.

I am not sure how much flexibility we should give to the end user as it might have a negative impact on us in terms of maintenance as well as users breaking the expectations.

lahsivjar: That sounds good to me. @AlexanderWert WDYT? Does the simplified model meet the user's requirement?

lahsivjar: Rethinking, service_summary can be customized using the resource attributes and span, dp, log record attributes doesn't make much sense for service summary. I am going ahead with the suggested simplified implementation for now to unblock this issue.

connector/elasticapmconnector/config.go
```diff
@@ -66,6 +69,26 @@ type AggregationConfig struct {
 	Intervals []time.Duration `mapstructure:"intervals"`
 }
 
+// CustomConfig defines customizations for the metrics produced by the connector.
+type CustomConfig struct {
```
axw, line 73: Can we delete this now? Maybe add a doc comment to the new fields above

connector/elasticapmconnector/connector_test.go
```diff
@@ -103,6 +120,7 @@ func TestConnector_MetricsToMetrics(t *testing.T) {
 
 			allMetrics := nextMetrics.AllMetrics()
 			require.NotEmpty(t, allMetrics)
+			fmt.Println(allMetrics[0].ResourceMetrics().Len())
```
axw, line 123:
```suggestion
```

axw (approved):

axw (approved):
LGTM!

isaacaflores2 (approved):

PR #699, opened 2025-08-07
fix(otelbench): use per iteration ptrace for batching to avoid panics

zmoog:


zmoog:
Already fixed with https://github.com/elastic/opentelemetry-collector-components/pull/686

PR #704, opened 2025-08-11
loadgen: Allow shuffling signal benchmark order

marclop:
Adds a new `-shuffle` flag to randomize the order of signals in the benchmark runs. This is useful for cases where multiple instances of the loadgen are run concurrently, allowing more varied benchmarking.


carsonip (approved):
approving to unblock. One comment: this is not a full shuffle, as the combinations are not shuffled, only signal order and exporter orders separately

vigneshshanmugam (approved):

Issue #708, opened 2025-08-13
[extension/apmconfig] Remove error message when sending ServerToAgent flags


rogercoll:
If the error message field is defined in the ServerToAgent, no other fields should be set.

Context: https://github.com/open-telemetry/opamp-spec/issues/250

PR #710, opened 2025-08-14
fix: error message should not be set with other fields

rogercoll:
The server should just return the flag if the agent is unidentified (service.name not known).

Fixes https://github.com/elastic/opentelemetry-collector-components/issues/708


LikeTheSalad (approved):
Cheers!

LikeTheSalad:
> cc @trentm @LikeTheSalad (please, let me know if there's an APM wide Github group I can mention)

I think @elastic/apm-agent-approvers should do for this PR.

trentm (approved):

PR #716, opened 2025-08-21
[loadgen/otelbench] Add mixed signal benchmark

carsonip:
Add a mixed signal benchmark that is useful for load generation purpose. Default to enabled but can disable by `-mixed=false`


vigneshshanmugam (approved):
Nice

PR #719, opened 2025-08-22
[Otelbench] Set datapath for signals

gizas:
This adds the ability to provide `<signal>-data-path` parameter to your tests in order to provide custom input for your tests
The PR includes also the update of the name of the test in case the data-path is provided

```bash
./loadgen -config="./config_traces.yaml" -concurrency=60 -test.benchtime=1m -test.count=3 -traces-data-path="./testdata/tracesTestFile.jsonl"

BenchmarkOTelbench/traces-otlp-tracesTestFile-60    	    7710	   9093050 ns/op	         0 failed_logs/s	         0 failed_metric_points/s	         0 failed_requests/s	         0 failed_spans/s	         0 logs/s	         0 metric_points/s	       110.0 requests/s	       549.9 spans/s

```

The above is handy when you want to provide custom load for the benchmark test we want to run

Verified that the the custom data path is applied in e2e tests:

Screenshot 2025-08-22 at 3 24 07 PM

### DIFF
--- a/loadgen/cmd/otelbench/.chloggen/setdatapath_signals.yaml
+++ b/loadgen/cmd/otelbench/.chloggen/setdatapath_signals.yaml
@@ -10,7 +10,7 @@ note: Adds <signal>-data-path config option n order to provide custom data input
 component: otelbench
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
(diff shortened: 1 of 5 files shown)


carsonip (approved):
lgtm, can you add a changelog?

PR #725, opened 2025-09-03
[Build loadgen image] Updating platform for build

gizas:
Quick fix to make the image compatible with both platforms

#Relates to https://github.com/elastic/hosted-otel-collector/pull/1313

See https://github.com/elastic/hosted-otel-collector/actions/runs/17427847867/job/49478996691

Error for a linux/arm64 platform:

```bash
--- Otelbench logs from the last 30s ---
exec /ko-app/loadgen: exec format error
```


ChrsMark (commented):
https://github.com/elastic/opentelemetry-collector-components/actions/runs/17435255712/job/49503209693?pr=725#step:6:57

Any idea what is the CI error about and if it's related?

PR #733, opened 2025-09-08
[receiver/elasticapmintake] Add apm to otel histrogram metric conversion

isaacaflores2:
## Overview
Implements histogram conversion  and addresses remaining metric event fields. 

1. Add histogram metric conversion
2. Set metric unit
3. Add notes describing why the interval and doc_count do not need to be set
4. Add summary metric conversion
    * APM does not seem to support summaries, so I do not think an explicit conversion is needed. I discovered this after the fact. Let me know if you think I should just drop the commit for the`populateOTelSummaryDataPoint` implementation
5. Add safety checks on nil `event.Service` and `event.Event` to avoid panic
   * I do not expect this to happen in practice since the apm-data lib decoder performs extra validations, but its something I noticed could happen when adding unit test
## Testing
1. Covered by unit test
2. Ingested data on a local cluster using the `latency_distribution` sample from the test data. Example: 
intakeV2_receiver_histrogram_metrics_latency_distribution_example


receiver/elasticapmintakereceiver/receiver_conversion_test.go
```diff
@@ -0,0 +1,316 @@
...
+	sm = expectedEmptyMetric.ScopeMetrics().AppendEmpty()
+	emptyMetric := sm.Metrics().AppendEmpty()
+	emptyHistrogramDP := emptyMetric.SetEmptyHistogram().DataPoints().AppendEmpty()
+	emptyHistrogramDP.SetTimestamp(pcommon.NewTimestampFromTime(currTime))
+
+	testCases := []struct {
+		name             string
+		receiver         *elasticAPMIntakeReceiver
+		apmMetricEvent   *modelpb.APMEvent
+		resourceMetric   pmetric.ResourceMetrics
```
1pkg, line 240: nit: why to have this temporary resource directly in the testcase structure if we can just use a local variable instead.

receiver/elasticapmintakereceiver/receiver_conversion_test.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+
+	"github.com/elastic/apm-data/model/modelpb"
+)
+
+// Test_Receiver_elasticMetricsToOtelMetrics specific test to cover edge
+// cases when converting from elastic metrics to otel metrics
+func Test_Receiver_elasticMetricsToOtelMetrics(t *testing.T) {
```
1pkg, line 34: Any benefit of using the verbose marshalling in unit test like this one vs using the `tesdata` approvals pattern? If we just initialize the receiver locally and use `tc.receiver.elasticMetricsToOtelMetrics` but swap manual initialization with testdata files it should be less verbose

receiver/elasticapmintakereceiver/receiver.go
```diff
@@ -348,6 +367,111 @@ func (r *elasticAPMIntakeReceiver) populateDataPointCommon(dp otelDataPoint, eve
...
+	// bucket where the boundary is at infinity. This format is intentionally
+	// compatible with the OpenMetrics histogram definition.
+	//
+	// If bucket_counts length is 0 then explicit_bounds length must also be 0,
+	// otherwise the data point is invalid.
+	explicitBounds := dp.ExplicitBounds()
+
+	// explicit bounds are derived from the sample.Histogram.Values, where each value is the upper bound for a bucket.
+	// Except the last bound value which is implied to be +Inf bucket, so it is not set.
+	explicitBounds.FromRaw(apmHistogramValues[:len(apmHistogramValues)-1])
```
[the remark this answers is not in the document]
isaacaflores2: Sorry @1pkg not sure I understand the question. I think you mean are we now converting twice? The answer is yes. We will convert the historgram from apm to otlp here in the receiver. Then the ES exporter converts it back to an "apm" model.

receiver/elasticapmintakereceiver/receiver_conversion_test.go
isaacaflores2: Good call. I moved the test cases over to the existing `TestMetrics` func. I also delete the summary conversion code since the code path is not reachable (its more evident when the test send the http request and go through the initial validation)

PR #769, opened 2025-09-16
improve errorinfo for ratelimiter and auth errors

vigneshshanmugam:
+ Improve the gRPC error info with additional details which can be leveraged in the upstream wrapper component to provide detailed error info to users.

### DIFF
--- a/extension/apikeyauthextension/authenticator.go
+++ b/extension/apikeyauthextension/authenticator.go
@@ -246,7 +246,6 @@ func (a *authenticator) Authenticate(ctx context.Context, headers map[string][]s
 	if err != nil {
 		detailedErr := errorWithDetails(codes.Unauthenticated, err.Error(), map[string]string{
 			"component": "apikeyauthextension",
-			"api_key":   id,
 		})
 		return ctx, detailedErr
 	}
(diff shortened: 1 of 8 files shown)


extension/apikeyauthextension/authenticator.go
```diff
@@ -243,7 +244,11 @@ func (a *authenticator) getCacheKey(ctx context.Context, id string, headers map[
 func (a *authenticator) Authenticate(ctx context.Context, headers map[string][]string) (context.Context, error) {
 	authHeaderValue, id, err := a.parseAuthorizationHeader(headers)
 	if err != nil {
-		return ctx, status.Error(codes.Unauthenticated, err.Error())
+		detailedErr := errorWithDetails(codes.Unauthenticated, err.Error(), map[string]string{
+			"component": "apikeyauthextension",
+			"api_key":   id,
```
constanca-m, line 249: Id is [granted to be empty ](https://github.com/elastic/opentelemetry-collector-components/blob/a1c4a25f5efbcef2d407649402617ca0cc839b8f/extension/apikeyauthextension/authenticator.go#L148)here if there is an error, so there is no point in adding it here

vigneshshanmugam: Good call, ll fix it.

axw (approved):

PR #775, opened 2025-09-18
[receiver/elasticapmintake] Add apm to otel conversion for remaining service attributes

isaacaflores2:
# Overview
 1.  Refactor processing for all signals translate resource level attributes in the parent function (`processBatch`) for readability.
 - We now call `SetElasticSpecificResourceAttributes` for metrics and spans
 2. Add mapping for `Service.Origin` and `Service.Target` apm fields as elastic specific resource attributes
 3. Updates `event.Agent` to be conditionally added only when the field is not empty
    
    This adds additional mapping for metric events
 # Testing
 1. Covered by unit test. Updates to expected files include: 
 - Spans: Now include all elastic specific resource attributes
 - Errors: Now include the recently added  attributes which are mapped from `Service.Origin` and `Service.Target`


1pkg (approved):

rubvs (approved):

gregkalapos (approved):
LGTM. 

I like the change with `setResourceAttributes` - makes it more readable.

PR #779, opened 2025-09-19
[receiver/elasticapmintake] Add mapping for APM labels to OTel resource attributes

isaacaflores2:
# Overview
 1.  Add mappings for labels matching the existing APM behavior. The `event.Labels` are represented as `labels` or `tags` in the apm data decoder model. 
 - https://github.com/elastic/apm-data/blob/main/input/elasticapm/internal/modeldecoder/v2/model.go#L75
 - https://github.com/elastic/apm-data/blob/main/input/elasticapm/internal/modeldecoder/v2/model.go#L433
 - https://github.com/elastic/apm-data/blob/main/input/elasticapm/internal/modeldecoder/v2/model.go#L969
 2. A separate PR will take care of https://github.com/elastic/opentelemetry-collector-components/pull/775 mapping resource level (metadata) attributes for all signals. Once that PR is merged expected test files should change here
    
# Testing
 1. Covered by unit test.

### DIFF
--- a/receiver/elasticapmintakereceiver/receiver_test.go
+++ b/receiver/elasticapmintakereceiver/receiver_test.go
@@ -689,5 +689,9 @@ func runComparisonForMetrics(t *testing.T, inputJsonFileName string, expectedYam
 	}
 	expectedMetrics, err := golden.ReadMetrics(expectedFile)
 	require.NoError(t, err)
-	require.NoError(t, pmetrictest.CompareMetrics(expectedMetrics, actualMetrics, pmetrictest.IgnoreMetricsOrder()))
+	require.NoError(t, pmetrictest.CompareMetrics(expectedMetrics, actualMetrics, pmetrictest.IgnoreMetricsOrder(),
+		// golden.WriteMetrics will sort metrics and resource metrics before writing the golden file
+		// so we need to ignore order when comparing.
+		pmetrictest.IgnoreResourceMetricsOrder(),
+	))
 }
--- a/receiver/elasticapmintakereceiver/testdata/invalid_ids_expected.yaml
+++ b/receiver/elasticapmintakereceiver/testdata/invalid_ids_expected.yaml
@@ -109,6 +109,12 @@ resourceSpans:
         - key: host.os.platform
           value:
             stringValue: darwin
+        - key: labels.tag1
+          value:
+            stringValue: one
+        - key: numeric_labels.tag2
+          value:
+            doubleValue: 2
     scopeSpans:
       - scope: {}
         spans:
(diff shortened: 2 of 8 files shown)


receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToElasticSpecificFields.go
```diff
@@ -116,6 +116,39 @@ func SetElasticSpecificMetadataFields(event *modelpb.APMEvent, attributesMap pco
...
+// - https://github.com/elastic/apm-data/blob/main/input/elasticapm/internal/modeldecoder/v2/model.go#L969
+func setLabels(event *modelpb.APMEvent, attributesMap pcommon.Map) {
+	for key, labelValue := range event.Labels {
+		if key != "" && labelValue != nil {
+			if labelValue.Value != "" {
+				attrKey := "labels." + key
+				attributesMap.PutStr(attrKey, labelValue.Value)
+			}
+		}
+	}
```
rubvs, line 142: I feel the following reads easier, but not a pref:

```
	for key, label := range event.Labels {
		if key != "" && label != nil && label.Value != "" {
			attributesMap.PutStr("labels." + key, label.Value)
		}
	}
```

simitt (changes requested):
This PR only applies the labels from the `metadata`, but the event specific labels are missing. 
Also, only somewhat related to this PR, while reviewing I noticed that the `SetElasticSpecificMetadataFields` functionality is only called for `error` and `log` events, so the metadata mapping (incl. the `labels`) would not be applied to `spans`, `transactions` and `metrics`. If not already a known issue, this will need to be addressed to also be able to cover `labels` from `metadata` for all types.

simitt:
>Also, only somewhat related to this PR, while reviewing I noticed that the SetElasticSpecificMetadataFields functionality is only called for error and log events, so the metadata mapping (incl. the labels) would not be applied to spans, transactions and metrics. If not already a known issue, this will need to be addressed to also be able to cover labels from metadata for all types.

Should have reviewed https://github.com/elastic/opentelemetry-collector-components/pull/775 first; 
will re-review once PR #775 is merged and the expected test outcome files here are updated.

isaacaflores2:
> This PR only applies the labels from the metadata, but the event specific labels are missing.

From my understanding the apm-data library takes care of this when converting the input to the `modelpb.APMEvent`, https://github.com/elastic/apm-data/blob/main/input/elasticapm/internal/modeldecoder/v2/decoder.go#L1623, so we just need to set the event labels as attributes here. I updated the `setLabels` godoc to clarify. 

> Also, only somewhat related to this PR, while reviewing I noticed that the SetElasticSpecificMetadataFields functionality is only called for error and log events, so the metadata mapping (incl. the labels) would not be applied to spans, transactions and metrics. If not already a known issue, this will need to be addressed to also be able to cover labels from metadata for all types.

Should have reviewed https://github.com/elastic/opentelemetry-collector-components/pull/775 first;
will re-review once PR https://github.com/elastic/opentelemetry-collector-components/pull/775 is merged and the expected test outcome files here are updated.

Yes that is correct @simitt . Sorry for the confusion. I was attempting to keep the PRs small and separate. I updated this PR description to clarify, my apologies.

PR #791, opened 2025-09-29
feat(ratelimitprocessor): add class-based limiting

marclop:
Introduces class-based dynamic rate limiting to the `ratelimitprocessor` allowing different rate limiting based on the received client metadata.

The new architecture allows users to define named rate limit classes in the processor configuration, each with its own static rate, burst, and whether or not they support dynamic rate limiting. A new class resolver extension can then be used to map unique keys to class names, enabling dynamic assignment of rate limits based on custom resolution providers.

The processor now supports a clear precedence order: per-key overrides take highest priority, followed by resolved class, default class, and finally the global fallback configuration.

More telemetry has been added to provide visibility into rate limit resolution, dynamic escalations, degraded operation, and resolver failures.

See the updated README for detailed usage instructions.

### DIFF
--- a/processor/ratelimitprocessor/README.md
+++ b/processor/ratelimitprocessor/README.md
@@ -216,7 +216,7 @@ processors:
     strategy: requests
 
     # define the class resolver ID.
-    class_resolver: configmapclassresolverextension
+    class_resolver: configmapclassresolver
     # Define named classes
     classes:
       trial:


processor/ratelimitprocessor/processor_test.go
```diff
@@ -388,7 +388,7 @@ func TestConcurrentRequestsTelemetry(t *testing.T) {
 	}
 
 	metrics := pmetric.NewMetrics()
-	for i := 0; i < numWorkers; i++ {
+	for range numWorkers {
```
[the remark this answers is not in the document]
marclop: meh.

processor/ratelimitprocessor/README.md
```diff
@@ -191,3 +193,113 @@ processors:
...
+        burst: 1000     # burst capacity of 1000
+        static_only: false
+
+      enterprise:
+        rate: 2000      # 2000 requests/second for enterprise
+        burst: 4000     # burst capacity of 4000
+        static_only: false   # allow gradual increase.
+
+    # Default class when resolver returns unknown class
+    default_class: "trial"
```
vigneshshanmugam, line 238: nit: wouldnt it be better to move this inside classes key to act as a catch all? 
```
classes:
   default: ? 

```

processor/ratelimitprocessor/metadata.yaml
```diff
@@ -82,3 +113,12 @@ attributes:
   reason:
     description: rate limit reason
     type: string
+  class:
+    description: rate limit class used
+    type: string
+  source_kind:
+    description: precedence source used to resolve rate
```
vigneshshanmugam, line 120:
```suggestion
    description: precedence source used to resolve rate limiter settings 
```

processor/ratelimitprocessor/gubernator.go
```diff
@@ -116,19 +149,50 @@ func (r *gubernatorRateLimiter) Shutdown(context.Context) error {
...
 	if r.cfg.DynamicRateLimiting.Enabled && !cfg.disableDynamic {
+		attrs := metric.WithAttributeSet(attribute.NewSet(
+			attribute.String("source_kind", string(sourceKind)),
+			attribute.String("class", className),
+		))
 		rate, burst = r.calculateRateAndBurst(ctx, cfg, uniqueKey, hits, now)
-		if rate < 0 {
-			return fmt.Errorf("error calculating dynamic rate limit for unique key %s", uniqueKey)
+		if rate < 0 { // Degraded mode - Gubernator unreachable. Fallback to static rate.
+			r.telemetryBuilder.RatelimitGubernatorDegraded.Add(ctx, 1, attrs)
```
vigneshshanmugam, line 190: nit: Can we have a single metric event with single dimension add these dynamic/static attributes instead of having them separate?

processor/ratelimitprocessor/gubernator.go
```diff
@@ -75,21 +95,34 @@ func newGubernatorDaemonConfig(logger *zap.Logger) (gubernator.DaemonConfig, err
...
+		behavior:         gubernator.Behavior_BATCHING,
+		daemonCfg:        daemonCfg,
+		telemetryBuilder: telemetryBuilder,
+		classResolver:    noopResolver{},
 	}, nil
 }
 
-func (r *gubernatorRateLimiter) Start(ctx context.Context, _ component.Host) (err error) {
+func (r *gubernatorRateLimiter) Start(ctx context.Context, host component.Host) (err error) {
+	if res := r.cfg.ClassResolver; res.String() != "" {
```
vigneshshanmugam, line 115: nit: better to detect for empty extension id in config validation?

vigneshshanmugam (approved):
LGTM, mostly nonblocking comments.

processor/ratelimitprocessor/gubernator.go
marclop: I'll look into that.

processor/ratelimitprocessor/README.md
marclop: That's a bit awkard since `classes` is a `map[string]Class`. That means that you always require a named class `"default"` to be declared, is that what you meant?

processor/ratelimitprocessor/gubernator.go
marclop: Fixed

processor/ratelimitprocessor/gubernator.go
vigneshshanmugam: I meant to explicitly check if resolver is defined, then it cant be empty.

marclop: Added config validation so there's an error if classes are set (they are optional), but no class_resolver is set.

vigneshshanmugam (approved):
LGTM

PR #797, opened 2025-10-05
update error code if ES dial connection error

rubvs:
- If ES is down, we want to return an error unavailable code, along with a user facing error: `retryable server error`.
- If `elasticsearchErr` is **not** ok, it implies the ES instance could not be reached, and therefore does not return a specific ES error, but rather a `code.Internal` error. The error happens on the TCP connection level.

## Manual Test

```shell
# Console 1: Spin up a cluster
> tilt up

# Console 2: Make ES unavailable
> kubectl scale statefulset elasticsearch-es-default --replicas=0

# Console 3: Ensure a retryable error is returned to the user when trying to auth to an unavailable ES.
> TEST_COUNT=10 make run-otelbench mode=local

2025-10-05T15:56:21.572Z  error  internal/queue_sender.go:51  Exporting failed. Dropping data.
{
  "resource": {
    "service.instance.id": "645a4319-43c4-4f8b-b97b-27f5adf6dc0c",
    "service.name": "/ko-app/loadgen",
    "service.version": "0.0.1"
  },
  "otelcol.component.id": "otlp",
  "otelcol.component.kind": "exporter",
  "otelcol.signal": "logs",
  "error": "interrupted due to shutdown: rpc error: code = Unavailable desc = rpc error: code = Unavailable desc = retryable server error \"o1voipUB6v8S06vJAX5J\": an error happened during the HasPrivileges query execution: dial tcp 10.96.217.236:9200: connect: connection refused",
  "dropped_items": 325
}
```


extension/apikeyauthextension/authenticator_test.go
```diff
@@ -77,7 +77,7 @@ func TestAuthenticator(t *testing.T) {
 				},
 				Status: 400,
 			}),
-			expectedErr: `rpc error: code = Unauthenticated desc = error checking privileges for API Key "id": status: 400, failed: [a_type], reason: a_reason`,
+			expectedErr: `rpc error: code = Unavailable desc = error checking privileges for API Key "id": status: 400, failed: [a_type], reason: a_reason`,
```
isaacaflores2, line 80: It looks like the `newCannedErrorHandler` always returns a `types.ElasticsearchError`, do we need a handler that returns no type for this scenario: `error on the TCP connection level`?

extension/apikeyauthextension/authenticator.go
```diff
@@ -286,7 +286,10 @@ func (a *authenticator) Authenticate(ctx context.Context, headers map[string][]s
 				return ctx, status.Error(codes.Unauthenticated, err.Error())
 			}
 		}
-		return ctx, status.Errorf(codes.Unauthenticated, "error checking privileges for API Key %q: %v", id, err)
+		// If no ES error type is found, it implies an error on the TCP connection level.
+		// In this case, we want to return an unavailable code so we have to option of
+		// handling this as a user-defined error later on.
+		return ctx, status.Errorf(codes.Unavailable, "error checking privileges for API Key %q: %v", id, err)
```
isaacaflores2, line 292: It looks like we can reach this return if the error is not of type `ElasticsearchError` or if the `elasticsearchErr.Status` is not unauthorized or forbidden.  Are we supposed to return `codes.Unavailable` for both?

extension/apikeyauthextension/authenticator_test.go
rubvs: How can an ES error be returned if ES is never reached?

extension/apikeyauthextension/authenticator.go
rubvs: I'm not sure if `codes.Unavailable` is the "correct" type to return, but it's the only one that makes sense to me, and is the closest representation of ES being down or unavailable.

extension/apikeyauthextension/authenticator_test.go
isaacaflores2: Sorry my understanding from the below comment is that we want to test the scenario when no ES error type is returned. But this handler always returns an ES error type, so I was wondering if we need to return a different error type
```
// If no ES error type is found, it implies an error on the TCP connection level.
// In this case, we want to return an unavailable code so we have to option of
// handling this as a user-defined error later on.
```

isaacaflores2: My guess is that this test is passing due to the check being false: https://github.com/elastic/opentelemetry-collector-components/pull/797/files#diff-caf0d6c293198c76d9ba0005aa8a22d099b5baec8b5ea4d20d769c4fcbba3e35R285

extension/apikeyauthextension/authenticator.go
isaacaflores2: > My understanding here, is that an ES error cannot be returned if ES is unavailable, since no ES instance is ever "hit" that will return an ES-type-error.

Makes sense. My question was around  the if [statement](https://github.com/elastic/opentelemetry-collector-components/pull/797/files#diff-caf0d6c293198c76d9ba0005aa8a22d099b5baec8b5ea4d20d769c4fcbba3e35R285) I mentioned in the other comment. When ES returns a 400 or 429,  we will end up returning `codes.Unavailable` because it is not an expected status not because it returned no es error type

isaacaflores2: Ah I see now. If we want to change the default response status code then I think we also need to update this comment (or just remove it)
```
// If no ES error type is found, it implies an error on the TCP connection level.
// In this case, we want to return an unavailable code so we have to option of
// handling this as a user-defined error later on.
```

vigneshshanmugam: I don't think this is the correct way to handle it, Two things

1. If the error is of types.ElasticsearchError, then we need to convert that to gRPC specific error and then return that. This could happen when deployment is not found etc. 
2. For all other unknown types, It would be better to return it as Internal Error rather than Unavailable.

extension/apikeyauthextension/authenticator.go
```diff
@@ -281,12 +281,18 @@ func (a *authenticator) Authenticate(ctx context.Context, headers map[string][]s
...
+		case *types.ElasticsearchError:
+			switch elasticsearchErr.Status {
+			case http.StatusUnauthorized, http.StatusForbidden:
 				return ctx, status.Error(codes.Unauthenticated, err.Error())
+			default:
+				return ctx, status.Errorf(codes.Internal, "error checking privileges for API Key %q: %v", id, err)
 			}
+		default:
+			// If no ES error type is found, it implies an error on the TCP connection level.
+			return ctx, errorWithDetails(codes.Unavailable, fmt.Sprintf("retryable server error %q: %v", id, err), nil)
```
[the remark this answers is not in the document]
isaacaflores2: It looks like the metadata is not used in the collector func `GetSanitizedGRPCStatus` but we add it in other places providing the below. I assume this metadata will appear in our logs so I think it would be good to add.  Or at least keep it consistent with other used of `errorWithDetails`
```go
map[string]string{
				"component": "apikeyauthextension",
				"api_key":   id,
}
```

extension/apikeyauthextension/authenticator.go
```diff
@@ -281,12 +281,18 @@ func (a *authenticator) Authenticate(ctx context.Context, headers map[string][]s
...
+		switch elasticsearchErr := err.(type) {
+		case *types.ElasticsearchError:
+			switch elasticsearchErr.Status {
+			case http.StatusUnauthorized, http.StatusForbidden:
 				return ctx, status.Error(codes.Unauthenticated, err.Error())
+			default:
+				return ctx, status.Errorf(codes.Internal, "error checking privileges for API Key %q: %v", id, err)
 			}
+		default:
+			// If no ES error type is found, it implies an error on the TCP connection level.
```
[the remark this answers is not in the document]
rubvs: Updated in https://github.com/elastic/opentelemetry-collector-components/pull/797/commits/973290c9ce88b59c884a2dbb8c0c5a80c73495ec

extension/apikeyauthextension/authenticator.go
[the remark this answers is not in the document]
rubvs: Updated in https://github.com/elastic/opentelemetry-collector-components/pull/797/commits/973290c9ce88b59c884a2dbb8c0c5a80c73495ec

rubvs:
@carsonip any non-elasticsearch error will return a retryable error. So if the proxy - sitting between ingest and ES - returns any error, a retryable error should be returned.

@isaacaflores2 I've added a basic unit test to mock the proxy returning a `502`.

isaacaflores2 (approved):
Thanks for the updates!

rubvs:
How did this merge without passing CI first. Creating a fix quickly.

PR #817, opened 2025-10-08
Elastictrace processor deprecation

gregkalapos:
In order to not break users, but not increasing the maintenance  work, adding back a very minimal `elastictrace` processor which just forwards to `elastictrace` processor. At the same time, marking it as deprecated.


processor/elastictraceprocessor/metadata.yaml
```diff
@@ -0,0 +1,12 @@
...
+scope_name: github.com/elastic/opentelemetry-collector-components/processor/elastictraceprocessor
+
+status:
+  class: processor
+  stability:
+    deprecated: [traces]
+  distributions: []
+  deprecation:
+    traces:
+      date: "2025-10-08"
```
gregkalapos, line 11: This is TBD - it must be a string, otherwise it's rejected. I'd wanted to have a string saying, something like "since Elastic stack release 9.2", but we can't have that, so we need a date.

PR #828, opened 2025-10-09
[beatsauthextension] Add support for kerberos

khushijain21:
This PR adds support for Kerberos authentication

Testing Steps:

1. Start elasticsearch with kerberos enabled.
2. Start collector with following config

```
receivers:
  filelog:
    include_file_name: true
    include:
      - "./otlp-all.json"

extensions:
  beatsauth:
   kerberos: 
     auth_type: "password"
     config_path: "/Users/khushijain/Documents/beats/libbeat/outputs/elasticsearch/testdata/krb5.conf"
     username: "beats"
     password: "testing"
     realm: "elastic"
    
exporters:
  debug:
    verbosity: detailed
  elasticsearch:
   endpoints:
    -  "http://elasticsearch_kerberos.elastic:9203"
   auth:
    authenticator: beatsauth
   mapping:
     mode: raw
   telemetry:
     log_response_body: true    

service:
  telemetry:
    logs:
      level: debug
  extensions: [beatsauth]
  pipelines:
    logs:
      receivers: [filelog]
      processors: []
      exporters: [debug, elasticsearch]
```

3. Observer no error logs 
4. Observer number of documents indexed to `logs-generic-default`.

```
curl -X GET -u admin:testing "http://localhost:9203/logs-generic*/_count?pretty"
{
  "count" : 16,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  }
}
```

### DIFF
--- a/extension/beatsauthextension/go.mod
+++ b/extension/beatsauthextension/go.mod
@@ -1,6 +1,6 @@
 module github.com/elastic/opentelemetry-collector-components/extension/beatsauthextension
 
-go 1.24.9
+go 1.24.0
 
 require (
 	github.com/elastic/beats/v7 v7.17.29


extension/beatsauthextension/go.mod
```diff
@@ -1,9 +1,11 @@
 module github.com/elastic/opentelemetry-collector-components/extension/beatsauthextension
 
-go 1.24.0
+go 1.24.9
 
 require (
+	github.com/elastic/beats/v7 v7.17.29
```
khushijain21: We only import beats for `Kerberos` config creating/validation. I believe we should `kerberos`  package from `beats` to `elastic-agent-libs` instead. And not have to import a large dependency like beats here.

PR for the same https://github.com/elastic/elastic-agent-libs/pull/360

khushijain21:
`beatsauthextension` will be moved to beats. Before that a test collector distro will require to be setup in beats for easy testing. Changes from this PR will be added directly on the beats repo.

khushijain21:
Beatsauth extension has been moved into beats codebase. The related PR to support kerberos hence is also moved there
See https://github.com/elastic/beats/pull/47534

PR #849, opened 2025-10-22
[receiver/elasticapmintake] Fix APM error http semconv translations

isaacaflores2:
# Overview
I was reviewing the SemConv fields in intakeV2ToSemConv.go and noticed a couple issues with error events. 

- `TranslateIntakeV2LogToOTelAttributes` was using the incorrect function `setHTTP` which sets elastic specific http fields. Updated to use  `setHTTPAttributes` (renamed to `translateHttpAttributes`)
- The http attributes were also being applied at the resource level instead of the log level.
- Renamed http and url helper functions in intakeV2ToSemConv.go to avoid future confusion

# Testing
-  Covered by unit test

### DIFF
--- a/receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToSemConv.go
+++ b/receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToSemConv.go
@@ -95,8 +95,8 @@ func translateElasticServiceLanguageToOtelSdkLanguage(language string) string {
 
 // TranslateIntakeV2TransactionToOTelAttributes translates transaction attributes from the Elastic APM model to SemConv attributes
 func TranslateIntakeV2TransactionToOTelAttributes(event *modelpb.APMEvent, attributes pcommon.Map) {
-	setHttpAttributes(event, attributes)
-	setUrlAttributes(event, attributes)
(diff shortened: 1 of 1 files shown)


receiver/elasticapmintakereceiver/testdata/errors_expected.yaml
```diff
@@ -160,102 +160,6 @@ resourceLogs:
         - key: labels.organization_uuid
           value:
             stringValue: 9f0e9d64-c185-4d21-a6f4-4673ed561ec8
-        - key: http.version
-          value:
-            stringValue: "1.1"
-        - key: http.request.headers
```
isaacaflores2: Most of these fields are removed since they are the elastic specific fields that are already set in the log record

receiver/elasticapmintakereceiver/testdata/errors_expected.yaml
```diff
@@ -292,6 +196,36 @@ resourceLogs:
               - key: error.exception.code
                 value:
                   stringValue: "42"
+              - key: http.request.method
```
isaacaflores2: These fields are the otel semconv fields that were missing (`http.*`) or moved over from the resource attributes (`url.*`)

Issue #850, opened 2025-10-22
[processor/ratelimits] Allow dynamic limiters to scale ingest rate to below static limits


lahsivjar:
The current algorithm in `ratelimitprocessor` makes the defined static limits the lower bound on the ingest rate. Due to this, the static limits need to be defined based on the lowest possible ingest rate that should be allowed. Deciding the lowest possible ingest rate is not easy, as we cannot foresee how ES would scale, especially in ECH. If the current rate that the system can sustain goes below the allowed ingest rate, then the whole purpose of the rate limit is defeated.

This issue aims to fix these problems:

- [x] Use static limit as the initial value for rate limiter when dynamic limiter is enabled.
- [x] Allow the multiplier to scale down the ingest rate to ~ 0 if there was some traffic in the previous window and the multiplier is still returned to be less than 1 (meaning multiplier is suggesting decreasing the ingest rate). This would also entail that we put a validation on the window configurator to not return `0` i.e. to not scale ingestion rate to `0`.
- [x] Update the limits configured in the ingest collector.
- [x] Update the [documentation](https://docs.elastic.dev/hotel/architecture/components) and [dynamic rate limiting ADR](https://docs.elastic.dev/hotel/architecture/decisions/dynamic-rate-limiting-gubernator).

PR #851, opened 2025-10-22
Allow dynamic limiter to scale ingest rate below static limit

lahsivjar:
The dynamic limiter can now scale the ingest rate below the static limits if ALL of the following conditions are met:

1. There was some traffic in the previous interval (if there was no traffic then static limit is used irrespective of the mulitplier).
2. The window configurator is indicating scale down (by returning <=1 multiplier).

The above 2 conditions ensure a quick recovery when the ES bottleneck is mitigated. Even though the rate goes close to `0`, once consumer lag recovers the rate would jump to the static rate limit in a max of one window duration.

The tests are a bit flaky right now since `time.Sleep` is used. When go mod version is upgraded to 1.25 we will migrate to `synctest` to fix the flakiness.

Related to: https://github.com/elastic/opentelemetry-collector-components/issues/850

### DIFF
--- a/.github/workflows/bump_otelbench.yaml
+++ b/.github/workflows/bump_otelbench.yaml
@@ -57,7 +57,7 @@ jobs:
         with:
           token: ${{ steps.get_token.outputs.token }}
 
-      - uses: elastic/oblt-actions/git/setup@605dfce047ddfa05cbb32ebb3700d0d3b8a61738
+      - uses: elastic/oblt-actions/git/setup@fbb7a9dbd4bd082c394fd7f998e8764d23a4f736
         with:
           github-token: ${{ steps.get_token.outputs.token }}
 
--- a/distributions/elastic-components/manifest.yaml
+++ b/distributions/elastic-components/manifest.yaml
@@ -6,12 +6,12 @@ dist:
   output_path: ./_build
 
 extensions:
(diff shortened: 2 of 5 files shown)


processor/ratelimitprocessor/gubernator.go
```diff
@@ -407,11 +407,21 @@ func (r *gubernatorRateLimiter) getDynamicLimit(ctx context.Context,
...
+	// static rate (to as low as `1`). As soon as the window multiplier suggests
+	// increasing the ingestion rate, the MaxAllowed will jump to a minimum of
+	// static rate.
 	//
 	// NOTE(marclop) We may want to add a follow-up static ceiling to avoid
 	// unbounded growth.
-	maxAllowed := math.Max(staticRate, previous*windowMultiplier)
+	var maxAllowed float64
+	if previous > 0 && windowMultiplier <= 1 {
+		maxAllowed = max(1, previous*windowMultiplier)
```
lahsivjar: [**For reviewers**] I am putting a lower bound of `1` to avoid edge cases with zero ingestion rate. Let me know if it doesn't make sense OR a lower value should be used instead.

vigneshshanmugam (approved):
LGTM

marclop (approved):

PR #861, opened 2025-10-23
Fix dynamic rate limit escalation metrics

lahsivjar:
Since the component now supports decreasing the rate limit, adapt the `RatelimitDynamicEscalations` metric accordingly.

### DIFF
--- a/processor/ratelimitprocessor/gubernator.go
+++ b/processor/ratelimitprocessor/gubernator.go
@@ -229,20 +229,25 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
 				)...)),
 			)
 			rate, burst = cfg.Rate, cfg.Burst
-		} else if rate > cfg.Rate { // Dynamic escalation occurred
+		} else if rate > cfg.Rate {
 			r.telemetryBuilder.RatelimitDynamicEscalations.Add(ctx, 1,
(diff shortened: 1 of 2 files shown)


processor/ratelimitprocessor/gubernator.go
```diff
@@ -229,20 +229,25 @@ func (r *gubernatorRateLimiter) RateLimit(ctx context.Context, hits int) error {
...
+		} else if rate > cfg.Rate {
 			r.telemetryBuilder.RatelimitDynamicEscalations.Add(ctx, 1,
 				metric.WithAttributeSet(attribute.NewSet(append(attrs,
-					attribute.String("reason", "success"),
+					attribute.String("reason", "escalation"),
 				)...)),
 			)
-		} else { // Dynamic escalation was skipped (dynamic <= static)
+		} else if rate < cfg.Rate {
+			r.telemetryBuilder.RatelimitDynamicEscalations.Add(ctx, 1,
```
lahsivjar: [**For reviewers**] (CC: @marclop) I am not sure about this metric with the latest changes. I was thinking to base this off of the value of the multiplier but I am not sure if that will be helpful either. I have updated the current metric based on if it is higher or lower than the configured static rate but it won’t catch if it is under degradation and starting to decrease from the static rate. We can probably do this by comparing against the previous rate rather than the static rate. Let me know what you think.

PR #862, opened 2025-10-23
Handle spiky load pattern by limiting based on static rate

lahsivjar:
While testing some load patterns, a spiky load pattern would cause full static ingest rate for intervals where the previous interval didn’t have any data even though the multiplier suggests lowering the rate. Another option would be to look at a bigger interval or a weighed average over past data (as mentioned [here](https://github.com/elastic/apm-server/pull/19285/files)), however, the current PR provides a simpler mitigation with the current code.

Related to: https://github.com/elastic/opentelemetry-collector-components/issues/850


vigneshshanmugam (approved):
LGTM, makes sense 👍🏽

PR #867, opened 2025-10-28
[processor/elasticapm] Update processor to handle new supported fields

isaacaflores2:
# Overview
The `receiver/elasticapmintaker` has introduced new support edOTEL SemConv fields. Additionally the `receiver/elasticapmintaker` has also recently  added new APM fields at the resource level. The processor needs to be able to handle these new fields to avoid adding `labels.` prefix  by mistake. 

# Testing
1. Update unit-test a couple of the newly added fields
2. Ran a local collector with the elasticapm processor

### DIFF
--- a/processor/elasticapmprocessor/internal/ecs/ecs_translation.go
+++ b/processor/elasticapmprocessor/internal/ecs/ecs_translation.go
@@ -25,6 +25,48 @@ import (
 	semconv "go.opentelemetry.io/otel/semconv/v1.27.0"
 )
 
+// Supported ECS resource attributes
+const (
+	ecsAttrServiceLanguageName       = "service.language.name"
+	ecsAttrServiceLanguageVersion    = "service.language.version"
+	ecsAttrServiceFrameworkName      = "service.framework.name"
+	ecsAttrServiceFrameworkVersion   = "service.framework.version"
+	ecsAttrServiceRuntimeName        = "service.runtime.name"
+	ecsAttrServiceRuntimeVersion     = "service.runtime.version"
+	ecsAttrServiceOriginID           = "service.origin.id"
+	ecsAttrServiceOriginName         = "service.origin.name"
+	ecsAttrServiceOriginVersion      = "service.origin.version"
+	ecsAttrServiceTargetName         = "service.target.name"
+	ecsAttrServiceTargetType         = "service.target.type"
+	ecsAttrCloudOriginAccountID      = "cloud.origin.account.id"
+	ecsAttrCloudOriginProvider       = "cloud.origin.provider"
+	ecsAttrCloudOriginRegion         = "cloud.origin.region"
+	ecsAttrCloudOriginServiceName    = "cloud.origin.service.name"
+	ecsAttrCloudAccountName          = "cloud.account.name"
+	ecsAttrCloudInstanceID           = "cloud.instance.id"
+	ecsAttrCloudInstanceName         = "cloud.instance.name"
+	ecsAttrCloudMachineType          = "cloud.machine.type"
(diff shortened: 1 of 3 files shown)


processor/elasticapmprocessor/internal/ecs/ecs_translation.go
```diff
@@ -80,15 +91,25 @@ func isSupportedAttribute(attr string) bool {
...
+		"cloud.project.name":
 		return true
 
 	// container.*
 	case string(semconv.ContainerNameKey),
 		string(semconv.ContainerIDKey),
 		string(semconv.ContainerImageNameKey),
-		"container.image.tag",
-		"container.runtime":
+		string(semconv.ContainerImageTagsKey),
```
simitt: this is semantically different as it translates to `container.image.tags` - oversight or why is `container.image.tag` not needed anymore?

processor/elasticapmprocessor/testdata/ecs/elastic_span_db/input.yaml
```diff
@@ -16,6 +16,12 @@ resourceSpans:
...
             doubleValue: 42.0
+        - key: user.id
+          value:
+            stringValue: user-1234
+        - key: service.framework.name
+          value:
+            stringValue: Node
     scopeSpans:
       - scope: {}
         spans:
```
simitt: Are there more fields missing in the test here? Shouldn't there be many more changes in the test output file, looking at the `ecs_translation` changes?

isaacaflores2:
> In `opentelemetry-lib` we have constants for elastic specific attributes [here](https://github.com/elastic/opentelemetry-lib/blob/main/elasticattr/attributes.go). That list is not complete - it's only a few of them currently. It'd be super nice to extend that `attributes.go` file with the elastic specific fields used here and then use those in the `ecs_translation.go` file.
> 
> That'd also improve readability - for each field, we'd immediately see if it's SemConv (prefixed be `semconv[x].`) or elastic speific (prefixed by `elasticattr.`).

Thanks @gregkalapos I think this a good idea. I can also reference `attributes.go` from the elasticapmintake receiver [here](https://github.com/elastic/opentelemetry-collector-components/blob/main/receiver/elasticapmintakereceiver/internal/attributes.go) or add these fields to the `opentelemery-lib` you shared. Before I make these changes I would like to confirm we are okay with this approach. 

I was thinking this approach of also adding ECS attributes might not be sustainable, since there would be a lot of fields (I think it will most likely require to list all of the metadata fields defined by the apm-data lib). The only alternative I can think of is to skip this flow where we add the `labels*` prefix when we know we have received APM events from the `elasticapmintakereceiver`. @gregkalapos @simitt please let me know what you think. If we are ensure if we want to add ECS attributes here then I cant drop those changes so this PR only adds OTEL sem conv attributes for now

processor/elasticapmprocessor/internal/ecs/ecs_translation.go
isaacaflores2: That was an oversight on my part. I see`container.image.tag` is used in the apm-data library and `semconv.ContainerImageTagsKey` is used in the `elasticapmintakereceiver`. I updated to support both for now

simitt: thanks!

gregkalapos (approved):
> I have updated ecs_translation.go to use constants for the ecs attribute names and updated test cases. Once we have finalized all the ecs resource attributes, I can add them to the open-telemetry-lib and follow up with a PR to import those constants here. How does that sound?

Sounds ok to me. 

PR is also ok to merge from my side.

Issue #847, opened 2025-10-21
[processor/ratelimit] Simplify `overrides` to not depend on the internal unique key


lahsivjar:
Currently, the `overrides` depend on the unique key algorithm, which joins the provided metadata keys (configured by the user) and values (retrieved from the context). This leaks the internal implementation details as the user must know how the join happens in order to be able to configure the overrides for a project/target ID. Another issue with this approach is that the older key is made redundant on adding new metadata keys.

The fix proposal is to implement overrides using a key-value based matching pattern instead of a derived key. For example:

```yaml
overrides:
  - matches:
       x-elastic-target-id: <pid1>
       x-elastic-target-type: <typ1>
    burst: 10_000
    rate: 100_000
```

Overrides will be applied from top to bottom, and the first match will be prioritised. All the key value pairs in the `matches` MUST match the incoming data for the match to be success (AND operation).

PR #871, opened 2025-10-28
Refactor overrides to allow simpler matching expressions

lahsivjar:
The PR simplifies the previous override by allowing a map of key-value pairs instead of an encoded string. The key-value pairs for a specific override MUST be a subset of the client metadata for the override to be applied. The first override that matches is chosen.

The PR also maintains backward compatibility with the previous format of overrides. This is temporary and will be removed once the PR is merged and released to production, followed by the existing overrides being migrated to the new format.

Fixes: https://github.com/elastic/opentelemetry-collector-components/issues/847


simitt (approved):

lahsivjar:
@gizas Yes but not immediately. We have to wait for the changes to go to prod and then update all the current rate limits too (the current PR is backward compatible). I have created an internal issue for this: https://github.com/elastic/hosted-otel-collector/issues/1678

PR #874, opened 2025-10-31
(ratelimitprocessor): measure ingested uncompressed bytes before limits

vigneshshanmugam:
+ Measure the uncompressed bytes in addition to the histogram buckets as both can be really useful to plot the rate of data before throttling. This is only measured when the strategy is set to bytes.

### DIFF
--- a/processor/ratelimitprocessor/metadata.yaml
+++ b/processor/ratelimitprocessor/metadata.yaml
@@ -30,6 +30,13 @@ telemetry:
         value_type: int
         monotonic: true
       attributes: ["class", "source_kind", "result"]
+    ratelimit.request.uncompressed.size:
+      enabled: true
+      description: Size of the uncompressed request in bytes. Only available if strategy is set to bytes.
+      unit: "{bytes}"
+      sum:
+        value_type: int
+        monotonic: true
     ratelimit.request_duration:
       enabled: true
       description: Time(in seconds) taken to process a rate limit request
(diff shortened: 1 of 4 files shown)


processor/ratelimitprocessor/metadata.yaml
```diff
@@ -30,6 +30,13 @@ telemetry:
         value_type: int
         monotonic: true
       attributes: ["class", "source_kind", "result"]
+    ratelimit.request.uncompressed.size:
```
axw:
```suggestion
    ratelimit.request_uncompressed_size:
```

So it's closer to `ratelimit.request_size`?

processor/ratelimitprocessor/processor.go
```diff
@@ -260,6 +260,7 @@ func recordRequestSize(ctx context.Context, tb *metadata.TelemetryBuilder, strat
 	if tb != nil && strategy == StrategyRateLimitBytes {
 		attrsCommon := getAttrsFromContext(ctx, metadataKeys)
 		tb.RatelimitRequestSize.Record(ctx, int64(hits), metric.WithAttributes(attrsCommon...))
+		tb.RatelimitRequestUncompressedSize.Add(ctx, int64(hits), metric.WithAttributes(attrsCommon...))
```
marclop: Aren't we already measuring this with the metric above? What am I missing?

PR #882, opened 2025-11-03
profilingmetricsconnector: Add kernel stacktrace classification

christos68k:
This PR adds kernel stacktrace classification logic for the following categories:

1. network/tcp/read
2. network/tcp/write
3. network/udp/read
4. network/udp/write
5. ipc/read
6. ipc/write
7. network/other/read
8. network/other/write
9. disk/read
10. disk/write
11. memory
12. sycnhronization

I also merged @florian's #834 PR to execute inside the same, single-pass, frame iteration logic.

**TODO:**

- ~More tests incl. multiple timestamps~
- ~Merge the logic from @florianl syscalls PR (#834)~

### DIFF
--- a/connector/profilingmetricsconnector/frame.go
+++ b/connector/profilingmetricsconnector/frame.go
@@ -111,6 +111,9 @@ var (
 	// stacktrace which would be incorrect. The elements are ordered in decreasing
 	// priority: if there are multiple classifications, the element with the lower
 	// slice index always wins.
+	//
+	// TODO: CNIs like Cilium may introduce complications and the following list
+	// may need to be updated.
 	classes = []class{
 		{name: "network/tcp/read",
 			rx: regexp.MustCompile(`^(?:tcp_recvmsg|tcp_read|tcp_rcv|tcp_v4_rcv|tcp_v6_rcv)`)},


connector/profilingmetricsconnector/frame.go
```diff
@@ -19,7 +19,9 @@ package profilingmetricsconnector // import "github.com/elastic/opentelemetry-co
 
 import (
 	"fmt"
+	"log/slog"
```
rogercoll: We should not use `slog` but the logger provided to the connector instance. The code will need a bit of refactor to replace it, but the logger can be propagated during the `createProfilesToMetrics` constructor as part of the `connector.Settings.Logger` parameter.  Example: https://github.com/elastic/opentelemetry-collector-components/blob/main/processor/elasticapmprocessor/factory.go#L57C49-L57C55

This will help have unified logs (with the corresponding resource attributes) when run with a collector.

connector/profilingmetricsconnector/frame.go
```diff
@@ -123,67 +190,161 @@ func fetchFrameInfo(dictionary pprofile.ProfilesDictionary,
...
-	if sm := rx.FindStringSubmatch(fi.filename); sm != nil {
-		nativeCounts[sm[1]]++
+	if className == "" && syscall == "" {
+		// No kernel category / syscall match
+		counts[metricKernel] += multiplier
 	} else {
-		counts[metric]++
+		kernelCounts[kernelStackAttr{
+			class:   className,
+			syscall: syscall}] += multiplier
```
rogercoll: If found this a bit misleading, shouldn't `counts[metricKernel]` always be incremented? Even if we did not match the category/syscall, we are sure it is from the kernel

rogercoll: Thanks for the clarifications, it helped :+1:

connector/profilingmetricsconnector/frame.go
```diff
@@ -70,44 +96,88 @@ var (
...
+		{name: "network/udp/read",
+			rx: regexp.MustCompile(`^(?:udp_recvmsg|udp_read|udp_rcv|udp_v4_rcv|udp_v6_rcv)`)},
+		{name: "network/udp/write",
+			rx: regexp.MustCompile(`^(?:udp_sendmsg|udp_write|udp_push|udp_sendpage)`)},
+		{name: "ipc/read",
+			rx: regexp.MustCompile(`^(?:pipe_read|pipe_read_iter|eventfd_read` +
+				`|unix_stream_read_generic|unix_stream_recvmsg)`)},
+		{name: "ipc/write",
+			rx: regexp.MustCompile(`^(?:pipe_write|pipe_write_iter|eventfd_write` +
+				`|unix_stream_write_generic|unix_stream_sendmsg)`)},
```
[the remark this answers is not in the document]
christos68k: There are multiple options for visualization:

1. Hierarchical (e.g. treemap with 3 levels of nesting or pie chart with same)
2. Different panels that simulate the above. Example: The first panel would be a treemap or pie chart that shows the root level (kernel0), then a second panel to show a level deeper (kernel1) and finally a third panel for the third level.

I think the first option is preferable as it shows everything in one panel and the relationships are immediately obvious. 

There's also visualization over time, e.g. stacked bargraph with flattened categories (e.g. each dimension would be the sum of metrics with all three attributes).


Regarding attributes, I can rename them from kernel_0/kernel_1/kernel_2 to kernel_area, kernel_proto, kernel_io if that's more clear but maybe you mean something else?

connector/profilingmetricsconnector/connector.go
```diff
@@ -23,6 +23,8 @@ import (
 	"fmt"
 	"regexp"
 
+	"go.uber.org/zap"
```
florianl: Can we use `log/slog` over `zap`?

connector/profilingmetricsconnector/frame.go
```diff
@@ -120,70 +195,171 @@ func fetchFrameInfo(dictionary pprofile.ProfilesDictionary,
...
+	counts map[metric]int64,
+	nativeCounts map[attrInfo]int64,
+	multiplier int64) {
+	ft := fi.typ
+
+	// We have a distinct metricUser (even if there's no specific frame type for it)
+	// so that we can calculate the total number of stacktraces for any given time
+	// interval: metricUser.count + metricKernel.count
+	counts[metricUser] += multiplier
+	metric := allowedFrameTypes[ft]
```
florianl: Maybe we should use a helper function on allowedFrameTypes that returns something for frame types that are not listed in allowedFrameTypes.

connector/profilingmetricsconnector/connector.go
```diff
@@ -55,6 +57,7 @@ type profilesToMetricsConnector struct {
 	nextConsumer consumer.Metrics
 	config       *Config
 	aggregations []aggregation
+	logger       *zap.Logger
```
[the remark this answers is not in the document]
christos68k: See https://github.com/elastic/opentelemetry-collector-components/pull/882#discussion_r2493570395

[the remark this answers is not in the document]
rogercoll: This will be aligned with internal collector's telemetry configuration as being provided to the components' factory https://github.com/elastic/opentelemetry-collector-components/pull/882/files#diff-308c87590e839f13259183e5aceffd54525c6bae0f8fa0092f650a28f7dcebc2R52

connector/profilingmetricsconnector/frame.go
```diff
@@ -35,10 +37,37 @@ type metric struct {
...
+	shlib string
+}
+
+const (
+	nativeLibraryAttrName = "shlib_name"
+	syscallAttrName       = "syscall_name"
+	kernelClassAttrPrefix = "kernel_"
+)
+
+var kernelClassAttrNames = [...]string{
```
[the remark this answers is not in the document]
christos68k: It does benefit from a fixed size array as it is a narrower, more constrained type. Why use a slice here when we don't need the extra flexibility?

connector/profilingmetricsconnector/frame.go
christos68k: This would be complicating the code more I think, isn't it much simpler to just validate once by checking if value exists in the array? 

We don't use any other value so the only reason we check the array is to determine if we should error / abort processing.

PR #920, opened 2025-12-01
apikeyauthextension: support dynamic resources

axw:
Add support for checking resources based on client metadata. This can be used for checking that a client has permissions for an application-specific resource that they specify as a separate header.

This was created using Cursor.


extension/apikeyauthextension/config.go
```diff
@@ -99,3 +117,57 @@ func (cfg *CacheConfig) Validate() error {
...
+
+// Validate validates the DynamicResource configuration.
+func (dr *DynamicResource) Validate() error {
+	if dr.Metadata == "" {
+		return fmt.Errorf("metadata must be non-empty")
+	}
+	format := dr.Format
+	if format == "" {
+		format = "%s"
+	}
```
constanca-m: A bit odd to define the default on `validate`. Maybe change this to the `createDefault` config function?

extension/apikeyauthextension/authenticator_test.go
```diff
@@ -477,6 +477,179 @@ func TestAuthenticator_AuthorizationHeader(t *testing.T) {
...
+	}}
+	config.Cache.KeyMetadata = []string{"X-Resource-Name"}
+	authenticator := newTestAuthenticator(t, newMockElasticsearch(t, newCannedHasPrivilegesHandler(successfulResponse)), config)
+
+	// Multiple metadata values should return an error
+	withMultipleValues := client.NewContext(context.Background(), client.Info{
+		Metadata: client.NewMetadata(map[string][]string{
+			"X-Resource-Name": {"resource1", "resource2"},
+		}),
+	})
```
constanca-m: I need help to understand this part. Why is this an issue?

axw: Are you asking why we expect an error when there are multiple values for a header? Because otherwise how would we format the resource?

extension/apikeyauthextension/config.go
axw: I've added an Unmarshal method which sets the default now.

extension/apikeyauthextension/authenticator_test.go
constanca-m: Yes. From my understanding, we can have multiple resources, right? For example, in our elastic config in TF [we use a wildcard](https://github.com/elastic/obs-ds-managed-services-testing/blob/931d0ada48fa35f28f4524c7c655c99c98b4946d/terraform/elastic/serverless-project/main.tf#L28). So why would it be an issue if we define multiple here?

> Because otherwise how would we format the resource?

I would expect the formatter applies to all elements of the list

axw:
govulncheck error is unrelated, will be sorted when "oldstable" is updated in github CI

PR #924, opened 2025-12-02
[processor/elasticapm] Add `host.ip` enrichment using the client info address

isaacaflores2:
# Overview
- Add the ability to set the `host.ip` resource attribute value using the client address from `ctx.Client.Info.Addr` 
  * The enrichment is controlled via the `host_ip_enabled` config
- I opted to add the enrich here versus in the opentelemetry-lib [enricher](https://github.com/elastic/opentelemetry-lib/blob/main/enrichments/enricher.go) since the enrich [methods](https://github.com/elastic/opentelemetry-lib/blob/main/enrichments/enricher.go#L43) do not accept a ctx. 

~Edit: I forgot to add that I am working on an upstream issue https://github.com/open-telemetry/opentelemetry-collector/issues/14186 that will provide more flexibility on how the ctx.Client.Info.Addr`  is set.~

Edit: The  `ctx.Client.Info.Addr` manipulation will now be implemented as a middleware extension: https://github.com/elastic/opentelemetry-collector-components/pull/930 https://github.com/elastic/opentelemetry-collector-components/pull/936

## Misc updates
- Add missing log and metric test cases to `TestProcessorECS`


vigneshshanmugam:
> Apologies I did not see your comment earlier. APM-Server relies on extracting the host.ip from [headers](https://github.com/elastic/apm-server/blob/main/internal/netutil/netutil.go#L35) and the peer [address](https://github.com/elastic/apm-server/blob/main/internal/beater/interceptors/metadata.go#L45). So the work in the upstream https://github.com/open-telemetry/opentelemetry-collector/issues/14186 will provide the flexibility we need

Sounds good, thanks for the reference.

processor/elasticapmprocessor/testdata/ecs/elastic_log/output.yaml
```diff
@@ -0,0 +1,325 @@
...
+            boolValue: false
+        - key: labels.faas_trigger_request_id
+          value:
+            stringValue: req-123
+        - key: faas.execution
+          value:
+            stringValue: exec-456
+        - key: data_stream.dataset
+          value:
+            stringValue: apm.app.test_service
```
isaacaflores2: [for reviewer] note: the service specific data_stream.dataset. This is expected and consolidates the test cases under the delete `elastic_apm`

processor/elasticapmprocessor/testdata/ecs/elastic_metric/output.yaml
```diff
@@ -0,0 +1,327 @@
...
+            stringValue: v1.0.0,latest
+        - key: container.name
+          value:
+            stringValue: my-container
+        - key: container.runtime
+          value:
+            stringValue: docker
+        - key: data_stream.dataset
+          value:
+            stringValue: apm.app.test_service
```
isaacaflores2: [for reviewer] note: the service specific data_stream.dataset. This is expected and consolidates the test cases under the delete `elastic_apm`

simitt (commented):
Apologies for the late review. IMO setting the `host.ip` from context information is a perfect example for an enrichment, and given the current library architecture should live in the `opentelemetry-lib`. Not having the `context` in the signature is a problem, the only way I see around it is to introduce a new `enricher.EnrichTracesWithCtx(ctx context.Context, pt ptrace.Traces)` method (and similar for logs and metrics).
I am not sure it would be needed outside the `apmprocessor`, so it might just be a cosmetic thing whether we keep the logic here or in opentelemetry-lib, I'd find it slightly cleaner in the library. 
What do others think? @isaacaflores2 @axw @vigneshshanmugam @inge4pres

processor/elasticapmprocessor/testdata/elastic_hostname/spanss_output.yaml
```diff
@@ -0,0 +1,63 @@
+resourceSpans:
```
inge4pres: is the file rename intended?
it's moving from `spans_output.yaml` to `spanss_output.yaml`

processor/elasticapmprocessor/testdata/ecs/elastic_metric/output.yaml
inge4pres: Great switch here, cleaner testing 👍🏼

inge4pres (approved):
LGTM, a non blocking question on the file rename, is it intentional?

PR #932, opened 2025-12-09
Introduce dynamic routing connector

lahsivjar:
Dynamic routing connector allows selecting pipelines based on the configured set of metadata keys. For example: route to different batching pipelines based on observed connection count.

Related to: https://github.com/elastic/hosted-otel-collector/issues/1349

### DIFF
--- a/connector/dynamicroutingconnector/README.md
+++ b/connector/dynamicroutingconnector/README.md
@@ -43,9 +43,9 @@ The Dynamic Routing Connector fills this gap by providing **adaptive, cardinalit
 
 The connector uses the following approach:
 
-1. **Metadata Extraction**: For each incoming telemetry signal, the connector extracts metadata from the client context using the configured `primary_metadata_keys` and `metadata_keys`. The `primary_metadata_keys` are used to create a composite key that partitions cardinality estimates (e.g., per tenant, per tenant+service, per region+environment). Multiple keys can be specified to create composite partitions. The `metadata_keys` define what unique combinations are being counted for each composite primary key value.
+1. **Metadata Extraction**: For each incoming telemetry signal, the connector extracts metadata from the client context using the configured `routing_keys.partition_by` and `routing_keys.measure_by`. The `partition_by` keys are used to create a composite key that partitions cardinality estimates (e.g., per tenant, per tenant+type, per region+environment). Multiple keys can be specified to create composite partitions. The `measure_by` keys define what unique combinations are being counted for each composite partition key value.
 
-2. **Cardinality Estimation**: The connector uses the HyperLogLog algorithm to estimate the number of unique combinations of the configured `metadata_keys` for each composite value of the `primary_metadata_keys`. Composite keys are constructed by concatenating values from all specified primary metadata keys, separated by colons (`:`) for multiple values of the same key and semicolons (`;`) for different keys.
+2. **Cardinality Estimation**: The connector uses the HyperLogLog algorithm to estimate the number of unique combinations of the configured `measure_by` keys for each composite value of the `partition_by` keys. Composite keys are constructed by concatenating values from all specified partition keys, separated by colons (`:`) for multiple values of the same key and semicolons (`;`) for different keys.
 
 3. **Threshold-Based Routing**: Based on the estimated cardinality for each primary key value, the connector routes data to different pipelines defined by threshold boundaries. For example:
    - Low cardinality (0-10 unique combinations) → Pipeline A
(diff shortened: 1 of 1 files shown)


rogercoll (commented):
Could you add this component to the `versions.yaml` file? :pray: (created [this issue](https://github.com/elastic/opentelemetry-collector-components/blob/main/versions.yaml) to add a CI check)

connector/dynamicroutingconnector/config.go
```diff
@@ -0,0 +1,54 @@
...
+	"time"
+
+	"go.opentelemetry.io/collector/pipeline"
+)
+
+type Config struct {
+	DefaultPipelines   []pipeline.ID   `mapstructure:"default_pipelines"`
+	EvaluationInterval time.Duration   `mapstructure:"evalaution_interval"`
+	Pipelines          [][]pipeline.ID `mapstructure:"pipelines"`
+	Thresholds         []int           `mapstructure:"thresholds"`
```
vigneshshanmugam: Threshold directly linking to pipeline id also seems like a vialble option. ["10": "traces/low"]

connector/dynamicroutingconnector/config.go
```diff
@@ -0,0 +1,54 @@
...
+	"errors"
+	"sort"
+	"time"
+
+	"go.opentelemetry.io/collector/pipeline"
+)
+
+type Config struct {
+	DefaultPipelines   []pipeline.ID   `mapstructure:"default_pipelines"`
+	EvaluationInterval time.Duration   `mapstructure:"evalaution_interval"`
```
axw:
```suggestion
	EvaluationInterval time.Duration   `mapstructure:"evaluation_interval"`
```

connector/dynamicroutingconnector/logs.go
```diff
@@ -0,0 +1,75 @@
...
+func (c *logsConnector) Start(ctx context.Context, host component.Host) error {
+	return c.router.Start(ctx, host)
+}
+
+func (c *logsConnector) Shutdown(ctx context.Context) error {
+	return c.router.Shutdown(ctx)
+}
+
+func (c *logsConnector) Capabilities() consumer.Capabilities {
+	return consumer.Capabilities{MutatesData: true}
```
axw: Does it? I don't see any mutation, and wouldn't expect any

connector/dynamicroutingconnector/README.md
```diff
@@ -0,0 +1,288 @@
...
+    thresholds: [10, 50, 200]
+    pipelines:
+      # 0-10 unique connections: Small batches, frequent flush
+      - ["traces/small_batch"]
+      # 11-50 unique connections: Medium batches
+      - ["traces/medium_batch"]
+      # 51-200 unique connections: Large batches
+      - ["traces/large_batch"]
+      # 200+ unique connections: Very large batches, aggressive batching
+      - ["traces/xlarge_batch"]
```
axw:
```suggestion
    dynamic_pipelines:
      # 0-10 unique connections: Small batches, frequent flush
      - max_count: 10
        pipelines: ["traces/small_batch"]
      # 11-50 unique connections: Medium batches
      - max_count: 50
        pipelines: ["traces/medium_batch"]
      # 51-200 unique connections: Large batches
      - max_count: 200
        pipelines: ["traces/large_batch"]
      # 200+ unique connections: Very large batches, aggressive batching
      - pipelines: ["traces/xlarge_batch"]
```

WDYT? We can have validation that there's exactly one entry without "max" defined, which will be used for the +Inf case.

connector/dynamicroutingconnector/router.go
```diff
@@ -0,0 +1,263 @@
...
+	if !ok {
+		return r.defaultConsumer
+	}
+	cardinality := hll.Estimate()
+	r.logger.Debug(
+		"returning non-default consumer",
+		zap.String("primary_key", pk),
+		zap.Uint64("cardinality", cardinality),
+	)
+	return r.consumers[sort.SearchInts(r.thresholds, int(cardinality))]
```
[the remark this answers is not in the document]
lahsivjar: > should we pre-calculate the cardinality estimate when updating decisions, so each call to getNextConsumer doesn't need to repeat it?

As a followup, I was thinking of updating the calculation to also account for the data collected in the current window. For example, if the current window allows batching then use that and don't look into the previous one. If it doesn't allow batching then peek into the previous window.

But, I think it makes sense to pre-calculate the cardinality estimate of the previous window and save cardinality -> counsumer mapping. I will update it.

connector/dynamicroutingconnector/config.go
```diff
@@ -0,0 +1,54 @@
...
+	Pipelines          [][]pipeline.ID `mapstructure:"pipelines"`
```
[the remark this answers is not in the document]
lahsivjar: Done!

connector/dynamicroutingconnector/README.md
lahsivjar: Like it! updated the code.

connector/dynamicroutingconnector/metrics.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (c *metricsConnector) Start(ctx context.Context, host component.Host) error {
+	return c.router.Start(ctx, host)
+}
+
+func (c *metricsConnector) Shutdown(ctx context.Context) error {
+	return c.router.Shutdown(ctx)
+}
+
+func (c *metricsConnector) Capabilities() consumer.Capabilities {
```
vigneshshanmugam: Ques - Is this required? Shouldn't it default to false anyway?

connector/dynamicroutingconnector/router.go
```diff
@@ -0,0 +1,263 @@
...
+// updateDecisions updates the current cache to be used for decisions, discarding
+// the previous decision map.
+func (r *router[C]) updateDecisions() {
+	r.dmu.Lock()
+	defer r.dmu.Unlock()
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.decision = r.m
+	r.m = make(map[string]*hyperloglog.Sketch)
```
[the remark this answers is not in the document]
axw: OK, I think it's fine - not going to be replacing the map super frequently anyway.

axw (approved):
LGTM. I like @vigneshshanmugam's suggestion for the config to be in terms of partition keys and counting keys - seems clearer to me. I'm good with it either way though.

PR #938, opened 2025-12-12
processor(elasticapm): route error to data stream

inge4pres:
### Reason for this PR

We should be ingesting data with the correct data_stream.datasetfied set when intake/v2protocol from APM is processed after the receiver translated it to OTle.

### Details

* set `data_stream.dataset` attribute to the LogRecord attributes

### DIFF
--- a/processor/elasticapmprocessor/internal/routing/data_stream.go
+++ b/processor/elasticapmprocessor/internal/routing/data_stream.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/elastic/opentelemetry-lib/elasticattr"
 	"go.opentelemetry.io/collector/pdata/pcommon"
-	semconv "go.opentelemetry.io/otel/semconv/v1.27.0"
+	semconv "go.opentelemetry.io/otel/semconv/v1.38.0"
 )
 
 // DataStreamType tracks the text associated with a data stream type.
--- a/processor/elasticapmprocessor/internal/routing/data_stream_test.go
+++ b/processor/elasticapmprocessor/internal/routing/data_stream_test.go
@@ -23,7 +23,7 @@ import (
 	"github.com/elastic/opentelemetry-collector-components/processor/elasticapmprocessor/internal/routing"
 	"github.com/stretchr/testify/assert"
 	"go.opentelemetry.io/collector/pdata/pcommon"
-	semconv "go.opentelemetry.io/otel/semconv/v1.27.0"
+	semconv "go.opentelemetry.io/otel/semconv/v1.38.0"
 )
 
 func TestDataStremaEncoderDefault(t *testing.T) {


processor/elasticapmprocessor/internal/routing/data_stream.go
```diff
@@ -92,3 +93,21 @@ func replaceReservedRune(r rune) rune {
 	}
 	return r
 }
+
+// IsErrorLog checks if a log record represents an APM error event.
+// The processor.event attribute is set by the elasticapmintakereceiver
+// when converting APM error events to OTLP logs.
```
simitt: How would this work for incoming data through the `otlp` API on the`.apm` route

inge4pres:
I looked at ES exporter code and found that we can use record-level attributes to route documents to the correct data stream.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/f370181737737b8773db8fae105f3603beb8ee45/exporter/elasticsearchexporter/data_stream_router.go#L160-L182

Pay attention to how the function is called
```
dataset, datasetExists := getFromAttributes(elasticsearch.DataStreamDataset, defaultDataStreamDataset, recordAttr, scopeAttr, resourceAttr)
```

This is where the attributes are searched to find a corresponding field

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/f370181737737b8773db8fae105f3603beb8ee45/exporter/elasticsearchexporter/attribute.go#L19-L26


Since the order of processing is inner-to-outer and stopping on the first match, we confirmed that we can set record-level `data_stream.dataset` attributes to span events and error logs into a wider resource that is marked with a different attribute 👍🏼 
This means we don't have to extract, but simply iterate and mark with the appropriate attribute using the check conditions.

inge4pres:
Last commit add support for processing span events.

processor/elasticapmprocessor/processor.go
```diff
@@ -76,6 +76,23 @@ func (p *TraceProcessor) ConsumeTraces(ctx context.Context, td ptrace.Traces) er
...
+			// Iterate through spans to find errors in span events
+			scopeSpans := resourceSpan.ScopeSpans()
+			for j := 0; j < scopeSpans.Len(); j++ {
+				spans := scopeSpans.At(j).Spans()
+				for k := 0; k < spans.Len(); k++ {
+					span := spans.At(k)
+					events := span.Events()
+					for l := 0; l < events.Len(); l++ {
+						event := events.At(l)
+						if routing.IsErrorEvent(event.Attributes()) {
```
inge4pres: Doubt: we are saying here that only spans in ECS mode are processed as error events.
Is this correct?

simitt: Yes, should only be applied in `ecs` mode. In `otel native`, we comply with otel conventions and do not store error events in dedicated datastreams, see https://github.com/elastic/elasticsearch/tree/main/x-pack/plugin/otel-data/src/main/resources/index-templates.

processor/elasticapmprocessor/internal/routing/data_stream_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/elastic/opentelemetry-collector-components/processor/elasticapmprocessor/internal/routing"
 	"github.com/stretchr/testify/assert"
 	"go.opentelemetry.io/collector/pdata/pcommon"
+	semconv "go.opentelemetry.io/otel/semconv/v1.27.0"
```
dmathieu:
```suggestion
	semconv "go.opentelemetry.io/otel/semconv/v1.38.0"
```

PR #942, opened 2025-12-18
Shutdown shared component after the last returned instance is shut down

lahsivjar:
Full explanation available in https://github.com/elastic/hosted-otel-collector/issues/1967#issuecomment-3668678605

Fixes: https://github.com/elastic/hosted-otel-collector/issues/1967

### DIFF
--- a/internal/sharedcomponent/sharedcomponent.go
+++ b/internal/sharedcomponent/sharedcomponent.go
@@ -174,6 +174,7 @@ func (h *hostWrapper) addSource(s componentstatus.Reporter) {
 // Note that the logic for ref counting relies on each component's shutdown being called,
 // in any order, during graceful shutdown.
 func (c *Component[V]) Shutdown(ctx context.Context) error {
+	var err error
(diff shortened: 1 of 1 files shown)


internal/sharedcomponent/sharedcomponent.go
```diff
@@ -163,26 +167,35 @@ func (h *hostWrapper) addSource(s componentstatus.Reporter) {
...
-		}
-		err = c.component.Shutdown(ctx)
-		if c.hostWrapper != nil {
-			if err != nil {
-				c.hostWrapper.Report(componentstatus.NewPermanentErrorEvent(err))
-			} else {
-				c.hostWrapper.Report(componentstatus.NewEvent(componentstatus.StatusStopped))
+	if c.refCounter.Add(-1) == 0 {
+		var err error
+		c.stopOnce.Do(func() {
```
axw: Do we need the Once anymore? Seems that was a bandaid

axw (approved):

PR #943, opened 2025-12-18
[processor/elasticapm] Update elastic/opentelemetry-lib to v0.28.0

isaacaflores2:
# Overview
- Bumping the opentelemtry-lib version to include changes from: https://github.com/elastic/opentelemetry-lib/pull/251
- Update test input files to include a spanId and traceId

Note: The opentelemetry-lib requires a new version of otel libraries so not sure if I should wait for https://github.com/elastic/opentelemetry-collector-components/pull/941 to be merged first


isaacaflores2:
> @isaacaflores2 looks like the version bumps are merged, this should be unblocked

@axw I may need to make some updates to the opentelemetry-lib and release a different version. This is in regards to our discussion on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45014 to keep the code which sets the span id and span name

PR #950, opened 2026-01-02
feat: add flush_interval for metrics downsample

rogercoll:
In-memory aggregation of delta metrics can now be enabled by defining a flush_interval greater than 0. The implementation wraps the downstream metrics consumer to intercept and accumulate values over the defined interval. This works because all current metrics are **counters**; thus, summing their values preserves the total event count, effectively downsampling the temporal resolution to reduce export volume.

Unlike the collector's [interval processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/intervalprocessor), which forward the latest value (causing data loss for Deltas), this implementation explicitly sums the delta values over the interval.

Metrics resource mapping is based on the [interval processor logic](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/intervalprocessor/processor.go#L57-L63).

### DIFF
--- a/connector/profilingmetricsconnector/config.go
+++ b/connector/profilingmetricsconnector/config.go
@@ -41,8 +41,8 @@ type Config struct {
 	// in memory before sending them to the next pipeline consumer.
 	//
 	// If set to a value greater than 0s, metrics are buffered and flushed
-	// collectively at the end of each interval.
-	// If set to 0s (default), aggregation is disabled and metrics are forwarded after
+	// collectively at the end of each interval (30s default).
+	// If set to 0s, aggregation is disabled and metrics are forwarded after
 	// a Profile is received.
 	FlushInterval time.Duration `mapstructure:"flush_interval"`
 }
--- a/connector/profilingmetricsconnector/factory.go
+++ b/connector/profilingmetricsconnector/factory.go
@@ -19,6 +19,7 @@ package profilingmetricsconnector // import "github.com/elastic/opentelemetry-co
 
 import (
 	"context"
+	"time"
 
 	"github.com/elastic/opentelemetry-collector-components/connector/profilingmetricsconnector/internal/metadata"
 	"go.opentelemetry.io/collector/component"
@@ -38,6 +39,7 @@ func NewFactory() connector.Factory {
 func createDefaultConfig() component.Config {
 	return &Config{
 		MetricsBuilderConfig: metadata.DefaultMetricsBuilderConfig(),
+		FlushInterval:        30 * time.Second,
 	}
 }
 


christos68k:
@rogercoll Does this need further work before removing 'draft' status (so that it can be reviewed) ?

connector/profilingmetricsconnector/internal/identity/resource.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+	"hash"
+	"hash/fnv"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/pdatautil"
+)
+
+type resource = Resource
```
christos68k: Do we need this? Why can't we embed `Resource` ?

connector/profilingmetricsconnector/internal/identity/scope.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (s Scope) Resource() Resource {
+	return s.resource
+}
+
+func (s Scope) String() string {
+	return fmt.Sprintf("scope/%x", s.Hash().Sum64())
+}
+
```
christos68k: Can we add a docstring for this function?

connector/profilingmetricsconnector/internal/identity/stream.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (s Stream) Metric() Metric {
+	return s.metric
+}
+
+func (s Stream) String() string {
+	return fmt.Sprintf("stream/%x", s.Hash().Sum64())
+}
+
```
christos68k: Can we add a docstring for this function?

connector/profilingmetricsconnector/consumer_agg.go
```diff
@@ -0,0 +1,213 @@
...
+	mLookup      map[identity.Metric]pmetric.Metric
+	numberLookup map[identity.Stream]pmetric.NumberDataPoint
+}
+
+func newAggConsumer(nextConsumer consumer.Metrics, flushInterval time.Duration, logger *zap.Logger) *aggConsumer {
+	return &aggConsumer{
+		nextConsumer:  nextConsumer,
+		flushInterval: flushInterval,
+		md:            pmetric.NewMetrics(),
+		stateLock:     sync.Mutex{},
```
christos68k: This is not needed as the zero value is already an unlocked `Mutex`:


```suggestion
```

connector/profilingmetricsconnector/config.go
```diff
@@ -32,4 +36,13 @@ type Config struct {
 
 	// CustomAggregations allows to generate custom metrics.
 	CustomAggregations []Aggregation `mapstructure:"aggregations"`
+
+	// FlushInterval determines the time window for aggregating delta metrics
+	// in memory before sending them to the next pipeline consumer.
+	//
+	// If set to a value greater than 0s, metrics are buffered and flushed
+	// collectively at the end of each interval.
+	// If set to 0s (default), aggregation is disabled and metrics are forwarded after
```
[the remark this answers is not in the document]
rogercoll: Yes, let's not force users to modify the default config https://github.com/elastic/opentelemetry-collector-components/pull/950/commits/a81e705982da1b2d4143461f4fc25e7f4d674d39

PR #957, opened 2026-01-13
Remove elastictraceprocessor

axw:
elastictraceprocessor is deprecated, elasticapmprocessor should be used instead.

Tidy up some stray references elsewhere.


gregkalapos (approved):
I think it's ok to remove it, but I'm not sure about deprecation periods. 

This is a PR in `elastic-agent` which needs to be reverted: https://github.com/elastic/elastic-agent/pull/10588 Things changed over there, but it may still help a bit.

PR #959, opened 2026-01-14
elasticapmprocessor: refactor and simplify tests

axw:
The main change is to remove the old TestProcessor, and consolidate on what was TestProcessorECS. i.e. we have a single test with subtests for both otel and ecs mapping modes.

TestECSTraces is removed. It was doing nothing. Since the input was metrics (surely a human error), it meant that `golden.ReadTraces` was reading no traces at all. We can add subtests to TestProcessor as needed; we already have tests covering traces.

Changed all tests to not explicitly set a Logger on TelemetrySettings; just use the nop logger that is created by processortest.NewNopSettings.


marclop (approved):
LGTM!

PR #983, opened 2026-01-30
ratelimitprocessor: remove gubernator (& dynamic)

marclop:
Removes all the gubernator (& dynamic) dependencies and leaves just the local rate limiter.


axw (approved):
LGTM!

PR #985, opened 2026-01-30
feat: use helper funcs for repeated check+update operations

lanre-ade:
This pull request refactors how attributes are added to maps in the Elastic APM intake receiver, improving code readability and reducing duplication. New helper functions are introduced to handle common patterns for setting `string`, `integer`, and `boolean` attributes only when values are present.

The most important changes are:

**Helper function additions:**
- Added `putPtrInt` and `putPtrBool` generic helper functions in `attribute_helper.go` to concisely handle setting integer and boolean attributes from pointer values, including logic to clamp large unsigned integers to avoid overflow.

**Refactoring attribute assignment in mappers:**
- Replaced repeated conditional checks for setting string, integer, and boolean attributes throughout the mapping code with calls to the new helper functions, resulting in much cleaner and more maintainable code in `intakeV2ToDerivedFields.go` and `intakeV2ToElasticSpecificFields.go`.

### DIFF
--- a/receiver/elasticapmintakereceiver/internal/mappers/attribute_helper_test.go
+++ b/receiver/elasticapmintakereceiver/internal/mappers/attribute_helper_test.go
@@ -1,0 +1,464 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


receiver/elasticapmintakereceiver/internal/mappers/attribute_helper.go
```diff
@@ -28,3 +30,28 @@ func putNonEmptyStr(attributes pcommon.Map, key, value string) {
 		attributes.PutStr(key, value)
 	}
 }
+
+// putPtrInt puts an int attribute in the given map
+// only if the provided value is not nil.
+// Supports signed and unsigned integer pointer types.
+// For uint64 values exceeding int64 max, the value is clamped
+// to math.MaxInt64 to prevent overflow.
+func putPtrInt[T ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~int8 | ~int16 | ~int32 | ~int64](attributes pcommon.Map, key string, value *T) {
```
isaacaflores2: suggestion: I do not see any issues with this. I think we can add a unit test for this func to test all the edge cases. I do not think the current test data files integer values that will overflow

isaacaflores2 (commented):
Thanks for the clean up! I just had one suggestion

isaacaflores2 (approved):
Thanks!

PR #988, opened 2026-02-02
[elasticapm processor] Add support for DB Semvonv v33+ (after DB stabilization)

gregkalapos:
After DB SemConv became stable, there were some breaking changes introduced: https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/ - very basic field names changed as well.

This small PR adds support to handle the new DB SemConv key names. The change is incremental - old semconv versions still work.

Without this, the current code does not recognize basic db spans as db span and does not add `span.destination.service.resource`.

### DIFF
--- a/processor/elasticapmprocessor/internal/enrichments/span.go
+++ b/processor/elasticapmprocessor/internal/enrichments/span.go
@@ -196,7 +196,7 @@ func (s *spanEnrichmentContext) Enrich(
 			s.isRPC = true
 			s.rpcService = v.Str()
 		case string(semconv25.DBStatementKey),
-			string(semconv25.DBUserKey), string(semconv27.DBQueryTextKey):
+			string(semconv25.DBUserKey), string(semconv37.DBQueryTextKey):
 			s.isDB = true
 		case string(semconv25.DBNameKey), string(semconv37.DBNamespaceKey):
 			s.isDB = true


processor/elasticapmprocessor/internal/enrichments/span.go
```diff
@@ -196,12 +196,12 @@ func (s *spanEnrichmentContext) Enrich(
 			s.isRPC = true
 			s.rpcService = v.Str()
 		case string(semconv25.DBStatementKey),
-			string(semconv25.DBUserKey):
+			string(semconv25.DBUserKey), string(semconv37.DBQueryTextKey):
 			s.isDB = true
-		case string(semconv25.DBNameKey):
+		case string(semconv25.DBNameKey), string(semconv37.DBNamespaceKey):
```
gregkalapos: These attributes were introduced in v33, but since we already have semconv37 pulled in, I think it's sufficient to just use that. These are stable since v33, so v37 is the same, there seems to be no changes in between.

carsonip (approved):
thanks, I'm wondering if we should add the same support to apm server cc @raultorrecilla

raultorrecilla:
> thanks, I'm wondering if we should add the same support to apm server cc @raultorrecilla

thanks for the heads up @carsonip, https://github.com/elastic/apm-server/issues/20248 for next it

PR #994, opened 2026-02-04
[receiver/elasticapmintake] Add missing transaction attribute mappings for error events

isaacaflores2:
# Overview

This PR adds missing transaction mappings for errors. APM error events may contain info on the correlated transaction: https://github.com/elastic/apm-data/blob/df67c09/input/elasticapm/internal/modeldecoder/v2/decoder.go#L409-L419

# Testing
- Updated output file to assert expected `transaction.*` attributes are present


gregkalapos (approved):

isaacaflores2:
> Looks fine, though I thought we should be doing this in elasticapmprocessor?

> My take is that if a field exists in the intake json, then we should take those values as-is and set those in the elasticapmintake receiver. So for OTLP, such fields are derived in the processor, but for elasticapmintake in my view, it's better to set everything possible here directly. My reasoning is that it's simpler than running some enrichment logic in the processor, even though we have the data already in the incoming data.

Yes that was my reasoning as well. These `transaction.*` fields already exists so we are not deriving/enriching the event here. We are simply mapping the fields to the correct otel attribute.

axw:
Right, makes sense. I had in mind that we were doing enrichment here.

PR #1007, opened 2026-02-13
Fix mis log discrepancies

lanre-ade:


processor/elasticapmprocessor/internal/ecs/ecs_translation.go
```diff
@@ -68,50 +68,31 @@ const (
...
-		} else if !isSupportedAttribute(k) {
-			// Other attributes that are not supported by ECS are moved to labels with a "labels." prefix.
-			attributes.PutStr("labels."+sanitizeLabelKey(k), v.AsString())
+		if !isSupportedAttribute(k) && !isInfraAttribute(k) {
+			// Unsupported resource attributes are moved to labels when
+			// their type is representable (str, bool, int, double, slice).
+			// Types that cannot be stored as labels (Map, Bytes, Empty)
+			// are intentionally dropped, matching apm-data behaviour
+			// (see input/otlp/traces.go case pcommon.ValueTypeMap).
+			setLabelAttributeValue(attributes, replaceDots(k), v)
```
isaacaflores2: suggestion: I think you should merge the latest from main. The replaceDots function was removed recently to use `sanitizeLabelAttributeKey`:  https://github.com/elastic/opentelemetry-collector-components/blob/main/processor/elasticapmprocessor/internal/ecs/ecs_translation.go#L96-L123

processor/elasticapmprocessor/internal/ecs/log_attribute_conventions.go
```diff
@@ -0,0 +1,98 @@
...
+	return false
+}
+
+// ApplyScopeDataStreamConventions sanitizes data_stream dataset/namespace on
+// log record attributes. If a value is already present on the record it is
+// sanitized in place; otherwise a sanitized copy is taken from the scope.
+// This is the single sanitization point for user-provided data_stream values
+// — downstream routing functions (EncodeErrorDataStream, etc.) only set
+// known-safe hardcoded values.
+func ApplyScopeDataStreamConventions(scopeAttributes, logAttributes pcommon.Map) {
```
isaacaflores2: question: Are these data_stream attributes expected to only exists at the scope level? I only see the apm-data library checking for these attributes in the log record

processor/elasticapmprocessor/internal/ecs/log_conventions.go
```diff
@@ -0,0 +1,96 @@
...
+const (
+	defaultUnknownAgentVersion = "unknown"
+	defaultUnknownServiceName  = "unknown"
+	defaultUnsetEnvironment    = "unset"
+)
+
+var serviceNameInvalidRegexp = regexp.MustCompile("[^a-zA-Z0-9 _-]")
+
+// ApplyLogConventions applies defaults and sanitization for ECS log data flow.
+func ApplyLogConventions(resource pcommon.Resource) {
```
isaacaflores2: suggestion: The `service.name` and `service.environment` logic should apply to all signals. So it should not be specific to logs

processor/elasticapmprocessor/internal/ecs/log_conventions.go
```diff
@@ -0,0 +1,96 @@
...
+		return
+	}
+	deploymentEnvironmentName, deploymentEnvironmentNameOK := attrs.Get(string(semconv.DeploymentEnvironmentNameKey))
+	if deploymentEnvironmentNameOK && deploymentEnvironmentName.Str() != "" {
+		return
+	}
+	attrs.PutStr(string(semconv26.DeploymentEnvironmentKey), defaultUnsetEnvironment)
+}
+
+func setDefaultAgentVersion(attrs pcommon.Map) {
```
isaacaflores2: suggestion: I think the logic for the `agent.version` is already [present](https://github.com/elastic/opentelemetry-collector-components/blob/main/processor/elasticapmprocessor/internal/enrichments/resource.go#L146-L159). It is just disabled. @aelnahas should be looking into this already.

receiver/elasticapmintakereceiver/go.mod
isaacaflores2: question: why does `elasticattr` need to be updated here and in the other components?

aelnahas (commented):
Is it worth splitting this PR a little bit? IMO I think there is a lot of logic that we will be committing at once

processor/elasticapmprocessor/internal/ecs/label_attribute_mapping.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+// setLabelAttributeValue maps a value into labels.* / numeric_labels.*.
+// It returns true if the value was stored, false if the type is not
+// representable as a label. Elasticsearch label mappings only support
+// flat scalar values and homogeneous arrays thereof; Map, Bytes, and
+// Empty types cannot be stored and are intentionally dropped. This
+// matches the behaviour of apm-data's setLabel (input/otlp/metadata.go)
+// which also silently ignores these types.
+func setLabelAttributeValue(attributes pcommon.Map, key string, value pcommon.Value) bool {
```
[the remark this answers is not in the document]
lanre-ade: @isaacaflores2 I initially returned a bool to indicate whether the value was set or dropped. It isn't being used but I was worried that some values were silently dropped for example are `pcommon.ValueTypeMap`, `pcommon.ValueTypeBytes` value types we can expect, how should they be treated? 

The other scenarios are no-ops (empty value, empty list) so can be ignored safely.

processor/elasticapmprocessor/internal/ecs/log_attribute_conventions.go
lanre-ade: `apm-data` writes these in `translateScopeMetadata` in [metadata.go](https://github.com/elastic/apm-data/blob/34677210900a68d6204cdb79da4ce0d1ee685d9a/input/otlp/metadata.go#L462-L461) and then overwrites them again in [logs.go ](https://github.com/elastic/apm-data/blob/6fb3770cc7ea9531de38f5804582fe9e52f7d020/input/otlp/logs.go#L192-L191) so I believe that's the case but I might be missing something

PR #1021, opened 2026-02-18
Add retry backoff to apikeyauth Elasticsearch client

constanca-m:
The error `rpc error: code = Internal desc = error checking privileges for API Key: status: 503` was causing SLO failures. On Elasticsearch Serverless, 503 is a transient error. Our ES client was retrying immediately (no backoff), so all retries hit the same 503 before the cluster had time to recover.

We now expose a new `retry` field to add exponential backoff to the retries so they don't run immediately after each other.


carsonip (commented):
nit on backoff without jitter, otherwise lgtm

PR #1023, opened 2026-02-18
Bump upstream OTel dependencies - Release/v0.33.0

gregkalapos:
For the previous release, most of the upstream components were not updated correctly. We merged https://github.com/elastic/opentelemetry-collector-components/pull/993, but that missed most of `go.mod` files and left old dependencies.

In this PR I tried to update things everywhere.

Goal: Make sure upstream is on [v1.51.0/v0.145.0](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.145.0).

Side note: since then - at the same time as I worked on it, a new release was out. It wasn't fully released when I started with this, so I targeted v1.51.0/v0.145.0 - since that was the latest full release.

PR #1025, opened 2026-02-19
Add note to release process about missing bumps 

gregkalapos:
From https://github.com/elastic/opentelemetry-collector-components/pull/1023#issuecomment-3925332152

I saw this happening already twice, so as @ChrsMark suggested, adding a note about it.


carsonip (commented):
suggestion: will it be better to add a makefile target and/or gh workflow and/or ci check to make the process more foolproof? We have something similar in our team to bump upstream version. We can develop some tooling quite quickly these days with AI and get ROI fast compared to a release documentation addition.

PR #1026, opened 2026-02-20
[receiver/elasticapmintakereceiver] Update `span.self_time.sum.us` to set value in microseconds

isaacaflores2:
# Overview

The apm-data provided value for `SelfTime.Sum` is in nanoseconds but JSON marshaled as microseconds during index time [here](https://github.com/elastic/apm-data/blob/v1.19.5/model/modeljson/internal/metricset.go#L115). This updates the attribute value to be set as microseconds, so no conversion is needed at export time .

# Testing
1. Updated unit-test to assert the expected value


carsonip (approved):

isaacaflores2:
Hmm the govulncheck is failing since the otel sdk needs to be bumped to `go.opentelemetry.io/otel/sdk@v1.40.0`. The upstream releases for [v1.52.0/v0.146.1](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.146.1) should solve this, so hopefully dependabot does its thing and I will follow up on upgrading all the components.

PR #1033, opened 2026-02-20
[receiver/elasticapmintake] Stop setting `timestamp.us` and `event.outcome`

isaacaflores2:
# Overview
This PR updates the intake receiver logic so `timestamp.us` and `event.outcome` will no longer be set for metric data points . These fields are not set in apm-data so this will match the existing apm behavior:
- timestamp.us only for transaction/span/error: https://github.com/elastic/apm-data/blob/main/model/modeljson/apmevent.pb.json.go#L99-L109
- metricset spec has no event/outcome: https://github.com/elastic/apm-data/blob/main/input/elasticapm/docs/spec/v2/metricset.json

# Testing 
1. Updated unit test

### DIFF
--- a/receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToDerivedFields.go
+++ b/receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToDerivedFields.go
@@ -31,6 +31,8 @@ import (
 
 // SetDerivedFieldsForTransaction sets fields that are NOT part of OTel for transactions. These fields are derived by the Enrichment lib in case of OTLP input
 func SetDerivedFieldsForTransaction(event *modelpb.APMEvent, attributes pcommon.Map) {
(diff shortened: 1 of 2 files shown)


gregkalapos (approved):
Change looks ok to me.

Only thing I was thinking about is that `SetDerivedFieldsCommon` is not that `common` anymore with this change. Do we even use that for something other than spans? 🤔 If it's not used elsewhere, maybe it'd be better to either rename it, or just move the code to the point where it's used.

isaacaflores2:
> Change looks ok to me.
> 
> Only thing I was thinking about is that `SetDerivedFieldsCommon` is not that `common` anymore with this change. Do we even use that for something other than spans? 🤔 If it's not used elsewhere, maybe it'd be better to either rename it, or just move the code to the point where it's used.

Thanks I updated refactor the function into `setCommonDerivedFieldsForSpanAndTransaction`

PR #1046, opened 2026-02-25
Implement verifier receiver

jeniawhite:
Implementation of a Verifier receiver to the Elastic distribution of OpenTelemetry (EDOT). The receiver checks that Cloud Connector–based integrations have the required cloud/identity provider permissions and reports results as OTEL logs to Elasticsearch.

Cloud Connector deployments need to know whether the configured IAM/API credentials have the permissions required by each integration. This receiver runs on-demand (or scheduled) verification: it performs minimal API calls per required permission, groups results by Cloud Connector, policy, and integration, and emits structured OTEL logs so users can see granted/denied/error status and fix missing permissions.

Replaces #1005 (reopened from a direct branch to enable CI checks).

### DIFF
--- a/receiver/verifierreceiver/config.go
+++ b/receiver/verifierreceiver/config.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/elastic/opentelemetry-collector-components/receiver/verifierreceiver/internal/verifier"
 )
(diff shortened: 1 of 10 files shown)


gregkalapos:
I could review this from the general `opentelemetry-collector-components`, but certainly not for the specifics of how it works. Could maybe someone from your team @jeniawhite review the specific code? So my proposal would be that I can give an ok to merge into this repo, but then someone more familiar with the domain also looks at the specific logic as well. Would that be ok?

Few questions from my side:
- Why should this be in this repo and should it be part of the default EDOT distribution? 
  - Asking to see how we put this [here](https://github.com/elastic/opentelemetry-collector-components/blob/main/versions.yaml). Also depending on the answer you may need additional PRs in elastic agent.
 - Somewhat related: what's when the collector doesn't run in any of those cloud environments? I see this'd print warnings when `InitializeVerifier` fails. Just wanna make sure this behaves properly in environments where this receiver has nothing to do - especially if this'd be by default part of EDOT. I looked a bit at `receiver_test.go‎` and it's not clear to me that e.g. in a plain docker container (where from what I understand this receiver should't do anything), our EDOT collector would behave 100% the same with and without this receiver.

jeniawhite:
Closing in favor of https://github.com/elastic/elastic-agent/pull/13021 which moves the verifier receiver directly into elastic-agent, per feedback from @axw and @cmacknz that this receiver is EDOT Collector-only and should live in elastic-agent under ELv2.

PR #1061, opened 2026-03-03
[connector/dynamicrouting]add ttl-based decision persistence and near expiry updates

lahsivjar:
This PR decouples routing-decision lifetime from cardinality recording cadence in `dynamicroutingconnector`, so routing can remain stable for batching while measurements refresh on a shorter cadence. Previously one option (`evaluation_interval`) controlled both:
- how long routing decisions stayed stable, and
- how often measurements were recorded/re-evaluated.

That created a tradeoff for batching with `wait_for_result=true`:
- long interval: stable routing, stale measurement
- short interval: fresher measurement, more route churn

With this PR, I have separated the above concerns for stable decisions:
- `ttl` controls stability,
- `recording_interval` controls refresh cadence,
- near-expiry recording reduces unnecessary work while refreshing before expiry.

### Benchmark Diff (A/B) (benchmarks are executed and generated by Claude)

Compared **baseline HEAD (without current changes)** vs **current changes** using:

- `go test -run '^$' -bench BenchmarkRouter -benchmem -count=8`
- `benchstat`

#### CPU (`sec/op`)
- `Router/1pk_1mk/traces`: **+23.64%**
- `Router/1pk_1mk/metrics`: **+22.32%**
- `Router/1pk_1mk/logs`: **+21.68%**
- `Router/3pk_3mk/traces`: **-4.35%**
- `Router/3pk_3mk/metrics`: **-5.31%**
- `Router/3pk_3mk/logs`: **-2.71%**
- **Geomean**: **+8.39%**

#### Memory (`B/op`)
- `1pk_1mk`: **-4.65%**
- `3pk_3mk`: **-11.32%**
- **Geomean**: **-8.05%**

#### Allocations (`allocs/op`)
- `1pk_1mk`: **7 -> 6** (**-14.29%**)
- `3pk_3mk`: **11 -> 8** (**-27.27%**)
- **Geomean**: **-21.05%**

Interpretation:
- Small-key path pays fixed TTL/expiry overhead (CPU up).
- Larger-key path benefits from skipping unnecessary hash/HLL updates (CPU down).
- Allocation and memory improve across all benchmark variants.

**NOTE that the increase in CPU is expected as the logic for ttlcache requires more computation overall. Some of the overheads will be amortized under practical situations when `recording_interval << ttl`**

Related to: https://github.com/elastic/hosted-otel-collector/issues/2333


connector/dynamicroutingconnector/README.md
```diff
@@ -78,7 +78,8 @@ connectors:
       - pipelines: ["traces/very_high_cardinality"]
         max_cardinality: .inf
     default_pipelines: ["traces/default"]
-    evaluation_interval: 30s
+    recording_interval: 30s
+    ttl: 5m
```
vigneshshanmugam: Ques - Do we want the TTL to be configurable? I m worried its exposing too many configuration knobs. 

Wouldn't it be safe to keep the routing stable till the next recording interval is mature? Like ttl would be >= 2x the recording interval?

PR #1088, opened 2026-03-13
[poc] [connector/elasticapm] Add dynamic_resource_attributes_key to support global labels

isaacaflores2:
# Overview

This is a prototype to explore supporting dynamic resource attributes (global labels)

Related to: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46884

## Approach
**signaltometricsconnector** --  Provides a OTTL statement with a resource attribute filter which can include all attributes matching that prefix instead of a single key. 

**elasticapmconnector** -- Relies on the `x-dynamic-resource-attributes` key which represents a key to read a
comma-separated list of allowed attribute keys from the `client.Metadata`. The `elasticapmconnector` will  use  the  below OTTL statement to only include resource attributes that have a specific prefix and whose key is included in the allow list provided by `x-dynamic-resource-attributes` client metadata key. 
```yaml
FilterMapByKeyList(resource.attributes, otelcol.client.metadata["x-dynamic-resource-attributes"], ["labels.", "numeric_labels."])
```

The upstream components such as the `elasticapmintakereceiver` will add `x-dynamic-resource-attributes` to the context metadata to be propagated to the `elasticapmconnector`.  The `elasticapmintakereceiver` will also add all the `labels.*`/`numeric_labels.*` resource attributes which already happens today.


isaacaflores2:
Closing since this has been implemented in a separate PR

PR #1119, opened 2026-03-27
[receiver/akamaisiemreceiver] - Initial implementation of Akamai SIEM native OTEL receiver

ShourieG:
## Native Akamai SIEM Receiver

This is a dedicated OTel receiver for the Akamai SIEM API. It implements polling, the chain state machine, and NDJSON streaming directly with native OTel Collector interfaces. It targets the existing data stream and ingest pipeline in Elasticsearch for the Akamai integration.

### Why

The CEL-based input works but uses a generic CEL execution engine, which adds overhead and limits control over Akamai-specific offset and chain details. This receiver communicates with the Akamai API directly through EdgeGrid authentication, NDJSON streaming, offset pagination, and chain recovery, without the CEL layer.

Related: [elastic/security-integrations#733](https://github.com/elastic/security-integrations/issues/733)

### Proposed Commit message
```
feat(receiver/akamaisiemreceiver): add native Akamai SIEM receiver

A dedicated receiver for the Akamai SIEM API.

- EdgeGrid HMAC-SHA256 auth. NDJSON responses are parsed line by line with a one-line delay. This keeps the trailing offset-context line out of the event stream.

- Three-branch chain state machine, including offset drain, chain replay, and new chain, features 416 recovery, timestamp retries, clamping for too-old data, and offset TTL handling.

- Bounded-memory streaming sets up a small channel between the scanner and the batched ConsumeLogs consumer. This limits peak memory usage regardless of page size.

- Optional cursor persistence is available through the OTel storage extension. Without this option, restarts will re-fetch from initial_lookback.

- Each record has a body map keyed "message" containing the raw JSON, plus additional data_stream information such as {type, dataset, namespace} on the body and resource. It also includes elastic.mapping.mode: bodymap on the scope. The pipeline can remain at [batch] since the Akamai integration's ingest pipeline handles ECS enrichment.

- There are 16 metrics and spans across Poll, FetchPage, ProcessPage, EmitEvents, and PersistCursor. FetchPage also carries the truncated API error body for non-200 responses to simplify debugging.
```
### How it works

The receiver polls `/siem/v1/configs/{id}` using EdgeGrid HMAC-SHA256 signed requests. The responses are gzipped NDJSON. A scanner goroutine reads lines into a bounded channel (`stream_buffer_size`). It uses a one-line-delay pattern that separates events from the trailing offset-context line. A consumer goroutine batches these events and calls `ConsumeLogs` for each batch (`batch_size`). Peak memory is limited to `stream_buffer_size + batch_size` events, regardless of the page size.

The cursor state persists through the OTel storage extension interface (`storage.Client`) after each successful page. Storage is optional. Without it, chain state still tracks across poll cycles in memory, but every collector restart will re-fetch from `initial_lookback`.

The poller operates a three-branch state machine:
- **Offset drain**: continues from a stored offset (steady state).
- **Chain replay**: if the offset is expired or missing, it replays from a time window with overlap.
- **New chain**: for the first run or if caught up; it starts a fresh time-based window.

Offset TTL detection, 416 recovery, invalid timestamp retries, and "from too old" clamping are all managed within the state machine.

### Output shape

Each log record carries the raw Akamai JSON in `LogRecord.Body`, with a map keyed by `message`, alongside `data_stream.{type,dataset,namespace}` body keys for Kibana filters. The same `data_stream.*` values are also included in **resource attributes** for the Elasticsearch exporter’s dynamic routing. The **scope** carries `elastic.mapping.mode: bodymap`, allowing the ES exporter to serialize the body map fields directly into the indexed document.

Scope attributes are part of the `plog.Logs` data structure and endure batch processor flushes and exporter queues without needing `metadata_keys` configuration. This is recommended by the [ES exporter migration docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/elasticsearchexporter#migration-setting-mapping-mode-via-scope-attribute). Since the receiver writes the scope attribute itself, there is no need for a `transform` processor in the user pipeline; the common case simplifies to `[batch]`.

`data_stream.*` defaults to `logs / akamai.siem / default`, which can be configured on the receiver. The Akamai integration’s ingest pipeline handles ECS enrichment; the receiver does not parse or transform event content. Existing Kibana dashboards are immediately usable.

### Observability

There are 16 metrics covering API health (`requests`, `request_errors`, `request_duration`, `bytes_received`), event flow (`events_received`, `events_emitted`, `events_per_page`, `events_per_second`), pagination (`pages_processed`, `cursor_persists`, `offset_expired`, `offset_ttl_drops`), recovery (`recovery_attempts`, `invalid_timestamp_retries`), and timing (`page_processing_time`, `poll_duration`).

Spans cover `Poll`, `FetchPage`, `ProcessPage`, `EmitEvents`, and `PersistCursor`. For non-200 responses, the `FetchPage` span includes `akamai.api.status_code`, `akamai.api.detail`, and `akamai.api.body` (truncated to 2 KB). This helps operators debug proxy errors and HTML error pages without needing debug logs.

### Benchmarks

For the mock Akamai API, 100k events per page, nop exporter, using Apple M1 Max, the metrics are:

| Metric | Value |
|---|---|
| **EPS** | ~15.6k |
| CPU/event | 0.009ms |
| KB/event | 1.4 |
| RSS | 105 MB |

The process is I/O-bound on gzip decompression and NDJSON streaming. Body-map construction effectively uses zero CPU per event.

### What's in the PR

The PR contains 15 commits, organized for review:

| # | Commit | What |
|---|---|---|
| 1 | `feat: add module scaffold and telemetry definitions` | `go.mod`, `Makefile`, `metadata.yaml` (16 metrics), generated metadata code, `documentation.md`. |
| 2 | `feat: add EdgeGrid HMAC-SHA256 authentication` | HMAC-SHA256 signing, `http.RoundTripper` transport wrapper. |
| 3 | `feat: add HTTP client and NDJSON streaming` | HTTP client, URL builder, `APIError` types, `StreamEvents` with one-line-delay pattern. |
| 4 | `feat: add cursor persistence via storage extension` | `Cursor`, `CursorStore` over `storage.Client`, TTL detection. |
| 5 | `feat: add poller state machine and fetch loop` | Three-branch state machine, page processing, error recovery, span/metric instrumentation. |
| 6 | `feat: add config and validation` | `Config` (squashed `confighttp.ClientConfig`), `data_stream` block, `Validate`. |
| 7 | `feat: add factory wiring` | OTel `receiver.Factory`, `createLogsReceiver`. |
| 8 | `feat: add receiver lifecycle and bodymap emission` | `Start`/`Shutdown`, `pollLoop`, `emitEvents` (body map + scope + resource attrs). |
| 9 | `test: add receiver, integration, and tracing tests` | Mock-API integration tests, scope/resource/body assertions, cursor persistence, tracing. |
| 10 | `test: add benchmarks` | `emitEvents` and full-poll benchmarks at multiple event counts. |
| 11 | `docs: add README` | Architecture, configuration reference, tuning guide, pipeline component reference, scenarios. |
| 12 | `docs: add architecture diagrams` | Two SVGs — overall architecture and bodymap data flow. |
| 13 | `feat(distributions/elastic-components): register akamaisiemreceiver` | Manifest gomod entry + replace directive. |
| 14 | `chore: register akamaisiemreceiver ownership and ignore local artifacts` | `CODEOWNERS` entry, `.gitignore` additions. |
| 15 | `chore: regenerate metadata` | `mdatagen` output refresh. |

### Package structure

```
receiver/akamaisiemreceiver/
├── config.go, factory.go, receiver.go
└── internal/
    ├── akamaiclient/    HTTP client, NDJSON streaming, API errors.
    ├── auth/            EdgeGrid HMAC-SHA256 signing.
    ├── cursor/          Cursor state, storage extension persistence.
    ├── metadata/        Auto-generated telemetry.
    └── poller/          Three-branch state machine, fetch loop.
```

### Tests

Integration tests run against a local `httptest` server with realistic NDJSON fixtures. They cover body map shape, scope and resource attribute injection, cursor persistence through a mock storage extension, severity defaults, and error recovery paths. Unit tests address config validation, EdgeGrid signing, NDJSON streaming, cursor operations, and poller state transitions. Benchmarks gauge emit throughput and full poll cycles.

End-to-end tests against a local Elasticsearch 8.17 stack with the Akamai integration installed confirm that documents reach `logs-akamai.siem-default` with full ECS enrichment from the ingest pipeline. This is true for both batch and non-batch pipeline configurations.

### Quick start

```yaml
receivers:
  akamai_siem:
    endpoint: "https://akab-xxxxx.luna.akamaiapis.net"
    config_ids: "12345"
    authentication:
      client_token: "${AKAMAI_CLIENT_TOKEN}"
      client_secret: "${AKAMAI_CLIENT_SECRET}"
      access_token: "${AKAMAI_ACCESS_TOKEN}"
    storage: file_storage   # optional — without it, every restart re-fetches from initial_lookback
    # data_stream defaults to logs / akamai.siem / default

extensions:
  file_storage:
    directory: /var/lib/otelcol/storage

processors:
  batch:                    # optional — recommended for amortizing flushes
    timeout: 10s

exporters:
  elasticsearch:
    endpoints: ["https://elasticsearch:9200"]
    api_key: "${ES_API_KEY}"

service:
  extensions: [file_storage]
  pipelines:
    logs:
      receivers: [akamai_siem]
      processors: [batch]
      exporters: [elasticsearch]
```

The minimum required setup includes just the receiver and an exporter. Everything else—`batch`, `file_storage`, `sending_queue`, metrics endpoint—is optional and noted in the README's scenarios. Additional scenarios, such as file export with rotation, console debug, and high-throughput production with persistent queueing, are detailed in the **README**.

---
### Screenshots
---

**ECS Discover Dashboard:**

akamai_ecs_discover

**Akamai SIEM Dashboard 1:** 

Akamai_dashboard_1

**Akamai SIEM Dashboard 2:**

Akamai_dashboard_2


receiver/akamaisiemreceiver/internal/mapper/otel.go
```diff
@@ -0,0 +1,388 @@
...
+//
+// Mapping strategy:
+//   - Fields with OTel semantic conventions use the standard attribute name
+//   - Akamai-specific fields use the "akamai.siem.*" namespace
+//   - The raw JSON is always preserved in the body for debugging/reprocessing
+func MapToOTelLog(rawJSON string, lr plog.LogRecord) error {
+	var event map[string]any
+	if err := json.Unmarshal([]byte(rawJSON), &event); err != nil {
+		return fmt.Errorf("failed to parse event JSON: %w", err)
+	}
```
axw: This is a very expensive way to do the translation. It would be best to avoid reflection altogether if possible, e.g. by walking the JSON and setting fields as you go. As a quicker win, define structs with well-defined SIEM event fields, and unmarshal into that to avoid a lot of interface boxing.

axw (commented):
I've only skimmed, since I was curious about the performance. There's some low hanging fruit for optimising the full otel SemConv configuration.

axw:
@ShourieG sorry for the delay, this dropped off my radar after the long weekend.

> Hey @axw, @andrewkroh wants to review this but he does not seem to have the permission to add himself as a reviewer.

I'll look into how we can fix that.

> The scraperhelper model does not fit the custom akamai poll logic, how we use chain draining, streaming the logs and emitEvents() per batch inside a page. scraperhelper would not fit with the streaming model with our custom chaining logic here.

OK. FYI, the reason I ask is because we'll soon have the ability to trigger scrapes with alternative methods other than a simple in-process timer: https://github.com/open-telemetry/opentelemetry-collector/pull/14469

receiver/akamaisiemreceiver/receiver.go
```diff
@@ -0,0 +1,456 @@
...
+	return nil
+}
+
+// emitRaw sends events as raw JSON body strings. The ES ingest pipeline handles all
+// field parsing and ECS enrichment downstream. Injects x-elastic-mapping-mode: ecs
+// into the OTel client context so the elasticsearchexporter serializes
+// LogRecord.Body → "message" field (not "body.text") without requiring explicit
+// mapping.mode config on the exporter.
+func (r *akamaiReceiver) emitRaw(ctx context.Context, events []string, entry *consumerEntry) error {
+	ctx = withMappingMode(ctx, "ecs")
```
cmacknz: If you want this to be an ECS like document, the beats receiver approach is to use bodymap and transport the the complete ECS document in the LogRecord body.

What this ECS will do is attempt to translate the SemanticConvention names to ECS compatible ones for you. I believe the history was that this was written so that the Observability curated UIs would work so I don't think this is a perfect conversion. You can see it doesn't cover all of ECS in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/fe7f484ccf9f6716783b515a8abbfe486c891a05/exporter/elasticsearchexporter/model.go#L55 for example.

I don't think we have any magic "convert this OTLP document into a perfect ECS document" translation.

cmacknz: See also https://github.com/elastic/opentelemetry-collector-components/pull/1153/changes#r3058758463

ShourieG:
> @ShourieG sorry for the delay, this dropped off my radar after the long weekend.
> 
> > Hey @axw, @andrewkroh wants to review this but he does not seem to have the permission to add himself as a reviewer.
> 
> I'll look into how we can fix that.
> 
> > The scraperhelper model does not fit the custom akamai poll logic, how we use chain draining, streaming the logs and emitEvents() per batch inside a page. scraperhelper would not fit with the streaming model with our custom chaining logic here.
> 
> OK. FYI, the reason I ask is because we'll soon have the ability to trigger scrapes with alternative methods other than a simple in-process timer: [open-telemetry/opentelemetry-collector#14469](https://github.com/open-telemetry/opentelemetry-collector/pull/14469)

@axw  I think we can revisit this in future but right now as it is there are some bottlenecks besides a custom polling logic:

- emitEvents() calls both raw and OTel consumers in parallel per batch, scraperhelper has no concept of per-request routing to different consumers, I think atm.

- The current scraper signature is `func scrape(ctx context.Context) (plog.Logs, error)`, this is a limitation for use due to the streaming concept we are using to reduce back pressure. Our NDJSON streaming uses a bounded channel between the scanner and the consumer. If we had to buffer everything into a single `plog.Logs` return value, we'd need to hold 100k+ events in memory per page instead of streaming them in batches.

- Currently a single poll cycle fetches N pages until the chain is drained. Each page triggers multiple ConsumeLogs calls (one per batch of 1000/batch_size events). scraperhelper expects one scrape → one return → one ConsumeLogs. 

Will all these limitations change with the future updates ?

receiver/akamaisiemreceiver/receiver.go
ShourieG: @cmacknz, So in our scenario, in RAW mode we are sending to an ES backend where the integration assets are already installed. Sending the message on the appropriate index and data-stream triggers the existing akamai ingest pipeline to do the actual mapping. 

In the case of `elasticapmintakereceiver` the whole ECS mapping seems to be done in the receiver itself, where as in our case for ECS (RAW mode) mapping, we are relying on the integration being already installed and for that to do the actual heavy lifting. This is basically just acts as a simple message receiver/streamer in RAW mode. In OTEL mode it actually does the mapping in accordance with current OTEL standards. The header `x-elastic-mapping-mode: ecs` is only attached when we configure `output_format: raw`, so that it is sent as the original NDJSON log wrapped in `message`.

cmacknz: Using `bodymap` will use only the content of the LogRecord body as the document content in _bulk. 

Using the `ecs` mode will mutate the document to change any semantic conventions attributes you included to their ECS names. This is not a passthrough mode it changes the document, and it changes it in a way that is only tested against the existing Observability curated UIs.

I think what you want to do with the data is valid but I don't thin you are doing it using the mapping mode we intended to cover this case :)

cmacknz: > raw (default) — raw JSON goes into LogRecord.Body

From your PR description, this is the definition of the [bodymap](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/fe7f484ccf9f6716783b515a8abbfe486c891a05/exporter/elasticsearchexporter#bodymap-mapping-mode) mode not the ECS mode. "In bodymap mapping mode, the Elasticsearch Exporter supports only logs and will take the "body" of a log record as the exact content of the Elasticsearch document without any transformation"

axw:
> Will all these limitations change with the future updates ?

All valid points. I don't think the dual-output `emitEvents` one would change, but I think the others will need to, and I'm planning to raise them in the Collector SIG tomorrow. If we need to support sending to two consumers concurrently, then I think that could possibly be done with sharedcomponent.

receiver/akamaisiemreceiver/receiver.go
ShourieG: @cmacknz, the latest refactor in [this commit](https://github.com/elastic/opentelemetry-collector-components/pull/1119/changes/923679212a17170bbaa833801652ce9f42ca8f40) should address this. Now we are leveraging the transform processor for **bodymap**, updated all the configs, readme's and tests.

receiver/akamaisiemreceiver/internal/cursor/cursor.go
```diff
@@ -0,0 +1,92 @@
...
+
+package cursor // import "github.com/elastic/opentelemetry-collector-components/receiver/akamaisiemreceiver/internal/cursor"
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"go.opentelemetry.io/collector/extension/xextension/storage"
```
cmacknz, line 26: Have you tested that this works with https://github.com/elastic/beats/tree/main/x-pack/otel/extension/elasticsearchstorage? That's what you would need for compatibility with agentless.

Outside of agentless, users will need to manually configure a filestorage extension, the input won't be stateful by default like it is in Filebeat.

cmacknz:
> emitEvents() calls both raw and OTel consumers in parallel per batch, scraperhelper has no concept of per-request routing to different consumers, I think atm.

So one question is if you have a specific target use case in mind for this, or if it is implemented because it was possible. Who do you expect the first users of this input to be and what schema do they want to use?

One reason I ask is that doing it this way is adding complexity, and also because if you GA it with both removing either of the OTel or raw mode will be difficult. You could however add an additional mode in later if it is needed.

ShourieG:
> > emitEvents() calls both raw and OTel consumers in parallel per batch, scraperhelper has no concept of per-request routing to different consumers, I think atm.
> 
> So one question is if you have a specific target use case in mind for this, or if it is implemented because it was possible. Who do you expect the first users of this input to be and what schema do they want to use?
> 
> One reason I ask is that doing it this way is adding complexity, and also because if you GA it with both removing either of the OTel or raw mode will be difficult. You could however add an additional mode in later if it is needed.

@cmacknz, so from the get go we wanted OTEL compatibility knowing well that the current user base of the Akamai integration is used to ECS and is using the CEL input based integration. With the support for both OTEL and ECS we can encourage experimentation and possible transition from ECS depending on their needs or they could strictly stick to ECS without having to worry about OTEL. This also supports the OTEL philosophy of being backend agnostic with the option to support ECS specifically for our needs. 

The simultaneous dual mode transmission is an experimental feature and it's purely optional on the users end. We could keep it simplified to a single mode of operation sure but having this option gives us more experimental advantages and could lead to new user use cases down the line. 

**A real world scenario for Dual Mode** would be someone/team using Elastic SIEM for alerting but also feeding a data lake  or a custom OTLP backend for long-term retention. Having this feature would allow us to explore/suggest these use cases to users.

receiver/akamaisiemreceiver/internal/cursor/cursor.go
ShourieG: @cmacknz, Currently it's not compatible with `elasticsearchstorage` because it lacks a `GetClient()` method.  `elasticsearchstorage` needs to implement the **storage.Extension interface** on it's end to actually make it compatible. I can add storage specific code and libs and do something like

```
switch {
case isStorageExtension(ext):
    return ext.(storage.Extension).GetClient(...)
case isBeatsRegistry(ext):
    store, _ := ext.(backend.Registry).Access(...)
    return &adapter{store}, nil
default:
    return nil, fmt.Errorf("unsupported storage type %q", id)
}
```
but this seems to break the philosophy of OTEL receivers as being storage agnostic. If `elasticsearchstorage` implements **storage.Extension{}** the current receiver implementation would work seamlessly without any change/storage specific code.

ShourieG: I think the complexity comes in how storage extension and the beats statestore handle the actual data, where otel storage extension works with raw []byte, `elasticsearchstorage` works with `interface{}` and does custom encoding on the value. I had a brainstorm session with Claude and came up with a [plan](https://drive.google.com/file/d/1IJWmrdI5aWF1fgt1x0KnTygMJF0FqxAW/view?usp=sharing) on how to handle this in Beats.

We can also stick to the approach of breaking convention and having a storage based approach inside the receiver itself. I think that's the general call to take here. If you find the plan somewhat viable, I can create a PR in beats for it.

cmacknz:
I don't have to maintain this so it's ultimately your call on if you want the dual mode feature, I just wanted to make sure you had a confirmed use case for this because it is adding complexity and this is to my knowledge the first receiver that works this way. Everything else is either/or, not that this is an invalid thing to do on its own.

receiver/akamaisiemreceiver/internal/cursor/cursor.go
cmacknz: Thanks for taking an initial look, the elasticsearch storage extension should be compatible with any receiver we just didn't have that need when we originally adapted it to the storage interface and it was done with some time pressure.

Probably easier to start with a PR if that is easy, then we can see exactly what the changes need to be instead of discussing what they might be in an issue. CC @VihasMakwana who did the initial port of the Beat ES registry storage to the collector storage interface.

ShourieG:
> I don't have to maintain this so it's ultimately your call on if you want the dual mode feature, I just wanted to make sure you had a confirmed use case for this because it is adding complexity and this is to my knowledge the first receiver that works this way. Everything else is either/or, not that this is an invalid thing to do on its own.

@andrewkroh, what do you feel about the dual mode operation ? I thought it would be a good experimental feature to have and would be labeled as such and could open up new use cases down the line.

andrewkroh:
> @andrewkroh, what do you feel about the dual mode operation ? I thought it would be a good experimental feature to have and would be labeled as such and could open up new use cases down the line.

I would keep it as simple as it can be. So avoid the complexity of having an otel mode that we don't currently have a planned use for.

receiver/akamaisiemreceiver/BENCHMARKS.md
andrewkroh, line 1: I don't think we should commit these reports into the repo. 

Yes, keep the benchmarks test, but drop the report and supporting images. This data will get stale quickly.

ShourieG: ok

ShourieG: @andrewkroh, in this [commit](https://github.com/elastic/opentelemetry-collector-components/pull/1119/commits/c2efdb7bce878a4edf981ca595f949de969c59c0), I've refactored the PR by removing:

- Benchmarks.md and related charts from repo
- Removed dual mode operation
- Tested locally against local stack, both modes of operation separately using the mock api
- Updated README and related config descriptions.

These changes should simplify the PR for now and reduce upkeep.

ShourieG:
> I don't have to maintain this so it's ultimately your call on if you want the dual mode feature, I just wanted to make sure you had a confirmed use case for this because it is adding complexity and this is to my knowledge the first receiver that works this way. Everything else is either/or, not that this is an invalid thing to do on its own.

@cmacknz, just as an update after discussions with @andrewkroh, latest commits have removed dual mode operation for the moment and have simplified the receiver operation.

ShourieG:
@andrewkroh, @axw, @cmacknz  if this PR looks good atm, can we proceed with the approval ?

receiver/akamaisiemreceiver/receiver.go
```diff
@@ -0,0 +1,382 @@
...
+	case "otel":
+		return r.emitOTel(ctx, events)
+	default:
+		return r.emitRaw(ctx, events)
+	}
+}
+
+// emitRaw sends events as raw JSON body maps. Each log record body is a map
+// with key "message" containing the raw Akamai JSON string. The downstream
+// pipeline should set elastic.mapping.mode: bodymap via a transform processor
```
cmacknz: Why don't you set the mapping mode attribute here directly instead of requiring it to be set in a transform processor?

ShourieG: This is mainly to stick to the otel design philosophy. If we want to commit this to upstream, should the receiver explicitly define the mapping mode ? Won't that tightly couple the receiver behaviour to Elastic specific implementations by default ?

cmacknz: This doesn't follow any upstream convention, the bodymap mapping mode is Elastic specific and we made it for ourselves to allow transporting non-OTel data in the OTLP logs signal.

The JSON document is ECS is it now? This entire mode of execution (along with beats receivers) is vendor specific and that's on purpose. It lets us standardize on the collector as a data collection framework without having to migrate customers outside of Observability to OTLP+SemConv, which has no automatic solution.

ShourieG: The JSON document in `output_mode:raw` is the raw json event wrapped in the message field, it does not have any ecs specific mapping as of yet. The entire ecs mapping happens in the ingest pipelines. In otel mode we do the sem-cov mapping in the receiver itself. 

Because this JSON is the raw event without any modifications, in this receivers case, I thought it made sense for the mapping to be set using the transform processor. But since `bodymap` is an Elastic convention, it probably makes sense to tie it into the receiver itself. I will make this change and test it, please let me know if there are any other concerns regarding the current approach and I can address them in the next commit accordingly.

cmacknz: Yes it's less about ECS then, and more what structure you want the document to have coming out of the exporter. `bodymap` will send only the JSON event without any of the containing OTLP logs structure, without `bodymap` you'd send the OTLP log as is with no attributes or scope values set. 

People who want the OTLP structure will just use the otel mode. I don't think there's a use for wrapping sending the raw JSON event in an OTLP log record with none of the attributes set properly.

receiver/akamaisiemreceiver/config.go
```diff
@@ -0,0 +1,171 @@
...
+	defaultMaxRecovery      = 3
+	defaultInvalidTSRetry   = 2
+	defaultBatchSize        = 1000
+	defaultStreamBufferSize = 4
+)
+
+// Config defines the configuration for the Akamai SIEM receiver.
+type Config struct {
+	// Endpoint is the Akamai API host (e.g. "https://akab-xxx.luna.akamaiapis.net").
+	Endpoint string `mapstructure:"endpoint"`
```
andrewkroh: `Config.Endpoint` and `Config.HTTP.Endpoint` coexist. Only the standalone field is used in receiver.go.

Recommend dropping the standalone Endpoint and using c.HTTP.Endpoint everywhere. And update `Validate()` accordingly.

receiver/akamaisiemreceiver/config.go
```diff
@@ -0,0 +1,171 @@
...
+	// OutputFormat controls the structure of emitted log records.
+	// "raw" (default): raw Akamai JSON placed in LogRecord.Body as a map with key
+	// "message". Use a transform processor to set the elastic.mapping.mode scope
+	// attribute for Elasticsearch backends (see README for details).
+	// "otel": receiver parses each JSON event and maps 30+ fields to OTel semantic
+	// convention attributes on the LogRecord. Use for non-ES backends or structured querying.
+	OutputFormat string `mapstructure:"output_format"`
+
+	// HTTP provides TLS, proxy, timeout, and other HTTP client settings.
+	HTTP confighttp.ClientConfig `mapstructure:"http"`
```
andrewkroh: If a user sets `http.auth.authenticator: ...`, that authenticator's RoundTripper wraps the transport, and then `receiver.go` wraps EdgeGrid on top. Both run per request, and EdgeGrid's signature is computed before the other authenticator can mutate the request producing broken auth. Reject this in `Validate()`.

receiver/akamaisiemreceiver/internal/mapper/otel.go
```diff
@@ -0,0 +1,444 @@
...
+	var sb strings.Builder
+	sb.WriteString(msg.Host)
+	if msg.Path != "" {
+		sb.WriteString(msg.Path)
+	}
+	if msg.Query != "" {
+		sb.WriteString("?")
+		sb.WriteString(msg.Query)
+	}
+	attrs.PutStr("url.full", sb.String())
```
andrewkroh: Per OTel semconv, `url.full` is supposed to be a fully-qualified absolute URL. This one has no scheme and no port.

More broadly though, I think we should drop the SemConv-attribute mapping entirely and go bodymap-only.

It currently has `raw` and `otel` modes emitting structurally different documents from the same source, and I don't have a use case for the SemConv version. We want to send the data to the integration which already owns the ECS enrichment. Duplicating that in Go is drift-prone and adds a support burden.

Given the destination reality for this receiver, I think that's the right trade for now. If this changes and we move away from ingest pipeline-based processing, then we will revisit this. But today we are trying replace the existing CEL collector with something that can stream-process the response, and the additional event format is a bit beyond that scope.

receiver/akamaisiemreceiver/documentation.md
```diff
@@ -0,0 +1,151 @@
...
+
+Total HMAC timestamp retries
+
+| Unit | Metric Type | Value Type | Monotonic | Stability |
+| ---- | ----------- | ---------- | --------- | --------- |
+| {retries} | Sum | Int | true | Alpha |
+
+### otelcol_akamai_siem.mapping_duration
+
+Time spent mapping events to output format (OTel mode includes JSON parse + semconv mapping; ECS mode is raw copy)
```
andrewkroh: I think this ref to "ECS mode" is out of date. The config only accepts `raw` and `otel`.

receiver/akamaisiemreceiver/integration_test.go
```diff
@@ -0,0 +1,297 @@
...
+	// Test all severity levels: deny → ERROR, alert/tarpit → WARN, monitor → INFO
+	events := []struct {
+		action   string
+		wantSev  plog.SeverityNumber
+		wantText string
+	}{
+		// Raw mode doesn't set severity — it's raw JSON passthrough.
+		// Severity is only set in OTel mode.
+	}
+	_ = events
```
andrewkroh: Looks like dead code, `events` is declared and only `_ = events` is used. Either exercise the assertions or delete the test.

receiver/akamaisiemreceiver/receiver.go
```diff
@@ -0,0 +1,382 @@
...
+			return fmt.Errorf("failed to get storage client: %w", err)
+		}
+		cursorStore = cursor.NewCursorStore(storageClient)
+	}
+
+	// Load persisted cursor.
+	var cur cursor.Cursor
+	if cursorStore != nil {
+		var err error
+		cur, err = cursorStore.Load(context.Background())
```
andrewkroh: Why `context.Background()` instead of the `ctx` parameter? Using `ctx` would let a slow storage extension be cancelled by the collector startup timeout.

receiver/akamaisiemreceiver/internal/poller/poller.go
```diff
@@ -0,0 +1,538 @@
...
+		)
+
+	case !p.cursor.CaughtUp && p.cursor.ChainFrom != 0:
+		// Branch 2: Chain in progress but offset gone/stale — replay chain window.
+		if p.cursor.IsOffsetStale(p.cfg.OffsetTTL) {
+			p.log.Warn("offset stale, replaying chain window",
+				zap.Duration("offset_age", time.Since(p.cursor.OffsetObtainedAt)),
+				zap.Duration("ttl", p.cfg.OffsetTTL),
+			)
+			p.telemetry.addCounter(context.Background(), p.telemetry.OffsetTTLDrops, 1)
```
andrewkroh: We have a parent context, but we don't pass it in. Could we use that here instead of context.Background().

receiver/akamaisiemreceiver/internal/poller/poller.go
```diff
@@ -0,0 +1,538 @@
...
+		p.log.Debug("time-based fetch (new chain)",
+			zap.Int64("from", params.From),
+			zap.Int64("to", params.To),
+		)
+	}
+
+	return params
+}
+
+func (p *Poller) fetchWithTimestampRetry(ctx context.Context, params akamaiclient.FetchParams) (*readCloserWrapper, error) {
```
andrewkroh: What's the point of the `readCloserWrapper`? It looks like we could be using `io.ReadCloser` by itself.

receiver/akamaisiemreceiver/internal/mapper/otel.go
```diff
@@ -0,0 +1,444 @@
...
+	parts := strings.Split(unescaped, ";")
+	var result []string
+	for _, p := range parts {
+		p = strings.TrimSpace(p)
+		if p == "" {
+			continue
+		}
+		decoded, err := base64.StdEncoding.DecodeString(p)
+		if err != nil {
+			// Try URL-safe base64.
```
andrewkroh: The URL-safe variants are `URLEncoding`/`RawURLEncoding`, but it is using `RawStdEncoding`. The comment and the code disagree.

receiver/akamaisiemreceiver/internal/poller/poller.go
```diff
@@ -0,0 +1,538 @@
...
+			span.SetStatus(codes.Error, err.Error())
+		}
+		span.End()
+	}
+}
+
+const (
+	chainOverlap    = 10 * time.Second
+	maxLookback     = 12 * time.Hour
+	apiSafetyBuffer = 60 // seconds subtracted from "now" for the `to` parameter
```
andrewkroh: This comment should say _why_ 60s is the safety buffer or why there is a safety buffer at all.

ShourieG:
@andrewkroh, addressed all your suggestions and removed OTEL mode for now, this means we no longer require an explicit output_format anymore. 

@cmacknz, implemented your suggestions and coupled `bodymap` into the log scope in the receiver as suggested so no need for a separate transform processor anymore. 

I had claude re-write the commit history and clean it up for easy review. The config is much simpler now and hopefully it will be easier to maintain as an initial release.

receiver/akamaisiemreceiver/receiver.go
cmacknz: Thanks, the mapping LGTM

andrewkroh:
> I had claude re-write the commit history and clean it up for easy review.

That might help if someone is starting a fresh review. But for reviewers that already started it means re-reviewing everything rather than doing a differential review of changes since last review.

andrewkroh (approved):

ShourieG:
@axw , @cmacknz could I get a sign-off on the PR so it's approved to merge.

axw (approved):
LGTM. I've still only skimmed it.

PR #1151, opened 2026-04-07
[process/elasticapmprocessor] enable service language attribute processing for traces

aelnahas:
### Summary 

trace events currently do not populate the default service name for otlp ecs. This PR fixes that by enabling the enricher `DefaultServiceLanguage` flag

We need to set the field `telemetry.sdk.language` so that our [exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/elasticsearchexporter/model.go#L60) would map it to `service.language.name`

### Tests

- Updated processor tests and executed them

### DIFF
--- a/processor/elasticapmprocessor/processor.go
+++ b/processor/elasticapmprocessor/processor.go
@@ -70,6 +70,7 @@ func NewTraceProcessor(cfg *Config, next consumer.Traces, logger *zap.Logger) *T
 	intakeECSEnricherConfig.Transaction.Result.Enabled = false
 	// The `host.os.type` field should not be added for APM events
 	intakeECSEnricherConfig.Resource.HostOSType.Enabled = false
+	// disable default service language to avoid adding it on apm events
 	intakeECSEnricherConfig.Resource.DefaultServiceLanguage.Enabled = false
 
 	return &TraceProcessor{
@@ -179,6 +180,7 @@ func newLogProcessor(cfg *Config, next consumer.Logs, logger *zap.Logger) *LogPr
 
 	intakeECSEnricherConfig := ecsEnricherConfig
 	intakeECSEnricherConfig.Resource.HostOSType.Enabled = false
+	// disable default service language to avoid adding it on apm events
 	intakeECSEnricherConfig.Resource.DefaultServiceLanguage.Enabled = false
 	intakeECSEnricherConfig.Log.TranslateUnsupportedAttributes.Enabled = false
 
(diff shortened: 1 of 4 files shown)


processor/elasticapmprocessor/processor.go
```diff
@@ -69,6 +70,7 @@ func NewTraceProcessor(cfg *Config, next consumer.Traces, logger *zap.Logger) *T
 	intakeECSEnricherConfig.Transaction.Result.Enabled = false
 	// The `host.os.type` field should not be added for APM events
 	intakeECSEnricherConfig.Resource.HostOSType.Enabled = false
+	intakeECSEnricherConfig.Resource.DefaultServiceLanguage.Enabled = false
```
carsonip, line 74: sorry I don't have much context. What does it mean to enable it for ecs enricher but not intake ecs enricher?

carsonip (commented):
> populate the default service name for otlp ecs

q: isn't this PR around `telemetry.sdk.language`?

processor/elasticapmprocessor/processor.go
carsonip: is the effect of this PR to enable default service language for otlp ecs path but not intake ecs path?

aelnahas: Yes, I saw it was done for the log event and potentially for the metrics event in https://github.com/elastic/opentelemetry-collector-components/pull/1122 and followed the same pattern.

From what I understood from the code, I think this is done so that we do not set a default language for `apm` events. That's what I understood from intake receiver [mapping](https://github.com/elastic/opentelemetry-collector-components/blob/main/receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToSemConv.go#L36).

So in this context, we would only add the default service language for otlp ecs path.

aelnahas:
> > populate the default service name for otlp ecs
> 
> q: isn't this PR around `telemetry.sdk.language`?

Sorry I should have added that in the pr body

but we need to populate `telemetry.sdk.language` so that our [exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/elasticsearchexporter/model.go#L60) would map it to `service.language.name`

processor/elasticapmprocessor/processor.go
aelnahas: I can add a comment so it is clearer for the next reader

lanre-ade: We can consider dedicated methods to build each class of configs (otlp ecs vs intake ecs) if the drift gets more significant

carsonip:
> but we need to populate telemetry.sdk.language so that our o that our [exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/elasticsearchexporter/model.go#L60) would map it to service.language.name

I see, so it is about service language name, not service name.

carsonip (approved):

PR #1153, opened 2026-04-08
receiver/entityanalyticsreceiver: add entity analytics receiver

efd6:
Add a new OTel Collector receiver that syncs identity data from external providers (Okta, EntraID, Active Directory, Jamf) via the entcollect shared library. The receiver has no dependency on beats; it resolves a storage extension at runtime and type-asserts it to entcollect.Registry.

The receiver runs a timer-based sync loop (full + incremental), converts each entcollect.Document to a plog.LogRecord, and publishes per-document to bound memory during large syncs. The provider registry is empty; concrete providers will be added in subsequent sessions via blank imports.

Includes config validation, extension resolution, publisher mapping, 7 tests with fake provider and in-memory store, distribution manifest entry, and CODEOWNERS for security-service-integrations.

ref elastic/beats#49159


receiver/entityanalyticsreceiver/provider.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// Register adds a named provider factory. Providers call this from
+// init() so that blank-importing the provider package is sufficient
+// to make it available.
+func Register(name string, f ProviderFactory) {
+	mu.Lock()
+	defer mu.Unlock()
+	factories[name] = f
+}
```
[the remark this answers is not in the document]
efd6: PTAL

receiver/entityanalyticsreceiver/go.mod
```diff
@@ -0,0 +1,59 @@
...
+	go.opentelemetry.io/collector/consumer/consumererror v0.149.0 // indirect
+	go.opentelemetry.io/collector/consumer/xconsumer v0.149.0 // indirect
+	go.opentelemetry.io/collector/featuregate v1.55.0 // indirect
+	go.opentelemetry.io/collector/internal/componentalias v0.149.0 // indirect
+	go.opentelemetry.io/collector/pdata/pprofile v0.149.0 // indirect
+	go.opentelemetry.io/collector/pipeline v1.55.0 // indirect
+	go.opentelemetry.io/collector/receiver/xreceiver v0.149.0 // indirect
+	go.opentelemetry.io/otel v1.42.0 // indirect
+	go.opentelemetry.io/otel/metric v1.42.0 // indirect
+	go.opentelemetry.io/otel/sdk v1.42.0 // indirect
```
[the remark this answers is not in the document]
efd6: Reassess.
```
$ govulncheck -show verbose ./...
Fetching vulnerabilities from the database...

Checking the code against the vulnerabilities...

The package pattern matched the following 2 root packages:
  github.com/elastic/opentelemetry-collector-components/receiver/entityanalyticsreceiver/internal/metadata
  github.com/elastic/opentelemetry-collector-components/receiver/entityanalyticsreceiver
Govulncheck scanned the following 29 modules and the go1.25.8 standard library:
  github.com/elastic/opentelemetry-collector-components/receiver/entityanalyticsreceiver
  github.com/cespare/xxhash/v2@v2.3.0
  github.com/elastic/entcollect@v0.0.0-20260408214411-765a9bedd20c
  github.com/go-viper/mapstructure/v2@v2.5.0
  github.com/gobwas/glob@v0.2.3
  github.com/hashicorp/go-version@v1.8.0
  github.com/json-iterator/go@v1.1.12
  github.com/knadh/koanf/maps@v0.1.2
  github.com/knadh/koanf/providers/confmap@v1.0.0
  github.com/knadh/koanf/v2@v2.3.4
  github.com/mitchellh/copystructure@v1.2.0
  github.com/mitchellh/reflectwalk@v1.0.2
  github.com/modern-go/concurrent@v0.0.0-20180306012644-bacd9c7ef1dd
  github.com/modern-go/reflect2@v1.0.3-0.20250322232337-35a7c28c31ee
  go.opentelemetry.io/collector/component@v1.55.0
  go.opentelemetry.io/collector/confmap@v1.55.0
  go.opentelemetry.io/collector/consumer@v1.55.0
  go.opentelemetry.io/collector/featuregate@v1.55.0
  go.opentelemetry.io/collector/internal/componentalias@v0.149.0
  go.opentelemetry.io/collector/pdata@v1.55.0
  go.opentelemetry.io/collector/pipeline@v1.55.0
  go.opentelemetry.io/collector/receiver@v1.55.0
  go.opentelemetry.io/otel@v1.42.0
  go.opentelemetry.io/otel/metric@v1.42.0
  go.opentelemetry.io/otel/trace@v1.42.0
  go.uber.org/multierr@v1.11.0
  go.uber.org/zap@v1.27.1
  go.uber.org/zap/exp@v0.3.0
  go.yaml.in/yaml/v3@v3.0.4

=== Symbol Results ===

No vulnerabilities found.

=== Package Results ===

Vulnerability #1: GO-2026-4947
    Unexpected work during chain building in crypto/x509
  More info: https://pkg.go.dev/vuln/GO-2026-4947
  Standard library
    Found in: crypto/x509@go1.25.8
    Fixed in: crypto/x509@go1.25.9

Vulnerability #2: GO-2026-4946
    Inefficient policy validation in crypto/x509
  More info: https://pkg.go.dev/vuln/GO-2026-4946
  Standard library
    Found in: crypto/x509@go1.25.8
    Fixed in: crypto/x509@go1.25.9

Vulnerability #3: GO-2026-4870
    Unauthenticated TLS 1.3 KeyUpdate record can cause persistent connection
    retention and DoS in crypto/tls
  More info: https://pkg.go.dev/vuln/GO-2026-4870
  Standard library
    Found in: crypto/tls@go1.25.8
    Fixed in: crypto/tls@go1.25.9

Vulnerability #4: GO-2026-4864
    TOCTOU permits root escape on Linux via Root.Chmod in os in
    internal/syscall/unix
  More info: https://pkg.go.dev/vuln/GO-2026-4864
  Standard library
    Found in: internal/syscall/unix@go1.25.8
    Fixed in: internal/syscall/unix@go1.25.9
    Platforms: linux

=== Module Results ===

Vulnerability #1: GO-2026-4869
    Unbounded allocation for old GNU sparse in archive/tar
  More info: https://pkg.go.dev/vuln/GO-2026-4869
  Standard library
    Found in: stdlib@go1.25.8
    Fixed in: stdlib@go1.25.9

Vulnerability #2: GO-2026-4865
    JsBraceDepth Context Tracking Bugs (XSS) in html/template
  More info: https://pkg.go.dev/vuln/GO-2026-4865
  Standard library
    Found in: stdlib@go1.25.8
    Fixed in: stdlib@go1.25.9

Your code is affected by 0 vulnerabilities.
This scan also found 4 vulnerabilities in packages you import and 2
vulnerabilities in modules you require, but your code doesn't appear to call
these vulnerabilities.
```

axw:
@efd6 did you consider using [scraperhelper](https://github.com/open-telemetry/opentelemetry-collector/tree/main/scraper/scraperhelper) for triggering scrapes? At the moment it's purely polling-based, but FYI that will soon be extended to additional methods -- see https://github.com/open-telemetry/opentelemetry-collector/pull/14469. Not sure if that's relevant for this receiver.

efd6:
Thanks @axw, unfortunately not: there are two independent intervals with different operations; the sync is transactional, writing into a state buffer during potentially long-running syncs and then being committed; we're collating logs, not emitting each.

axw:
So something like this?
 - Every X seconds, update local state
 - Every Y seconds, publish local state

IIANM that could still be done with a scraper receiver: the "scrape" would be publishing local state, and you would have a secondary poll loop for updating it. There may be limited value in this though.

efd6:
Every 24h/every 15m, but essentially yes. It feels like the approach would add complexity.

axw:
For posterity, Dan and I chatted about this on Slack. Here's a summary:
 - What I described in my comment isn't right. We don't gather local state and publish it. Rather there are two timers, one is for a full sync, the other for incremental syncs.
 - Technically we could enforce that the incremental sync interval divides the full sync interval, and decide based on the trigger time whether it's a incremental sync vs. full sync (if T%full==0). Even so, there are problems with the scraperhelper contract:
     - The receiver may publish very large volumes of logs per sync, which requires streaming logs out to avoid memory exhaustion. The scraperhelper contract means that a scraper implementation must return one big plog.Logs batch, which prevents streaming logs out.
     - The receiver wants to commit cursors. This should happen after the logs are consumed for guaranteed delivery.
 - The interface I've proposed in the RFC could be used directly by the receiver, without having to buy into the full scraperhelper contract. This would address the two issues mentioned above.

receiver/entityanalyticsreceiver/publisher.go
```diff
@@ -0,0 +1,77 @@
...
+		sl.Scope().SetVersion(scopeVersion)
+		lr := sl.LogRecords().AppendEmpty()
+		lr.SetTimestamp(pcommon.NewTimestampFromTime(doc.Timestamp))
+
+		lr.Attributes().PutStr("event.action", doc.Action.String())
+		kind := "asset"
+		if doc.Action == entcollect.ActionDeleted {
+			kind = "event"
+		}
+		lr.Attributes().PutStr("event.kind", kind)
```
cmacknz: This looks like the ECS [event.kind](https://www.elastic.co/docs/reference/ecs/ecs-event#field-event-kind), if you want an ECS document the convention is to put the entire ECS JSON document into the log record body and set the attribute indicating it should use the bodymap mapping mode.

This is how beats receivers does it https://github.com/elastic/beats/blob/c3dc5af138a777020b4103f31389596bd2d91b5f/libbeat/otel/otelconsumer/otelconsumer.go#L111-L112

Otherwise what you'll end up with is a document that uses the OTLP data model but follows the ECS schema. This would cause customers who use ECS to be exposed to the OTLP model which we are actively trying to avoid and haven't done anywhere else.

cmacknz: Thanks, LGTM. There's no README for the component but you probably want to document that this receiver outputs ECS somewhere to avoid surprising anybody that tries to use it from EDOT collector on their own.

receiver/entityanalyticsreceiver/provider.go
```diff
@@ -0,0 +1,78 @@
...
+	"go.opentelemetry.io/collector/confmap"
+
+	"github.com/elastic/entcollect"
+)
+
+// ProviderFactory creates an entcollect.Provider from receiver
+// configuration. Concrete providers (Okta, EntraID, etc.) register
+// their factory at init time via blank imports; the receiver looks
+// up the factory by name at Start.
+type ProviderFactory func(cfg *confmap.Conf) (entcollect.Provider, error)
```
axw, line 32: Would there be any value in implementing providers as extensions? i.e. by having extensions that implement the `entcollect.Provider` interface. Then the providers could be implemented without a dependency on the receiver.

It would also give you a fairly natural way to configure each provider (i.e. in the provider-specific extension config), if that's something that needs doing.

Maybe it's overkill though.

axw (approved):
LGTM, just a question about whether you considered providers-as-extensions

receiver/entityanalyticsreceiver/provider.go
efd6: Entity Analytics providers aren't standalone, they only make sense driven by this receiver's sync loop while extensions have their own lifecycle (Start/Shutdown), which doesn't map to the provider contract.

Making providers be extensions would mean the receiver discovers its provider via type assertion at runtime (a second runtime coupling alongside the storage extension), and the extension's life-cycle methods would be empty since the provider doesn't run independently; the contract we have is that the receiver calls `FullSync` or `IncrementalSync` on the configured schedule.

The current pattern avoids a compile dependency between provider and receiver. Providers register a factory at init time, and the `ProviderFactory` takes `*confmap.Conf` so that provider-specific configuration can be passed through without the receiver knowing the shape. It's the same decoupling you'd get from extensions, just without the extra life-cycle machinery.

axw: > The current pattern avoids a compile dependency between provider and receiver. Providers register a factory at init time

Isn't that a compile time dependency?

> and the ProviderFactory takes *confmap.Conf so that provider-specific configuration can be passed through without the receiver knowing the shape. It's the same decoupling you'd get from extensions, just without the extra life-cycle machinery.

:+1:

Anyway, it's fine as it is, just wanted to check if you had considered it.

efd6: Yeah, sorry, that was unclear. The issue is not the "compile time" part, but the direction. The receiver→provider dep is avoided, we have a provider→receiver that does exist, but this is part of the overall Entity Analytics design already.

axw: I see what you mean now, thanks for clarifying.

receiver/entityanalyticsreceiver/publisher.go
```diff
@@ -0,0 +1,81 @@
...
+		body["event.action"] = doc.Action.String()
+		body["event.kind"] = kind
+		body["asset.type"] = doc.Kind.String()
+		body["asset.id"] = doc.ID
+		body["labels.identity_source"] = provider
+
+		if err := lr.Body().SetEmptyMap().FromRaw(body); err != nil {
+			return fmt.Errorf("encoding document body: %w", err)
+		}
+		return cons.ConsumeLogs(ctx, logs)
```
andrewkroh, line 79: When this returns nil (happy path) does that typically mean the data has landed in some durable storage (been sent by the exporters successfully)?

andrewkroh (approved):
Just a non-blocking question about otel data pipeline behavior.

receiver/entityanalyticsreceiver/publisher.go
efd6: This is documented in the [comment on `Start`](https://github.com/elastic/opentelemetry-collector-components/pull/1153/changes#diff-32a3a863a2ee85124e3f8671f58b36ef0a76ec8fa0437f830e01010cfc64d428R60-R68); briefly, with the default agentless pipeline, acceptance and durable write coincide. If an async exporter or batch processor is in the pipeline, nil only means the batch was accepted into the queue, not that it was durably written. In that case there's a replay window bounded by the full sync interval; deterministic `_id` makes re-indexing safe.

PR #1155, opened 2026-04-09
Handle representative count and composite spans for intake-v2

lahsivjar:
- Encode `representative_count` into W3C tracestate T-value in the intake receiver so that `AdjustedCount()` in the signal-to-metrics connector returns the correct sampling weight
  - Handle composite spans in the connector by reading `span.composite.count` and `span.composite.sum.us` attributes for span destination metrics instead of using span wall-clock duration and default count of 1
 
Closes: https://github.com/elastic/hosted-otel-collector/issues/2710

### DIFF
--- a/connector/elasticapmconnector/config.go
+++ b/connector/elasticapmconnector/config.go
@@ -389,19 +389,54 @@ func (cfg Config) signaltometricsConfig() *signaltometricsconfig.Config {
 			Unit:      "us",
 			Histogram: configoptional.Some(transactionDurationSummaryHistogram),
 		}, {
+			// For composite spans, the response time sum is the composite
+			// duration (total of all compressed sub-spans).
 			Name:                      "span.destination.service.response_time.sum.us",
 			Description:               "APM span destination metrics",
 			IncludeResourceAttributes: spanDestinationResourceAttributes,
 			Attributes:                spanDestinationAttributes,
-			Unit:                      "us",
+			Conditions: []string{
+				`attributes["span.composite.sum.us"] != nil`,
+			},
+			Unit: "us",
+			Sum: configoptional.Some(signaltometricsconfig.Sum{
+				Value: `Double(attributes["span.composite.sum.us"])`,
+			}),
+		}, {
+			// For non-composite spans, use the span wall-clock duration.
+			Name:                      "span.destination.service.response_time.sum.us",
+			Description:               "APM span destination metrics",
+			IncludeResourceAttributes: spanDestinationResourceAttributes,
+			Attributes:                spanDestinationAttributes,
+			Conditions: []string{
+				`attributes["span.composite.sum.us"] == nil`,
+			},
+			Unit: "us",
 			Sum: configoptional.Some(signaltometricsconfig.Sum{
 				Value: "Double(Microseconds(end_time - start_time))",
 			}),
 		}, {
+			// For composite spans, the count is the number of compressed
+			// operations.
 			Name:                      "span.destination.service.response_time.count",
 			Description:               "APM span destination metrics",
 			IncludeResourceAttributes: spanDestinationResourceAttributes,
 			Attributes:                spanDestinationAttributes,
+			Conditions: []string{
+				`attributes["span.composite.count"] != nil`,
+			},
+			Sum: configoptional.Some(signaltometricsconfig.Sum{
+				Value: `Int(attributes["span.composite.count"])`,
+			}),
+		}, {
+			// For non-composite spans, count is the sampling weight.
+			Name:                      "span.destination.service.response_time.count",
+			Description:               "APM span destination metrics",
+			IncludeResourceAttributes: spanDestinationResourceAttributes,
+			Attributes:                spanDestinationAttributes,
+			Conditions: []string{
+				`attributes["span.composite.count"] == nil`,
+			},
 			Sum: configoptional.Some(signaltometricsconfig.Sum{
 				Value: "Int(AdjustedCount())",
 			}),
--- a/receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
+++ b/receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
@@ -121,7 +121,15 @@ func probabilityToTValue(probability float64) string {
 	if probability == 1 {
 		return "0"
 	}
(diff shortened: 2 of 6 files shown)


receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
```diff
@@ -90,8 +93,61 @@ func SetTopLevelFieldsSpan(event *modelpb.APMEvent, timestampNanos uint64, s ptr
 	}
 
 	durationNanos := event.GetEvent().GetDuration()
+
+	// For composite spans, use the composite sum as the effective duration
+	// so that downstream aggregation (end_time - start_time) reflects the
+	// total duration of all compressed sub-spans.
+	if composite := event.GetSpan().GetComposite(); composite != nil {
+		compositeSumNanos := uint64(composite.GetSum() * float64(time.Millisecond))
+		durationNanos = compositeSumNanos
```
carsonip: q: should we take max of the two, or simply override the span's duration?

carsonip (approved):
thanks

receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
```diff
@@ -90,8 +93,61 @@ func SetTopLevelFieldsSpan(event *modelpb.APMEvent, timestampNanos uint64, s ptr
...
+	}
```
lahsivjar: I realized that this handling of composite spans is not correct - this handling needs to happen in the aggregation layer.

connector/elasticapmconnector/config.go
```diff
@@ -389,19 +389,54 @@ func (cfg Config) signaltometricsConfig() *signaltometricsconfig.Config {
...
 			Description:               "APM span destination metrics",
 			IncludeResourceAttributes: spanDestinationResourceAttributes,
 			Attributes:                spanDestinationAttributes,
-			Unit:                      "us",
+			Conditions: []string{
+				`attributes["span.composite.sum.us"] != nil`,
+			},
+			Unit: "us",
+			Sum: configoptional.Some(signaltometricsconfig.Sum{
+				Value: `Double(attributes["span.composite.sum.us"])`,
```
lahsivjar, line 403: [**For reviewers**] In APM, we multiply this by representative count ([ref](https://github.com/elastic/apm-aggregation/blob/13d584260c7bc536224878fa1f390b9d5edfb96c/aggregators/converter.go#L553-L560)), so the equivalent of this should be to multiply by `AdjustedCount`. I am not sure if this was a miss on our part or it was intentional. @axw do you remember anything around this topic?

receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
lahsivjar: This is fixed now.

receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
```diff
@@ -92,6 +94,48 @@ func SetTopLevelFieldsSpan(event *modelpb.APMEvent, timestampNanos uint64, s ptr
...
+	// Encode representative_count into the W3C tracestate so that
+	// AdjustedCount() in the signal-to-metrics connector returns the
+	// correct sampling weight. We only set the tracestate when the
+	// representative count differs from 1 (the AdjustedCount default)
+	// to avoid noisy tracestate on every span.
+	if span := event.GetSpan(); span != nil {
+		repCount := span.GetRepresentativeCount()
+		if repCount > 0 && repCount != 1 {
+			if tvalue := probabilityToTValue(1.0 / repCount); tvalue != "" {
+				s.TraceState().FromRaw(fmt.Sprintf("ot=th:%s", tvalue))
```
vigneshshanmugam, line 108: can you add link so `ot and th` is clear for everyone. I belive it refers to otel threshold trace state from what I see here - https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/sampling/oteltracestate.go

vigneshshanmugam (approved):
LGTM

connector/elasticapmconnector/config.go
axw: I don't recall, but intuitively I think it makes sense to multiple by the adjusted count. A composite span may also have been subject to sampling, right?

lahsivjar: >  A composite span may also have been subject to sampling, right?

Yes, and APM Server also does this. I will create a separate issue for this since it is also a bug in mOTLP routes.

lahsivjar: > https://github.com/elastic/opentelemetry-collector-components/issues/1158

Here is the issue.

receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
```diff
@@ -92,6 +94,48 @@ func SetTopLevelFieldsSpan(event *modelpb.APMEvent, timestampNanos uint64, s ptr
 	durationNanos := event.GetEvent().GetDuration()
 	s.SetStartTimestamp(pcommon.Timestamp(timestampNanos))
 	s.SetEndTimestamp(pcommon.Timestamp(timestampNanos + durationNanos))
+
+	// Encode representative_count into the W3C tracestate so that
+	// AdjustedCount() in the signal-to-metrics connector returns the
+	// correct sampling weight. We only set the tracestate when the
+	// representative count differs from 1 (the AdjustedCount default)
+	// to avoid noisy tracestate on every span.
```
lahsivjar, line 102:
```suggestion
	// to avoid noisy tracestate on every span.
	// see: https://opentelemetry.io/docs/specs/otel/trace/tracestate-handling/#sampling-threshold-value-th
```

receiver/elasticapmintakereceiver/internal/mappers/intakeV2ToOtlpTopLevelFields.go
lahsivjar: Done!

PR #1159, opened 2026-04-10
[intakereceiver]Fix global labels shadowing

lahsivjar:
## Fix global label shadowing in aggregated metrics

**Fixes:** elastic/hosted-otel-collector#2865

### Problem

When an event-level tag shadows a metadata (global) label, the receiver includes the shadowed key in `x-elastic-dynamic-resource-attributes` for all events in the batch. The downstream connector then uses this key for aggregation grouping on the shadowing event, diverging from `apm-aggregation` which excludes shadowed keys per-event via `marshalEventGlobalLabels()`.

### Solution

During batch processing, detect events that shadow a global label and separate them into dedicated batches, each consumed with its own client metadata context containing only that event's effective global keys.

- First pass collects the batch-level global key set into a `keyIndex` map with bit positions for efficient grouping
- Second pass checks each event for shadowing; non-shadowed events stay in the main batch
- Shadowed events are grouped by a `big.Int` bitmask of their effective global keys, so events with identical shadow patterns share a batch
- Pdata structures (`Logs`, `Metrics`, `Traces`) are lazily initialized on first use, avoiding allocations for unused signal types

### Performance

Benchmarked against main (`-benchtime=5s -count=6`, `benchstat`):

| Scenario | sec/op | allocs/op |
|----------|--------|-----------|
| No shadowing (common) | ~ (no change) | **-2** (894 vs 896) |
| With shadowing (rare) | ~ (no change) | +43 (959 vs 916) |

The no-shadow path is allocation-equivalent to (or better than) main. The shadow path adds allocations proportional to the number of distinct shadow patterns, which is expected and rare in production.

### DIFF
--- a/receiver/elasticapmintakereceiver/receiver.go
+++ b/receiver/elasticapmintakereceiver/receiver.go
@@ -261,7 +261,7 @@ func (r *elasticAPMIntakeReceiver) processBatch(ctx context.Context, batch *mode
 		for key, lv := range event.Labels {
 			if lv != nil && lv.Global {
 				if _, ok := keyIndex[key]; !ok {
-					keyIndex[key] = globalKeyInfo{bitPos: bitPos, isNumeric: false}
+					keyIndex[key] = globalKeyInfo{bitPos: bitPos, prefixedKey: "labels." + key}
 					bitPos++
 				}
 			}
@@ -269,7 +269,7 @@ func (r *elasticAPMIntakeReceiver) processBatch(ctx context.Context, batch *mode
 		for key, nv := range event.NumericLabels {
 			if nv != nil && nv.Global {
 				if _, ok := keyIndex[key]; !ok {
-					keyIndex[key] = globalKeyInfo{bitPos: bitPos, isNumeric: true}
+					keyIndex[key] = globalKeyInfo{bitPos: bitPos, prefixedKey: "numeric_labels." + key}
 					bitPos++
 				}
 			}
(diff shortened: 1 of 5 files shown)


receiver/elasticapmintakereceiver/receiver.go
```diff
@@ -242,112 +243,186 @@ func (r *elasticAPMIntakeReceiver) processBatch(ctx context.Context, batch *mode
...
+	// its own context so that the shadowed key is excluded from
+	// "x-elastic-dynamic-resource-attributes" for those events.
+	keyIndex := make(map[string]globalKeyInfo)
+	var bitPos int
 	for _, event := range *batch {
 		for key, lv := range event.Labels {
 			if lv != nil && lv.Global {
-				labelKeySet[key] = struct{}{}
+				if _, ok := keyIndex[key]; !ok {
+					keyIndex[key] = globalKeyInfo{bitPos: bitPos, isNumeric: false}
```
carsonip: nit: will we save on allocs if we store the `labels.` or `numeric_labels.` prefixed version in a field `prefixedKey` instead of a bool `isNumeric` and later reconstruct the prefixed field?

carsonip (approved):
lgtm 1 nit

receiver/elasticapmintakereceiver/receiver.go
lahsivjar: Hmm, good idea. We should be able to save a few allocs.

lahsivjar: Done!

carsonip: did we save on allocs?

carsonip (approved):

PR #1163, opened 2026-04-13
[processor/elasticapmprocessor] Fix otlp span discrepancies

lanre-ade:
## Summary
Align ECS OTLP span handling in `elasticapmprocessor` more closely with `apm-data` by translating unsupported span attrs to `labels.*` / `numeric_labels.*`, improving HTTP destination and URL derivation, and preserving the right enrichment inputs during ECS processing.

## What changed
- add ECS span attribute fallback translation for OTLP spans
- derive HTTP destination fields, `url.original`, and default URL schemes more consistently
- preserve `peer.service` for `span.destination.service.*` while keeping URL-derived `service.target.*` parity
- support legacy `http.host` and `http.user_agent` inputs in the ECS span path

Note: branched off https://github.com/elastic/opentelemetry-collector-components/pull/1122 so carried over some unrelated commit history 

## Test plan
  - `go test ./...` from `processor/elasticapmprocessor`


processor/elasticapmprocessor/internal/enrichments/config/config.go
```diff
@@ -87,19 +87,20 @@ type ElasticSpanConfig struct {
...
+	TypeSubtype                    AttributeConfig `mapstructure:"type_subtype"`
+	DurationUs                     AttributeConfig `mapstructure:"duration_us"`
+	EventOutcome                   AttributeConfig `mapstructure:"event_outcome"`
+	ServiceTarget                  AttributeConfig `mapstructure:"service_target"`
+	DestinationService             AttributeConfig `mapstructure:"destination_service"`
+	InferredSpans                  AttributeConfig `mapstructure:"inferred_spans"`
+	UserAgent                      AttributeConfig `mapstructure:"user_agent"`
+	RemoveMessaging                AttributeConfig `mapstructure:"remove_messaging"`
+	MessageQueueName               AttributeConfig `mapstructure:"message_queue_name"`
+	TranslateUnsupportedAttributes AttributeConfig `mapstructure:"translate_unsupported_attributes"`
```
lahsivjar, line 103: I am not sure about these configs, I think we can be a bit opinionated about how translation works because that is the whole purpose of this processor. I have started a slack thread to discuss this: https://elastic.slack.com/archives/C0AED3Y25PU/p1776158286258689

That being said, this is NOT a blocker and PR can be merged without resolving this.

processor/elasticapmprocessor/internal/enrichments/enricher.go
```diff
@@ -52,16 +52,22 @@ type Enricher struct {
...
-				EnrichSpan(spans.At(k), e.Config, e.userAgentParser)
-			}
+		e.EnrichResourceSpans(resSpans.At(i))
+	}
+}
+
+// EnrichResourceSpans enriches a single ResourceSpans with the current enricher
+// configuration. This is used when ECS batches need per-resource enrichment
+// policies based on the origin of each ResourceSpans.
+func (e *Enricher) EnrichResourceSpans(resSpan ptrace.ResourceSpans) {
```
lahsivjar, line 62: **nit**: feel free to ignore, is there any advantage of this being a separate method? For me, it is causing more confusion.

lahsivjar (commented):
Have reviewed it briefly but I don't think I will be able to review this completely today - don't get blocked on my review if you get enough to merge.

Can you also run benchmarks to compare how apm processor is doing now vs main?

processor/elasticapmprocessor/internal/enrichments/enricher.go
lanre-ade: The main advantage is that we can handle a mixed batch of intake and non intake signals in ecs mode with a per resource method. 

I also thought this would help with simplification since I found how we choose the enricher for a batch for logs confusing (open question [here](https://github.com/elastic/opentelemetry-collector-components/pull/1107#discussion_r2967828956)) and the coderabbit bot called out a similar issue for potential mixed ecs batch with [my PR for metrics](https://github.com/elastic/opentelemetry-collector-components/pull/1122#discussion_r3025058817). 


I have a similar refactor for metrics and logs in https://github.com/elastic/opentelemetry-collector-components/pull/1164

lanre-ade: The method is used [here](https://github.com/elastic/opentelemetry-collector-components/blob/44d9c704c2830165fe88f7b64e3adcf9c463307a/processor/elasticapmprocessor/processor.go#L124) after the enricher is [selected](https://github.com/elastic/opentelemetry-collector-components/blob/44d9c704c2830165fe88f7b64e3adcf9c463307a/processor/elasticapmprocessor/processor.go#L98)

processor/elasticapmprocessor/internal/enrichments/config/config.go
lanre-ade: I agree, it would bring more clarity the ecs paths. I can clean them up in a follow up PR

processor/elasticapmprocessor/internal/enrichments/span.go
```diff
@@ -628,7 +683,123 @@ func (s *spanEnrichmentContext) setDestinationService(span ptrace.Span) {
...
+		}
+	}
+
+	details.serviceTargetName = resource
+	details.spanDestinationServiceName = urlForName.String()
+	details.spanDestinationServiceResource = resource
+	return details, true
+}
+
+func (s *spanEnrichmentContext) buildURLFromComponents() *url.URL {
```
aelnahas, line 803: I am not sure if I completely understand it but from what I have been reading in `apm-data` it seems like https://github.com/elastic/apm-data/blob/26adeeef7f92ba5e01e59fb9e4c735fb8c31b58e/input/otlp/traces.go#L848 there is a fall back on `net.peer.port` / `peer.port`, `net.peer.name`, `peer.address`, but that logic seems missing here

processor/elasticapmprocessor/internal/enrichments/span.go
```diff
@@ -219,8 +248,13 @@ func (s *spanEnrichmentContext) Enrich(
 		case string(semconv27.GenAISystemKey), string(semconv37.GenAIProviderNameKey):
 			s.isGenAi = true
 			s.genAiSystem = v.Str()
+		case string(semconv25.HTTPUserAgentKey):
+			if !s.userAgentOriginalSet {
+				s.userAgentOriginal = v.Str()
+			}
 		case string(semconv27.UserAgentOriginalKey):
 			s.userAgentOriginal = v.Str()
+			s.userAgentOriginalSet = true
```
aelnahas, line 294: ~~I think we may also want to handle [destination.service.resource](https://github.com/elastic/apm-data/blob/26adeeef7f92ba5e01e59fb9e4c735fb8c31b58e/input/otlp/traces.go#L902), I get the feeling it is being treated as a custom label right but would be worth testing this out~~
mis read the code its `span.destination.service.resource` which is already been handled here

aelnahas: Also it seems like in `apm-data` https://github.com/elastic/apm-data/blob/26adeeef7f92ba5e01e59fb9e4c735fb8c31b58e/input/otlp/traces.go#L816 `span.kind` is dropped / skipped . I might be misreading the code but I think this will end up being populated as a custom label `label.span_kind`?

aelnahas: Do you know if we also need to populate `message.queue.name`? I noticed that in `apm-data` we populate fields like https://github.com/elastic/apm-data/blob/26adeeef7f92ba5e01e59fb9e4c735fb8c31b58e/input/otlp/traces.go#L431 and also :
- `message_bus.destination`
- `messaging.operation.type`

I think I see them in translation key but just maybe it's worth having a processor test to confirm it is being processed correctly.

aelnahas (commented):
it is looking good, I had some questions / comments

processor/elasticapmprocessor/internal/enrichments/enricher.go
lahsivjar: Discussed this on slack (https://elastic.slack.com/archives/C076MUD9BK8/p1776200800130239), resolving this as we will do a future refactor to simplify the logic.

processor/elasticapmprocessor/internal/enrichments/span.go
lahsivjar: `net.peer.*` have been deprecated since semconv v1.12 and `peer.*` are even older. I would be in favor of not handling them.

aelnahas (approved):

processor/elasticapmprocessor/internal/enrichments/span.go
lahsivjar: @aelnahas

> I think we may also want to handle [destination.service.resource](https://github.com/elastic/apm-data/blob/26adeeef7f92ba5e01e59fb9e4c735fb8c31b58e/input/otlp/traces.go#L902),

I am not sure I understand this point, it is not an attribute. Can you explain a bit more?

aelnahas: I misread the code I thought there should be an attributed called `destination.service.resource` but actually its `span.destinat.service.resource` . This PR already addresses this.

lanre-ade: @aelnahas regarding your question about `message.queue.name`, the [current implementation](https://github.com/elastic/opentelemetry-collector-components/blob/90375cf3be573371ba24e13ac885b5e1606c7264/processor/elasticapmprocessor/internal/enrichments/span.go#L525) uses the `MessagingDestinationNameKey` (and related semconv attributes) to set the corresponding queue name for transactions and spans in `setMessageQueue`.

lanre-ade:
Closing this in favour of https://github.com/elastic/opentelemetry-collector-components/pull/1168

PR #1164, opened 2026-04-14
[processor/elasticapmprocessor] refactor: use per-resource ECS enricher selection for logs and metrics

lanre-ade:
### Summary

Refactors to the enrichment structure
- Adds two new methods, `EnrichResourceLogs` and `EnrichResourceMetrics`, to encapsulate per-resource enrichment logic. 
- The `EnrichLogs` and `EnrichMetrics` methods now delegate to these helpers rather than performing enrichment inline. 
- The log and metric processors were updated to select enrichers on a per-resource basis instead of batch-wide selection based on the first resource. 
- Tests were revised to validate correct per-resource enrichment behavior in mixed-origin batches.


lanre-ade:
Closed in favour of https://github.com/elastic/opentelemetry-collector-components/pull/1168

PR #1166, opened 2026-04-15
Add version check script to validate tags

### LINKED ISSUE #952: Sanity check for "push-tags" tag version
- Ensure the tag specified during the `make push-tags` is valid (highest available tag among components)
- Extend the release documentation to remind the release manager to send a release notification message in #otel-dev

jmmcorreia:
Create a script that validates that the new tag is greater than the highest tag currently set to any of the modules.

The script is run during `make push-tags` to ensure the new version is actually the highest version. 

However, as of now, it does not check if the provided version is a minor increment over the current highest as described here: https://github.com/elastic/opentelemetry-collector-components/blob/main/docs/release.md#create-the-new-tags 

- Related to issue #952

### DIFF
--- a/internal/versionscheck/main.go
+++ b/internal/versionscheck/main.go
@@ -19,18 +19,19 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"log"
+	"net/http"
 	"os"
-	"os/exec"
 	"strings"
+	"time"
 
+	"github.com/google/go-github/v84/github"
 	"golang.org/x/mod/semver"
 	"gopkg.in/yaml.v3"
 )
-
-const edotBaseKey = "edot-base"
 
 type versionsYAML struct {
 	ModuleSets      map[string]moduleSet `yaml:"module-sets"`
@@ -42,6 +43,17 @@ type moduleSet struct {
 	Modules []string `yaml:"modules"`
 }
 
+// tagFetchDeadline bounds how long we wait for paginated ListTags calls to GitHub.
+const tagFetchDeadline = 5 * time.Minute
+
+var httpClient = newVersionscheckHTTPClient()
+
+func newVersionscheckHTTPClient() *http.Client {
+	return &http.Client{
+		Timeout: tagFetchDeadline,
+	}
+}
+
 func main() {
 	log.SetFlags(0)
 	log.SetPrefix("versionscheck: ")
@@ -51,16 +63,25 @@ func main() {
 }
 
 func run() error {
-	versionsPath := flag.String("versions", "../../versions.yaml", "path to versions.yaml")
+	versionsPath := flag.String("versions", "", "required: path to versions.yaml")
+	githubRepo := flag.String("github-repo", "", "required: GitHub repository as owner/repo;")
+	moduleKey := flag.String("module-key", "", "required: module-sets key in versions.yaml (e.g. edot-base)")
+	allowEmptyTag := flag.Bool("allow-empty-tag", false, "allow a repository with no semver-shaped tags; otherwise fail when GitHub returns no v* tags")
+
 	quiet := flag.Bool("q", false, "suppress success message on stdout")
 	flag.Parse()
 
-	vpath := *versionsPath
-	if vpath == "" {
-		vpath = "../../versions.yaml"
+	if *versionsPath == "" {
+		return fmt.Errorf("-versions is required")
+	}
+	if *githubRepo == "" {
+		return fmt.Errorf("-github-repo is required (owner/repo)")
+	}
+	if *moduleKey == "" {
+		return fmt.Errorf("-module-key is required")
 	}
 
-	data, err := os.ReadFile(vpath)
+	data, err := os.ReadFile(*versionsPath)
 	if err != nil {
 		return fmt.Errorf("read versions file: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


jmmcorreia:
@coderabbitai review

internal/versionscheck/main.go
```diff
@@ -0,0 +1,145 @@
...
+func main() {
+	log.SetFlags(0)
+	log.SetPrefix("versionscheck: ")
+	if err := run(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func run() error {
+	versionsPath := flag.String("versions", "../../versions.yaml", "path to versions.yaml")
```
rogercoll: What do you think if we remove the relative default path and make it a required flag? This might be error-prone depending on the directory is being run.

internal/versionscheck/main.go
```diff
@@ -0,0 +1,145 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+
+	"golang.org/x/mod/semver"
+	"gopkg.in/yaml.v3"
+)
+
+const edotBaseKey = "edot-base"
```
rogercoll: Should we make this a parameter (flag provided in the Makefile)? Just to make the binary agnostic to our versions.yaml

internal/versionscheck/main.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func gitTagList() (string, error) {
+	cmd := exec.Command("git", "tag", "-l")
+	out, err := cmd.Output()
+	if err != nil {
+		return "", fmt.Errorf("git tag -l: %w", err)
+	}
+	return string(out), nil
+}
```
rogercoll: As normally working from a forked repository, we have to fetch the latest upstream tags. Upstream git configuration might be different in each set-up. Maybe we add a flag to specify the Github repository and use the GitHub API to get the latest tags instead?

internal/versionscheck/main.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return err
+	}
+	latest := latestComponentVersion(strings.Split(tagOut, "\n"))
+
+	// Require edot-base to be strictly greater than the highest semver tag (if any).
+	if latest != "" && semver.Compare(ms.Version, latest) <= 0 {
+		return fmt.Errorf("module-sets version %s must be greater than tag %s",
+			ms.Version, latest)
+	}
```
rogercoll: Should we separate this into two different errors? One for "couldn't get latest component version" and the other for the actual version check.

rogercoll (commented):
Good work, added a couple of nit comments. What do you think moving the whole package into `./internal/tools/versioncheck` instead?

internal/versionscheck/main.go
jmmcorreia: But do we want to error out if we couldn't get a valid component version? I didn't fail for that scenario since, for example, all tags could be a name, so no version is available and latest would be empty. (Or a repo has no tags yet)

I'm ok to change this being that the repo already has ongoing tags, so it creates an extra check. But just to confirm, we are okay then to assume that version tags exist in the repo right ?

internal/versionscheck/main.go
jmmcorreia: Sure will change this to use a github client and allow the repo to be specified as a flag.

internal/versionscheck/main.go
jmmcorreia: Makes sense, I will make this change

internal/versionscheck/main.go
jmmcorreia: Sure, I will make it a flag and add in the path in the Makefile.

internal/versionscheck/main.go
rogercoll: > But do we want to error out if we couldn't get a valid component version? I didn't fail for that scenario since, for example, all tags could be a name, so no version is available and latest would be empty.

Good point, but if `latest` is empty it would print: `module-sets version (some-version) must be greater than tag`.  What do you think if we use quoted verbs instead? In that case the error message will indicate empty strings:

```suggestion
	if latest != "" && semver.Compare(ms.Version, latest) <= 0 {
		return fmt.Errorf("module-sets version %q must be greater than tag %q",
			ms.Version, latest)
	}
```

jmmcorreia: Based on the initial feedback I modified the code and now this is the current behavior:

```go
// validateModuleSetVersionAgainstTagList enforces that, if a tag exists (i.e. latest is not empty),
// then the module set version must be greater than the latest tag.
// If latest is empty, an error is returned unless allowEmptyTag flag is set.
func validateModuleSetVersionAgainstTagList(msVersion, latest, tagName string, allowEmptyTag bool, githubRepo string) error {
	if latest == "" && !allowEmptyTag {
		return fmt.Errorf("no semver tags found for repo %s", githubRepo)
	}
	if latest != "" && semver.Compare(msVersion, latest) <= 0 {
		return fmt.Errorf("module-sets version %s must be greater than version %s from tag %s",
			msVersion, latest, tagName)
	}
	return nil
}
```

A flag was introduced to control if we want to fail when latest is empty (With a matching error message).
If not, then it defaults to the initial behavior.

Does this look ok to you ?

If latest is empty and allowed by the flag, then no error message should be printed.
 
Btw, the final print was also refactored into `formatVersionscheckOK` and checks for latest being empty before the final print.

jmmcorreia:
> Good work, added a couple of nit comments. What do you think moving the whole package into `./internal/tools/versioncheck` instead?

Right now tools seems mostly to serve as an import for external tools and make them available locally whilst this is repo specific tool maintained in the repo itself. I am ok to move it, just preferred initially to make separately due to the point above, but I don't have a strong preference tbh. Would you prefer to see it under `./internal/tools/versionscheck`

internal/versionscheck/main.go
rogercoll: Sounds great! Thanks

rogercoll (approved):
LGTM, would you mind adding a two lines README file describing the purpose and functionality of this tool? Thank you!

rogercoll (approved):
Thanks!

PR #1186, opened 2026-04-22
[receiver/elasticapmintakereceiver] reject non apm events sent to intake/v2/events endpoint

aelnahas:
# Summary
we currently accept both apm and ingest events on the intake/v2/events. This change now makes so that elasticapmintakereceiver would reject non apm events on that endpoint. The change is being done on the receiver side, to have consistent behavior across the different consumers.

### DIFF
--- a/receiver/elasticapmintakereceiver/receiver.go
+++ b/receiver/elasticapmintakereceiver/receiver.go
@@ -181,6 +181,17 @@ func (r *elasticAPMIntakeReceiver) newElasticAPMEventsHandler(ctxFunc func(*http
 	})
 
 	return func(w http.ResponseWriter, req *http.Request) {
+		var result struct {
+			Accepted int      `json:"accepted"`
+			Errors   []string `json:"errors,omitempty"`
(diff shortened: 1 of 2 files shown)


receiver/elasticapmintakereceiver/receiver.go
```diff
@@ -750,6 +757,13 @@ func mapSpanKind(kind string) ptrace.SpanKind {
 	}
 }
 
+// isAPMEndpoint reports whether the request arrived via an APM endpoint
+// by inspecting the Host header. Intake-v2 is only supported on .apm
+// endpoints, not .ingest.
+func isAPMEndpoint(req *http.Request) bool {
+	return strings.HasPrefix(req.Host, "apm.") || strings.Contains(req.Host, ".apm.")
```
carsonip, line 764: what kind of correctness do we need for the guard? strings.Contains seems a bit lax

aelnahas:
closing this for now as it isnt the right solution.

PR #1188, opened 2026-04-27
[receiver/prometheusremotewritev1] - Add prometheusremotewritev1 recevier 

VihasMakwana:
Closes https://github.com/elastic/opentelemetry-dev/issues/1312


receiver/prometheusremotewritev1receiver/receiver.go
```diff
@@ -0,0 +1,280 @@
...
+	instance   string
+	metricName string
+}
+
+// translate converts a v1 WriteRequest into OTLP pmetric.Metrics.
+//
+// Because the v1 protocol carries no metric metadata (type, unit, help text),
+// metric types is defaulted to Guage.
+// Stale markers (special NaN 0x7ff0000000000002) are translated to data points
+// with the NoRecordedValue flag set.
```
felixbarny: Does the remote write exporter reverse this? I.e. does a staleness marker survive the roundtrip?

receiver/prometheusremotewritev1receiver/receiver.go
```diff
@@ -0,0 +1,280 @@
...
+func (r *prometheusRWv1Receiver) translate(wr *prompb.WriteRequest) pmetric.Metrics {
```
felixbarny: I would make it even simpler:
- No grouping of resource metrics - job and instance simply become regular attributes so that there's a single resource metric which has no resource attributes.
- No special treatment of the `target_info` metric
- No grouping of metrics by metric key. Each `wr.Timeseries` simply becomes a metric and it's `ts.Samples` become data points.

receiver/prometheusremotewritev1receiver/receiver.go
VihasMakwana: No. The exporter deletes the metric if the `NoRecordedValue` is set. 

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/69ab06ff2702c0f73bdc0f795c421aec33a5ba7c/exporter/prometheusexporter/accumulator.go#L166-L169

receiver/prometheusremotewritev1receiver/receiver.go
VihasMakwana: Done!

receiver/prometheusremotewritev1receiver/receiver.go
VihasMakwana: How does the ES remote write endpoint handle this? We can set the value as is if needed and avoid setting `NoRecordedValue` flag.

felixbarny: I'd then keep the value as-is. The endpoint can't handle staleness markers, yet, but it will do so in the future. The receiver/exporter should always faithfully roundtrip. This is the single most important invariant.

receiver/prometheusremotewritev1receiver/receiver.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+var reservedLabels = map[string]struct{}{
+	"__name__": {},
+}
+
+func buildAttributes(labels []prompb.Label) pcommon.Map {
+	attrs := pcommon.NewMap()
+	for _, lbl := range labels {
+		if _, ok := reservedLabels[lbl.Name]; !ok {
```
felixbarny: With just `__name__` being a reserved label, I'd remove `reservedLabels` and just directly check the string. Maybe create a const for it as we also use it above.

receiver/prometheusremotewritev1receiver/receiver.go
```diff
@@ -0,0 +1,223 @@
...
+		return
+	}
+	// The v1 spec mandates application/x-protobuf. No proto= parameter is needed.
+	if !strings.HasPrefix(contentType, "application/x-protobuf") {
+		r.settings.Logger.Warn("Unsupported Content-Type", zap.String("content_type", contentType))
+		http.Error(w, "Content-Type must be application/x-protobuf", http.StatusUnsupportedMediaType)
+		return
+	}
+
+	reqBody, err := io.ReadAll(req.Body)
```
felixbarny, line 133: Does the receiver handle snappy decoding?

VihasMakwana: Yes. The [`confighttp` injects](https://github.com/open-telemetry/opentelemetry-collector/blob/261c6b9780f7e6a513fb0ee39682dc605972adb4/config/confighttp/server.go#L218-L224) a "decompressor" handler and that decodes the data based on `Content-Encoding` header.

VihasMakwana:
I feel like this is ready for review.

receiver/prometheusremotewritev1receiver/receiver.go
VihasMakwana: Done!

axw (commented):
Looks good. Do we need to translate job/instance to service.name/service.instance.id? Or will the data roundtrip through the PRW exporter cleanly without that?

receiver/prometheusremotewritev1receiver/receiver.go
```diff
@@ -0,0 +1,212 @@
...
+		return
+	}
+
+	md := r.translate(&wr)
+	if md.MetricCount() == 0 {
+		w.WriteHeader(http.StatusNoContent)
+		return
+	}
+
+	obsCtx := r.obsrecv.StartMetricsOp(req.Context())
```
vigneshshanmugam: this should be moved near start and we need to account for errors that happens inside no content-encoding/marshaling failures etc.

vigneshshanmugam (commented):
LGTM, just a small callout on the metrics reporting, rest looks good.

receiver/prometheusremotewritev1receiver/receiver.go
```diff
@@ -0,0 +1,212 @@
...
+		return
+	}
+	// The v1 spec mandates application/x-protobuf. No proto= parameter is needed.
+	if !strings.HasPrefix(contentType, "application/x-protobuf") {
+		r.settings.Logger.Warn("Unsupported Content-Type", zap.String("content_type", contentType))
+		http.Error(w, "Content-Type must be application/x-protobuf", http.StatusUnsupportedMediaType)
+		return
+	}
+
+	reqBody, err := io.ReadAll(req.Body)
```
felixbarny, line 133: Another question about this: do we enforce a maximum compressed/uncompressed request size here?

receiver/prometheusremotewritev1receiver/receiver.go
```diff
@@ -0,0 +1,212 @@
...
+		r.settings.Logger.Warn("Protobuf unmarshal failed", zap.Error(err))
+		http.Error(w, fmt.Sprintf("protobuf unmarshal: %v", err), http.StatusBadRequest)
+		return
+	}
+
+	md := r.translate(&wr)
+	if md.MetricCount() == 0 {
+		w.WriteHeader(http.StatusNoContent)
+		return
+	}
```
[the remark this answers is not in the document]
felixbarny: This has a point but we shouldn't fail the entire request and return early. Can we instead continue processing but still mark the request as a partial, non-retry-able error?

VihasMakwana:
@axw 
> Looks good. Do we need to translate job/instance to service.name/service.instance.id? Or will the data roundtrip through the PRW exporter cleanly without that?

The data will roundtrip cleanly. I'll confirm this once again.

receiver/prometheusremotewritev1receiver/receiver.go
VihasMakwana: We rely on [max_request_body_size](https://github.com/open-telemetry/opentelemetry-collector/tree/main/config/confighttp#:~:text=max_request_body_size%3A%20configures%20the%20maximum%20allowed%20body%20size%20in%20bytes%20for%20a%20single%20request.%20Default%3A%2020971520%20(20MiB)) for this. We're using default limit of 20MB for now.

VihasMakwana: That is, 20 MB of [compressed data](https://github.com/open-telemetry/opentelemetry-collector/blob/c9fcea821a4c32f05ca985e3be0c58337bf0059b/config/confighttp/server.go#L384-L385).

felixbarny: sounds good 👍

receiver/prometheusremotewritev1receiver/receiver.go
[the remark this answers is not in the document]
VihasMakwana: @felixbarny We can do this, but this might lead to duplication if the user retries the same request with correct labels.

receiver/prometheusremotewritev1receiver/receiver.go
VihasMakwana: Done!

receiver/prometheusremotewritev1receiver/receiver.go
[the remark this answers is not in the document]
VihasMakwana: Just found out that upstream receiver rejects the data straightaway if any error is encountered.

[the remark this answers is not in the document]
felixbarny: > this might lead to duplication if the user retries the same request with correct labels.

Not sure I understand. We'd mark the request as non-retry-able as a simple retry would result in the same issue. If the client fixes the request so that the `__name__` label is present, that's fine but I wouldn't call this a re-try. It would require code changes on their side. I also don't see how this would result in duplication. Elasticsearch would reject data points that are already ingested.

I also think we shouldn't fail the entire request if a single metric has a missing `__name__` label.

receiver/prometheusremotewritev1receiver/receiver.go
```diff
@@ -0,0 +1,223 @@
...
+	var listener net.Listener
+	if listener, err = r.config.ToListener(ctx); err != nil {
+		return fmt.Errorf("failed to create listener: %w", err)
+	}
+
+	r.shutdownWG.Go(func() {
+		if err := r.server.Serve(listener); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			r.settings.Logger.Error("HTTP server error", zap.Error(err))
+		}
+	})
```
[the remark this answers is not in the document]
VihasMakwana: > Also note: the codebase pattern is r.shutdownWG.Add(1) + defer r.shutdownWG.Done() rather than WaitGroup.Go.

You're incorrect. It's `shutdownWG.Go`

receiver/prometheusremotewritev1receiver/receiver.go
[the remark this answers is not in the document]
VihasMakwana: > Elasticsearch would reject data points that are already ingested.

Oh okay. That makes sense. 

I have added the ability to process other samples, if one of the samples is missing ` __name__` label.

receiver/prometheusremotewritev1receiver/receiver.go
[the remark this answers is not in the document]
VihasMakwana: Done!

felixbarny (approved):

rogercoll (commented):
Could you add the receiver into the `versions.yaml` file? (so it will be released when we bump the OTel deps) 🙏

VihasMakwana:
@rogercoll Ack. Should I also add it to `manifest.yml` ?

vigneshshanmugam (approved):
LGTM

rogercoll:
> @rogercoll Ack. Should I also add it to `manifest.yml` ?

Yes, it will ensure compilation with other components. Thanks!

axw (approved):
LGTM. Not convinced we should be logging warnings on read/unmarshal errors, since that's client-controlled, but we can follow up on that.

receiver/prometheusremotewritev1receiver/receiver.go
```diff
@@ -0,0 +1,225 @@
...
+
+// handleWrite is the HTTP handler for the remote write endpoint.
+// It validates the request per the v1 spec, decodes it, translates to OTLP, and forwards to the next consumer.
+func (r *prometheusRWv1Receiver) handleWrite(w http.ResponseWriter, req *http.Request) {
+	obsCtx := r.obsrecv.StartMetricsOp(req.Context())
+	contentType := req.Header.Get("Content-Type")
+	if contentType == "" {
+		r.settings.Logger.Warn("Request missing Content-Type header")
+		http.Error(w, "Content-Type header is required", http.StatusUnsupportedMediaType)
+		r.obsrecv.EndMetricsOp(obsCtx, "prometheusremotewritev1receiver", 0, errors.New("request missing Content-Type header"))
```
vigneshshanmugam: nit: you can use metadata.Type.String() here from receiver/prometheusremotewritev1receiver/internal/metadata`

VihasMakwana: Done!

VihasMakwana:
@axw The `__name__` and job/instance labels survive the round trip. No need to translate.

> LGTM. Not convinced we should be logging warnings on read/unmarshal errors, since that's client-controlled, but we can follow up on that.

Will do. I’ll also open a follow-up issue to add some e2e tests covering the prwv1 receiver and prw exporter round trip.

PR #1192, opened 2026-05-01
receiver/entityanalyticsreceiver: wire Jamf into provider registry

efd6:
Register the entcollect Jamf provider under the "jamf" name so that the entity analytics receiver can use it. The registration mirrors the pattern established by the receiver added in #1153.

A local mapstructure-tagged struct (jamfLocalConf) is defined inside jamfProvider for the cfg != nil path, bridging confmap.Conf to ecjamf.Config. SyncInterval and UpdateInterval are intentionally absent from jamfLocalConf: they are owned by the receiver's top-level Config. When cfg is nil (the current default, since the receiver passes no provider sub-config), credentials are read from the JAMF_TENANT, JAMF_USERNAME, and JAMF_PASSWORD environment variables.

TestJamfProviderConfigRoundTrip guards against future drift: a field-count assertion on ecjamf.Config catches unhandled additions, and a functional round-trip through confmap.Conf confirms that the six provider-owned fields are unmarshalled and validated end-to-end.

For elastic/beats#49164

### DIFF
--- a/receiver/entityanalyticsreceiver/config.go
+++ b/receiver/entityanalyticsreceiver/config.go
@@ -51,5 +51,8 @@ func (cfg *Config) Validate() error {
 	if cfg.UpdateInterval <= 0 {
 		return fmt.Errorf("update_interval must be positive")
 	}
+	if cfg.SyncInterval <= cfg.UpdateInterval {
+		return fmt.Errorf("sync_interval must be greater than update_interval")
+	}
 	return nil
 }
--- a/receiver/entityanalyticsreceiver/jamf.go
+++ b/receiver/entityanalyticsreceiver/jamf.go
@@ -71,19 +71,25 @@ func jamfProvider(cfg *confmap.Conf) (entcollect.Provider, error) {
 		ec.Username = os.Getenv("JAMF_USERNAME")
 		ec.Password = os.Getenv("JAMF_PASSWORD")
 		if v := os.Getenv("JAMF_PAGE_SIZE"); v != "" {
-			if n, err := strconv.Atoi(v); err == nil {
-				ec.PageSize = n
+			n, err := strconv.Atoi(v)
+			if err != nil {
+				return nil, fmt.Errorf("jamf: parse JAMF_PAGE_SIZE: %w", err)
 			}
+			ec.PageSize = n
 		}
 		if v := os.Getenv("JAMF_IDSET_SHARDS"); v != "" {
-			if n, err := strconv.Atoi(v); err == nil {
-				ec.IDSetShards = n
+			n, err := strconv.Atoi(v)
+			if err != nil {
+				return nil, fmt.Errorf("jamf: parse JAMF_IDSET_SHARDS: %w", err)
 			}
+			ec.IDSetShards = n
 		}
 		if v := os.Getenv("JAMF_TOKEN_GRACE_PERIOD"); v != "" {
-			if d, err := time.ParseDuration(v); err == nil {
-				ec.TokenGrace = d
+			d, err := time.ParseDuration(v)
+			if err != nil {
+				return nil, fmt.Errorf("jamf: parse JAMF_TOKEN_GRACE_PERIOD: %w", err)
 			}
+			ec.TokenGrace = d
 		}
 	}
 	err := ec.Validate()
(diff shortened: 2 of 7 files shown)


receiver/entityanalyticsreceiver/jamf.go
```diff
@@ -0,0 +1,81 @@
...
+
+func jamfProvider(cfg *confmap.Conf) (entcollect.Provider, error) {
+	ec := ecjamf.DefaultConfig()
+	if cfg != nil {
+		// jamfLocalConf mirrors ecjamf.Config with mapstructure tags for confmap.Conf
+		// Unmarshal. If ecjamf.Config gains a new field, add it here too.
+		type jamfLocalConf struct {
+			TenantID    string        `mapstructure:"jamf_tenant"`
+			Username    string        `mapstructure:"jamf_username"`
+			Password    string        `mapstructure:"jamf_password"`
```
andrewkroh: OTel appears to use `configopaque.String` for credentials.

https://pkg.go.dev/go.opentelemetry.io/collector/config/configopaque#String

receiver/entityanalyticsreceiver/jamf.go
```diff
@@ -0,0 +1,81 @@
...
+		ec.PageSize = lc.PageSize
+		ec.IDSetShards = lc.IDSetShards
+		ec.TokenGrace = lc.TokenGrace
+	} else {
+		// cfg is nil when the receiver has no provider-specific sub-config
+		// (the current default). Credentials are read from environment variables.
+		ec.TenantID = os.Getenv("JAMF_TENANT")
+		ec.Username = os.Getenv("JAMF_USERNAME")
+		ec.Password = os.Getenv("JAMF_PASSWORD")
+		if v := os.Getenv("JAMF_PAGE_SIZE"); v != "" {
```
andrewkroh, line 73: Should it also honor JAMF_IDSET_SHARDS and JAMF_TOKEN_GRACE_PERIOD?

receiver/entityanalyticsreceiver/jamf.go
```diff
@@ -0,0 +1,81 @@
...
+	"time"
+
+	"go.opentelemetry.io/collector/confmap"
+
+	"github.com/elastic/entcollect"
+	ecjamf "github.com/elastic/entcollect/provider/jamf"
+)
+
+func init() {
+	Register("jamf", jamfProvider)
```
andrewkroh, line 34: The README.md file needs updated.

> The provider registry is currently empty — concrete provider implementations (Okta, EntraID, Active Directory, Jamf) will be added in subsequent work.

andrewkroh (commented):
These are not strictly related to this diff and can be addressed/discussed independently of the PR.

----

After the review of JAMF implementation in elastic/entcollect, I was wondering what happens when a full sync interval collection fails for an HTTP 429 (or some other reason). Will it see that previous full sync failed and try again or will it proceed with an incremental sync?

And I think the answer is no from `receiver.go`. I think we would want it to be self-healing. As in if "last sync time" is missing or older than the configured sync interval then do a full sync, otherwise incremental sync. For JAMF this isn't an issue since full and incremental are effectively the same, but if could affect the future providers.

----

In the config validation, we should assert that `SyncInterval` > `UpdateInterval` to prevent a thrashy, incorrect schedule.
https://github.com/elastic/opentelemetry-collector-components/blob/554890212e63a2dc49d86e2dde2b29f4da9616dc/receiver/entityanalyticsreceiver/config.go#L48-L53

receiver/entityanalyticsreceiver/config.go
```diff
@@ -51,5 +51,8 @@ func (cfg *Config) Validate() error {
 	if cfg.UpdateInterval <= 0 {
 		return fmt.Errorf("update_interval must be positive")
 	}
+	if cfg.SyncInterval <= cfg.UpdateInterval {
+		return fmt.Errorf("sync_interval must be greater than update_interval")
+	}
```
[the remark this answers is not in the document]
efd6: They will not fail, they are never validated since it's injected directly, not going through the OTel framework. Will adjust though just for cleanliness.

receiver/entityanalyticsreceiver/jamf.go
```diff
@@ -0,0 +1,94 @@
...
+		if v := os.Getenv("JAMF_IDSET_SHARDS"); v != "" {
+			if n, err := strconv.Atoi(v); err == nil {
+				ec.IDSetShards = n
+			}
+		}
+		if v := os.Getenv("JAMF_TOKEN_GRACE_PERIOD"); v != "" {
+			if d, err := time.ParseDuration(v); err == nil {
+				ec.TokenGrace = d
+			}
+		}
```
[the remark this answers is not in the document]
efd6: PTAL

andrewkroh (approved):

PR #1194, opened 2026-05-04
Support batches larger than burst in ratelimit processor

straistaru:
## What this fixes

In delay mode, the rate limit processor was supposed to wait when there was too much traffic and let it through later. But if a single incoming batch had more events than the configured burst size (for example a batch of 100 events with burst set to 10), the processor would give up and return an error instead of waiting. That defeats the whole point of delay mode and causes events to get dropped upstream.

This PR makes delay mode actually delay, even for big batches.

## How it works

For each incoming batch, the processor now:

1. Splits the batch into smaller pieces, each one no larger than the burst size.
2. Reserves time slots for all the pieces at once, while holding a short lock so other concurrent batches don't sneak in between the pieces. This makes the upstream callers free up gradually instead of all at once, which gives a smoother backpressure signal.
3. Releases the lock and sleeps for the total wait time.
4. Forwards the full batch to the next stage when the wait is done.

The batch itself is never split when it leaves the processor — the next stage still sees one batch with all its events. The splitting only happens internally to figure out the right amount of time to wait.

## Behavior in other modes

Error mode is unchanged. It still rejects batches that are too big to fit in the burst.

## Tests

Added tests for:
- A single big batch that needs to wait gets through after the expected delay.
- Multiple concurrent big batches finish one after another in arrival order.
- Mixed small and big batches all complete without one starving the others.
- Cancelling the context mid-wait returns the unused time slots so the next caller doesn't inherit phantom waiting.
- Different metadata keys (different tenants) don't block each other.

All tests pass under `-race`.

### DIFF
--- a/processor/ratelimitprocessor/local.go
+++ b/processor/ratelimitprocessor/local.go
@@ -30,12 +30,26 @@ import (
 
 var _ RateLimiter = (*localRateLimiter)(nil)
 
+// keyState holds the per-rate-limit-key state: the underlying token-bucket
+// limiter and a mutex used to make a multi-chunk reservation atomic with
+// respect to concurrent callers sharing the same key.
+type keyState struct {
+	limiter *rate.Limiter
+	// reserveLock serializes reservation calculation across concurrent
+	// callers for this key, so callers acquire reservations FIFO instead
+	// of having their per-chunk ReserveN calls interleave. The lock is
+	// held only during the (microsecond-scale) reservation loop and is
+	// released before the caller waits, so concurrent callers queue on
+	// arrival but wait in parallel.
+	reserveLock sync.Mutex
+}
+
 type localRateLimiter struct {
 	cfg *Config
 	set processor.Settings
 	// TODO use an LRU to keep a cap on the number of limiters.
 	// When the LRU capacity is exceeded, reuse the evicted limiter.
-	limiters sync.Map
+	limiters sync.Map // map[string]*keyState
 }
 
 func newLocalRateLimiter(cfg *Config, set processor.Settings) (*localRateLimiter, error) {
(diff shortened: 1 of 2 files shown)


processor/ratelimitprocessor/local.go
```diff
@@ -58,26 +72,90 @@ func (r *localRateLimiter) RateLimit(ctx context.Context, hits int) error {
 	key := getUniqueKey(metadata, r.cfg.MetadataKeys)
 	cfg := resolveRateLimit(r.cfg, metadata)
 
-	v, _ := r.limiters.LoadOrStore(key, rate.NewLimiter(rate.Limit(cfg.Rate), cfg.Burst))
-	limiter := v.(*rate.Limiter)
+	v, _ := r.limiters.LoadOrStore(key, &keyState{
```
michalpristas, line 75: i think this could be breaking right?

michalpristas (approved):
i mean yeah whole whole delay when burst is large goes kind of against principles of distributed computing. we can plan on executing huge chunk of events/requests that can cause other cascading effects. but yes contract for this does not allow us to do something like partial allow.

processor/ratelimitprocessor/local.go
straistaru: I went through the line and I don't see a specific break.
Are you pointing at the `keyState` wrapper change (storage type went from `*rate.Limiter` to `*keyState`)? If this is the case `r.limiter` is unexported and the only other reader is `Shutdow()`, which still works.

If it's something else, let me know and I'll look into it.

michalpristas: you're right, i was thinking this gets persisted. but it's not.

PR #1197, opened 2026-05-05
[connector/elasticapmconnector] add _doc_count mapping hint and forward elasticsearch.doc_count

aelnahas:
This PR will enable us to `_doc_count` for spans events. The change sets the `_doc_count` mapping hint, and for composite spans it will also forward `span.composite.count` to `elasticsearch.doc_count` so that we can set the `_doc_count` correctly in the exporter.

This exporter PR: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48109 shows how `elasticsearch.doc_count` will be used to set the composite span _doc_count correctly.


aelnahas:
Closing this PR as I found out that the Kibana UI does not rely on `_doc_count` for these signals and instead it relies on the `SUM` on `span.destination.service.response_time.count`

PR #1198, opened 2026-05-05
[processor/elasticapmprocessor] map span.composite.count to elasticsearch.doc_count

aelnahas:
This pr will simply translate/ map the value of the `span.composite.count` to `elasticsearch.doc_count`. This will be used by the exporter to set the `_doc_count` attribute for composite spans.

This PR will be used in conjunction with:
- exporter change: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48109
- elasticapmconnector change: https://github.com/elastic/opentelemetry-collector-components/pull/1197


aelnahas:
Closing this PR as I found out that the Kibana UI does not rely on `_doc_count` for these signals and instead it relies on the `SUM` on `span.destination.service.response_time.count`

PR #1210, opened 2026-05-06
Update versions.yaml and codeowners

gregkalapos:
Adapting files according to the release process.

### CODEOWNERS
- Adding `elastic/ingest-otel-data` to `receiver/entityanalyticsreceiver`: we added `elastic/ingest-otel-data` for all modules because this team is primarily responsible for releasing this repo and keeping upstream OTel dependencies in sync - for this, with each upstream update we bump the dependency in `go.mod` files in each module resulting changes in `go.mod` and `go.sum` in all modules. In order to be able to do this as a single team we added this team in https://github.com/elastic/opentelemetry-collector-components/pull/1020 - this PR just keeps this in sync. If we ever need to do more than just bumping up the version, we always ping the real code owner of the module

### versions.yaml
- Bumping version, as usual
- Plus I found some new modules and added those as well, otherwise `make push-tags` fails with those modules missing.


gregkalapos:
Closing as outdated.

PR #1216, opened 2026-05-11
receiver/entityanalyticsreceiver: register Active Directory provider

efd6:
Wire the entcollect AD provider into the receiver's provider registry. Configuration is read from confmap with configtls for TLS, falling back to AD_URL/AD_BASE/AD_USER/AD_PASS environment variables when no provider sub-config is present.

Assisted-By: Cursor

For elastic/beats#49162


receiver/entityanalyticsreceiver/ad.go
```diff
@@ -0,0 +1,137 @@
...
+	} else {
+		ec.URL = os.Getenv("AD_URL")
+		ec.BaseDN = os.Getenv("AD_BASE")
+		ec.User = os.Getenv("AD_USER")
+		ec.Password = os.Getenv("AD_PASS")
+		if v := os.Getenv("AD_DATASET"); v != "" {
+			ec.Dataset = v
+		}
+		if v := os.Getenv("AD_PAGING_SIZE"); v != "" {
+			n, err := strconv.Atoi(v)
```
andrewkroh:
```suggestion
			n, err := strconv.ParseUint(v, 10, 32)
```

to prevent negatives that wrap after becomign uint32

receiver/entityanalyticsreceiver/ad.go
```diff
@@ -0,0 +1,137 @@
...
+	err := ec.Validate()
+	if err != nil {
+		return nil, err
+	}
+	return ecad.New(ec)
+}
+
+// adLocalConf mirrors ecad.Config with mapstructure tags for confmap.Conf
+// Unmarshal. If ecad.Config gains a new field, add it here too.
+type adLocalConf struct {
```
andrewkroh: The jamf equivalent made this type function-local. Since adLocalConf is only used inside adProvider, it also could be a local type to match the pattern. (feel free to ignore, just wanted to flag the inconsistency that has no effect)

https://github.com/elastic/opentelemetry-collector-components/blob/bee16dc744135b4899a234b0774d049f1e1b4fd9/receiver/entityanalyticsreceiver/jamf.go#L37-L49

andrewkroh (approved):
LGTM

receiver/entityanalyticsreceiver/ad.go
efd6: I missed this here.

PR #1219, opened 2026-05-12
[receiver/elasticapm]Handle dropped span stats by producing noindex spans

lahsivjar:
## Description

Implements MIS-equivalent `service_destination` aggregation for `transaction.dropped_spans_stats` in the intake-v2 receiver.

For each transaction event carrying `dropped_spans_stats`, the receiver now emits one synthetic CLIENT span per stat, carrying:
- `span.destination.service.resource`, `service.target.{type,name}`, `event.outcome`
- `span.composite.sum.us` and `span.composite.count` (consumed by elasticapmconnector's existing composite-span rules to produce `service_destination` metrics — matches MIS's `apm-aggregation.addDroppedSpanStatsMetrics`)
- `elasticsearch.mapping.hints: [_noindex]` to instruct elasticsearchexporter to skip indexing (depends on the upstream `_noindex` hint — see [linked PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48310))

Synthetic spans inherit `trace_id`, `tracestate` (so `AdjustedCount()` reflects the parent's sampling weight), and start/end timestamps from the parent transaction. `span.name` is intentionally empty, mirroring MIS's `setDroppedSpanStatsKey` (no span name dimension in dropped-stats aggregation). Span IDs are derived deterministically via `xxhash(parent_span_id || index)` for stable test fixtures; uniqueness within the trace is guaranteed and `_noindex` removes any downstream collision concern.

Also removes the previously unreachable `if event.Metricset != nil { setDroppedSpansStatsList(...) }` block in `intakeV2ToElasticSpecificFields.go` and the now-unused `TransactionDroppedSpansStats*` constants from `internal/elasticattr`. The block was dead in the receiver — `event.Metricset` is never populated alongside `Transaction` in the raw decoded intake-v2 path; that dual shape only exists in aggregator output (in MIS), and even there `apm-aggregation`'s `spanMetricsToAPMEvent` does not propagate `DroppedSpansStats` onto the emitted metric event, so the raw list was never actually indexed in MIS either.

## Related

- Upstream `_noindex` mapping hint in `elasticsearchexporter` (required for the synthesis to be silent on the indexed-traces side): [elastic/opentelemetry-collector-contrib#48310](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48310)
- https://github.com/elastic/hosted-otel-collector/issues/3114

## Testing

Added test cases


receiver/elasticapmintakereceiver/receiver.go
```diff
@@ -733,6 +736,85 @@ func (r *elasticAPMIntakeReceiver) elasticSpanToOTelSpan(s *ptrace.Span, event *
...
+		d := xxhash.New()
+		_, _ = d.Write(parentSpanID[:])
+		var idxBuf [4]byte
+		binary.BigEndian.PutUint32(idxBuf[:], uint32(i))
+		_, _ = d.Write(idxBuf[:])
+		var spanID pcommon.SpanID
+		binary.BigEndian.PutUint64(spanID[:], d.Sum64())
+
+		s := ss.Spans().AppendEmpty()
+		s.SetTraceID(parent.TraceID())
```
vigneshshanmugam, line 829: Do we need all these fields to be set for the span just for producing the summary metrics? Wondering if there is a need for all this just to be dropped again.

vigneshshanmugam (approved):
LGTM, Just had a minor comment https://github.com/elastic/opentelemetry-collector-components/pull/1219/changes#r3229173557

receiver/elasticapmintakereceiver/receiver.go
lahsivjar: I have removed a couple of them and added comments for others specifying why they are needed.

PR #1220, opened 2026-05-12
[connector/elasticapmconnector] make transaction name optional for APM aggregation

lanre-ade:
## Summary
  - Make `transaction.name` optional for `elasticapmconnector` transaction-level aggregation.
  - Add a regression test confirming `metrics-apm.transaction.1m` is emitted when an incoming transaction has no `transaction.name`.
  ## Why
  currently skipping `transaction.1m` aggregation when `transaction.name` is missing, because `transaction.name` is configured as a required signal-to-metrics attribute. Legacy APM intake/APM Server does not require `transaction.name` for transaction aggregation; it groups nameless transactions with an 
  empty/missing transaction name.

  ## Testing
  - `go test ./... -count=1` from `connector/elasticapmconnector`

### DIFF
--- a/connector/elasticapmconnector/config.go
+++ b/connector/elasticapmconnector/config.go
@@ -316,7 +316,7 @@ func (cfg Config) signaltometricsConfig() *signaltometricsconfig.Config {
 
 	transactionAttributes := append([]signaltometricsconfig.Attribute{
 		{Key: "transaction.root"},
-		{Key: "transaction.name"},
+		{Key: "transaction.name", Optional: true},
 		{Key: "transaction.type"},
 		{Key: "transaction.result", Optional: true},
 		{Key: "event.outcome"},
--- a/connector/elasticapmconnector/config.go
+++ b/connector/elasticapmconnector/config.go
@@ -310,6 +310,8 @@ func (cfg Config) signaltometricsConfig() *signaltometricsconfig.Config {
 
 	serviceTransactionAttributes := append([]signaltometricsconfig.Attribute{
 		{Key: "transaction.root"},
+		// processor defaults empty transaction.type to "unknown":
+		// https://github.com/elastic/opentelemetry-collector-components/blob/cdd457b05388ca0e5090301a32292f887752911e/processor/elasticapmprocessor/internal/enrichments/span.go#L476
 		{Key: "transaction.type"},
 		{Key: "metricset.name", DefaultValue: "service_transaction"},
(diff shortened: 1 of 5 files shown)


connector/elasticapmconnector/connector_test.go
```diff
@@ -182,6 +184,66 @@ func TestConnector_TracesToMetrics(t *testing.T) {
 	}
 }
 
+func TestConnector_TransactionMetricsWithoutName(t *testing.T) {
```
lahsivjar: Do we need a new test? Can we not add this case to existing tests?

lanre-ade: I considered adding a golden fixture for this. I can go ahead with that approach if that's what you had in mind

lahsivjar: What I mean is adding a new test case in `TestConnector_TracesToMetrics` that has no transcation.name

connector/elasticapmconnector/config.go
```diff
@@ -316,7 +316,7 @@ func (cfg Config) signaltometricsConfig() *signaltometricsconfig.Config {
 
 	transactionAttributes := append([]signaltometricsconfig.Attribute{
 		{Key: "transaction.root"},
-		{Key: "transaction.name"},
+		{Key: "transaction.name", Optional: true},
 		{Key: "transaction.type"},
 		{Key: "transaction.result", Optional: true},
 		{Key: "event.outcome"},
```
lahsivjar, line 328: Let's add a comment in `event.outcome` and `transaction.type` explaining they have defaults set to `unknown` and adding a link to the corresponding code.

connector/elasticapmconnector/testdata/traces/transaction_metrics_without_name/input.yaml
```diff
@@ -0,0 +1,37 @@
...
+                  stringValue: request
+              - key: event.outcome
+                value:
+                  stringValue: success
+              - key: transaction.representative_count
+                value:
+                  doubleValue: 1.0
+            startTimeUnixNano: "1581452772000000321"
+            endTimeUnixNano: "1581452783000000789"
+            name: http-span
```
lahsivjar, line 39: Technically, `span.name` is the same thing as `transaction.name` but since we don't have enrichments here the `transaction.name` is not populated and the test works - can you add this comment here?

lahsivjar (approved):
Just one nit.

PR #1224, opened 2026-05-14
update versions to 0.50.0

aelnahas:
update versions to 0.50.0


lanre-ade (approved):

lahsivjar (changes requested):
This is not complete, we are missing version up PR itself to bump up OTel component versions.

PR #1225, opened 2026-05-15
Update otel upstream components to v0.152.0/v1.58.0

lahsivjar:
As it says on the tin.


ChrsMark (approved):

ChrsMark:
@lahsivjar any breaking changes from the changelog that we should be cautious about?

lahsivjar:
`spanmetricsconnector`, `tailsamplingprocessor`, and `kafkametricsreceiver` have breaking changes ([ref](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.152.0)) that might be important for Elastic-Agent. From the perspective of mOTLP, the breaking changes looks good to me as we don't rely on those components.

In addition, there are a couple more breaking changes in the core libraries but they look okay to my eyes.

ishleenk17 (approved):

andrewkroh (approved):

PR #1226, opened 2026-05-18
Fix ECS label attribute handling in `TranslateResourceMetadata` and `Remap*` functions

lahsivjar:
### Bugs fixed

**1. Panic in `Remap*` functions during map iteration**

`RemapLogRecordAttributesToECSLabels` (and the equivalent span/metric variants) used `Range` to iterate the attribute map while inserting new keys into it. This caused a panic for three cases: map/bytes/empty-typed attribute values (removed with no compensating insert), a plain attribute whose sanitized destination key already existed, and two attributes whose keys both sanitize to the same output key. Fixed by switching to `RemoveIf` with deferred insertions collected into a slice and applied after iteration completes.

**2. APM intake path double-prefixing `labels.*` resource attributes**

The APM intake receiver places `labels.<key>` and `numeric_labels.<key>` directly on resource attributes (already sanitized). `TranslateResourceMetadata` was treating these as raw unsupported keys and fully re-normalizing them, producing `labels.labels_my_value` instead of `labels.my_value`. Fixed by adding a `sanitizeExistingLabels bool` parameter: the APM enricher path (`telemetry.sdk.name == "ElasticAPM"`) passes `true`, which sanitizes only the suffix of existing `labels.*`/`numeric_labels.*` keys; the OTel path passes `false` and retains the existing full re-normalization behaviour.

### Validation

Both behaviours are validated against `apm-data` (`input/otlp/traces_test.go`) — apm-data silently drops map/bytes/empty typed attributes and applies last-writer-wins on key collisions, which matches the processor's behaviour after this fix.


processor/elasticapmprocessor/internal/ecs/ecs_translation.go
```diff
@@ -37,17 +38,26 @@ const (
...
+// Moves unsupported attributes to labels.* / numeric_labels.* (key sanitized),
 // and leaves supported ECS attributes unchanged.
-func TranslateResourceMetadata(resource pcommon.Resource) {
+//
+// When sanitizeExistingLabels is true (APM intake path), existing labels.* /
+// numeric_labels.* keys have only their suffix sanitized (reserved characters
+// replaced). All other unsupported attributes are fully re-normalized.
+// When false (OTel path), all unsupported attributes — including any that
+// already carry a labels.* prefix — are treated as raw keys and re-normalized
+// from scratch.
```
lahsivjar, line 55: [**For reviewers**] This is validated in apm-data, see https://github.com/lahsivjar/apm-data/pull/new/mOTLP-test

processor/elasticapmprocessor/testdata/ecs/elastic_span_db/output.yaml
```diff
@@ -314,6 +277,43 @@ resourceSpans:
         - key: agent.activation_method
           value:
             stringValue: automatic
+        - key: labels.unsupported_key
+          value:
+            stringValue: foo
+        - key: labels.labels_my_value
```
lahsivjar, line 283: [**For reviewers**] The double normalization (`labels.labels*`) is expected, see the prior comment about apm-data behaviour for MIS OTel mode - I dond't change the input data so this is emulating OTel data sent to MIS with attributes set as `labels.xxx`.

processor/elasticapmprocessor/internal/ecs/ecs_translation.go
```diff
@@ -329,105 +357,97 @@ func RemapMetricDataPointAttributesToECSLabels(attributes pcommon.Map) {
...
-func truncatePreservedStringAttribute(attributes pcommon.Map, key string, value pcommon.Value) {
+func truncatePreservedStringAttribute(value pcommon.Value) {
 	if value.Type() != pcommon.ValueTypeStr {
 		return
 	}
-	truncated := sanitize.Truncate(value.Str())
-	if truncated != value.Str() {
-		attributes.PutStr(key, truncated)
+	if truncated := TruncateToECSMaxLength(value.Str()); truncated != value.Str() {
+		value.SetStr(truncated)
```
lahsivjar, line 383: [**For reviewers**] We are only changing the value here - this is okay but what we were doing previously was not okay as we were inserting a new attribute while iterating over the `pcommon.Map`

processor/elasticapmprocessor/internal/ecs/ecs_translation_test.go
```diff
@@ -381,6 +388,87 @@ func TestRemapLogRecordAttributesToECSLabels(t *testing.T) {
...
+				"labels.a3":        "v",
+				"labels.a4":        "v",
+				"labels.a5":        "v",
+				"labels.a6":        "v",
+				"labels.a7":        "v",
+				"labels.a8":        "v",
+				"labels.a9":        "v",
+			},
+			wantAbsent: []string{"a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7", "a8", "a9"},
+		},
```
lahsivjar, line 471: [**For reviewers**] These are the test cases that would fail/cause panic.

processor/elasticapmprocessor/internal/ecs/ecs_translation.go
```diff
@@ -37,17 +37,26 @@ const (
 	ecsAttrOpenCensusExporterVersion = "opencensus.exporterversion"
 )
 
+// kv is a key-value pair for collecting deferred attribute insertions after a RemoveIf pass.
+type kv struct {
```
vigneshshanmugam, line 41: nit: cant we use the otlpcommon.KeyValue ? I havent checked it though, just asking.

processor/elasticapmprocessor/internal/ecs/ecs_translation.go
```diff
@@ -329,105 +357,97 @@ func RemapMetricDataPointAttributesToECSLabels(attributes pcommon.Map) {
...
-		attributes.PutDouble("numeric_labels."+key, value.Double())
+		return kv{k: "numeric_labels." + sanitizedKey, v: pcommon.NewValueDouble(value.Double())}
 	case pcommon.ValueTypeSlice:
 		slice := value.Slice()
 		if slice.Len() == 0 {
-			return
+			return kv{}
 		}
 		switch slice.At(0).Type() {
+		// TODO(lahsivjar): Can we assume all are same type and just use pcommon.Value#CopyTo?
```
vigneshshanmugam, line 404: ++ was wondering the same as the logic seems to be the same for everything.

vigneshshanmugam (approved):
LGTM, thanks for the detailed test-cases, made it easier for the review.

processor/elasticapmprocessor/internal/ecs/ecs_translation.go
lahsivjar: That's from the proto package (and protocommon.KeyValue is also a proto struct), I don't think we need that heavy dependency just for this purpose. Happy to discuss further - can do a followup if we decide later.

processor/elasticapmprocessor/internal/ecs/ecs_translation.go
lahsivjar: Yeah, I am not sure too. I didn't want to do too drastic of a change to start with so will review this a bit later.

vigneshshanmugam: Sounds good.

PR #1231, opened 2026-05-22
[Extension][beats_encoding] feat: Introduce mappings feature

Kavindu-Dodan:
### Overview

This PR enhance the Beats Encoding extension by introducing `mappings` configuration.

Consider following config.yaml portion,

```yaml
extensions:
  beats_encoding/mapped:
    format: json
    unwrap:
      - data
      - items
    data_stream:
      dataset: custom
    input_type: aws-s3
    mappings:
      - source: log
        destination: message
        type: String
      - source: timestamp
        destination: "@timestamp"
        type: Integer
        multiplier: 1000  # seconds → millis
```

When processing following JSON,

```json
{"data": {"items": [{"id": 1, "log": "hello", "timestamp": 1779463864}]}}
```

Mappings will extract internal JSON entries as below,

```
- `message`: `"hello"`
- `@timestamp`: `1779463864000` // Scaler values with multiplier
```

### DIFF
--- a/extension/beatsencodingextension/config.go
+++ b/extension/beatsencodingextension/config.go
@@ -24,6 +24,9 @@ import (
 // Format defines how the incoming raw bytes should be interpreted.
 type Format string
 
+// FieldType defines the supported OTel pcommon value types for field mappings.
+type FieldType string
+
 const (
 	// FormatJSON indicates the input is a JSON document that may contain
 	// wrapped records (use Unwrap to extract them).
@@ -32,6 +35,12 @@ const (
 	// FormatText indicates the input is newline-delimited text where
 	// each line becomes a separate log record.
 	FormatText Format = "text"
+
+	// FieldTypeString maps to pcommon.Map.PutStr.
+	FieldTypeString FieldType = "String"
+
+	// FieldTypeInteger maps to pcommon.Map.PutInt.
+	FieldTypeInteger FieldType = "Integer"
 )
 
 // DataStreamConfig defines the data stream routing attributes.
@@ -69,8 +78,30 @@ type Config struct {
 	// inject custom metadata (e.g., environment, team) into each event.
 	Fields map[string]any `mapstructure:"fields,omitempty"`
 
+	// Mappings defines which JSON keys to extract from each decoded JSON
+	// element and how to store them in the log record body.
+	// Only used when Format is "json" and Unwrap is set.
+	Mappings []FieldMapping `mapstructure:"mappings,omitempty"`
+
 	// prevent unkeyed literal initialization
 	_ struct{}
+}
+
+// FieldMapping defines how a single JSON key is extracted and written
+// to the log record body.
+type FieldMapping struct {
+	// Source is the JSON key to read from the decoded object.
+	Source string `mapstructure:"source"`
+
+	// Destination is the key to write to in the log body map.
+	Destination string `mapstructure:"destination"`
+
+	// Type is the OTel pcommon value type: "String" or "Integer".
+	Type FieldType `mapstructure:"type"`
+
+	// Multiplier scales the numeric value before storing.
+	// Only applies to FieldTypeInteger. A value of 0 means no scaling.
+	Multiplier int64 `mapstructure:"multiplier,omitempty"`
 }
 
 func (c *Config) Validate() error {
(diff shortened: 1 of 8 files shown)


Kavindu-Dodan:
Needs https://github.com/elastic/opentelemetry-collector-components/pull/1232

extension/beatsencodingextension/extension.go
```diff
@@ -228,20 +243,43 @@ func (e *beatsEncodingExtension) newStreamingJSONDecoder(reader io.Reader, opts
...
-				continue
+			if err := e.appendLogRecord(sl, now, eventCreated, data); err != nil {
+				return plog.NewLogs(), err
 			}
 
-			e.appendLogRecord(sl, now, eventCreated, string(trimmed))
 			recordCount++
 			batchHelper.IncrementItems(1)
-			batchHelper.IncrementBytes(int64(len(raw)))
+			batchHelper.IncrementBytes(n)
```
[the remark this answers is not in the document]
MichaelKatsoulis: @Kavindu-Dodan Why is this resolved? Ins't this is a bug?

MichaelKatsoulis (approved):
Approved! Only concern worth looking at is that CodeRabbit's comment

extension/beatsencodingextension/extension.go
```diff
@@ -238,10 +256,36 @@ func (e *beatsEncodingExtension) newStreamingJSONDecoder(reader io.Reader, opts
...
 			}
 
-			e.appendLogRecord(sl, now, eventCreated, string(trimmed))
+			n = int64(len(raw))
+
+			if len(e.config.Mappings) > 0 {
+				var asMap map[string]any
+				if err := json.Unmarshal(trimmed, &asMap); err != nil {
+					return logs, fmt.Errorf("decoding array element: %w", err)
+				}
```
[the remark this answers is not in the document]
Kavindu-Dodan: Fixed with https://github.com/elastic/opentelemetry-collector-components/pull/1231/changes/7ef84869dc604a6b9689378466b9b4a0b3249a9d

vigneshshanmugam (commented):
Not opposed to this. But, do we need this in all the integrations for compatibility reasons? I m wondering if we can short circuit the logic in this and just do this for all of the data and call it like a mode

```yml
beats_encoding:
  format: json
  unwrap: [logEvents]
  cloudwatch_compat: true   # extracts message+timestamp, converts ms→ns
```

extension/beatsencodingextension/config.go
```diff
@@ -69,10 +78,32 @@ type Config struct {
...
+
+	// Destination is the key to write to in the log body map.
+	Destination string `mapstructure:"destination"`
+
+	// Type is the OTel pcommon value type: "String" or "Integer".
+	Type FieldType `mapstructure:"type"`
+
+	// Multiplier scales the numeric value before storing.
+	// Only applies to FieldTypeInteger. A value of 0 means no scaling.
+	Multiplier int64 `mapstructure:"multiplier,omitempty"`
```
vigneshshanmugam, line 104: we would need a upper cieling here if we decide to use explicit mappings approach.

extension/beatsencodingextension/config.go
```diff
@@ -92,5 +123,19 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("data_stream.namespace is required")
 	}
 
+	for i, m := range c.Mappings {
```
vigneshshanmugam, line 126: would the mappings apply even if the formatting is JSON?

Kavindu-Dodan: Mappings configs only applies to JSON. Here the validation does not cross check this fact though. I thought that's not needed.

Kavindu-Dodan:
> Not opposed to this. But, do we need this in all the integrations for compatibility reasons? I m wondering if we can short circuit the logic in this and just do this for all of the data and call it like a mode
> 
> ```yaml
> beats_encoding:
>   format: json
>   unwrap: [logEvents]
>   cloudwatch_compat: true   # extracts message+timestamp, converts ms→ns
> ```

I thought about this option but then went with the generic path. And that's to allow,

- Make the mappings extraction config driven (vs code driven)
- Easy change deployments with no code releases
- Allow usage by other CSPs (We share the extension with other CSPs)

I think generalization is better here as the core logic won't change. It's the way we derive configs. Let me know your view.

vigneshshanmugam (approved):
LGTM. Lets go with as we discussed offline.

Kavindu-Dodan:
> LGTM. Lets go with as we discussed offline.

Thank you @vigneshshanmugam Let's loop back if this model does not work or scale 🙏

PR #1232, opened 2026-05-22
chore: fix std lib & other vulnerabilities

Kavindu-Dodan:
### Overview 

Attempt to fix stdlib vulnerabilities.

Go vulnerability scanner start complaining about standard library errors in 1.25.x variants

For example, 

```
    Excessive resource consumption when printing error string for host
    certificate validation in crypto/x509
  More info: https://pkg.go.dev/vuln/GO-2025-4155
  Standard library
    Found in: crypto/x509@go1.25.4
    Fixed in: crypto/x509@go1.25.5
    Example traces found:
      #1: internal/metadatatest/generated_telemetrytest.go:35:37: metadatatest.NewSettings calls connectortest.NewNopSettings, which eventually calls x509.Certificate.Verify
      #2: internal/metadatatest/generated_telemetrytest.go:35:37: metadatatest.NewSettings calls connectortest.NewNopSettings, which eventually calls x509.Certificate.VerifyHostname
```

And see this build - https://github.com/elastic/opentelemetry-collector-components/actions/runs/26304085983/job/77436324160?pr=1231 

### The Fix

This PR attempts to bump to go 1.26.0 along with toolchain fix to get rid of the vulnerabilities.

- Why 1.26.0 ? There were some modules already in 1.26.0. Hence, upgraded all modules to align with tooling and module needs .


efd6 (approved):

straistaru (approved):

connector/profilingmetricsconnector/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/elastic/opentelemetry-collector-components/connector/profilingmetricsconnector
 
-go 1.25.0
+go 1.26.0
```
florianl, line 3: Go 1.25 still needs to be supported.

florianl (changes requested):

connector/profilingmetricsconnector/go.mod
efd6: Counter proposal here #1234.

florianl: #1234 looks like a more reasonable approach to me. Thanks @efd6

efd6:
@Kavindu-Dodan The counter proposal was merged, so this is no longer required.

Kavindu-Dodan:
> @Kavindu-Dodan The counter proposal was merged, so this is no longer required.

Thank you @efd6 for the alternative fix. I attempted to do something similar but it failed. Anyway this should have been the right approach.

PR #1237, opened 2026-05-27
Improve telemetry of the ratelimit processor

straistaru:
#### Two new metrics

| Metric | Type | When recorded |
|--------|------|---------------|
| `otelcol_ratelimit.delay_duration` | Histogram | Only when a request actually waited (`throttle_behavior: delay`) |
| `otelcol_ratelimit.tokens` | Gauge | Every request — shows current token bucket level, negative means active debt |

#### Richer `decision` attribute on `otelcol_ratelimit.requests`

Previously only `accepted` / `throttled` were distinguishable via the error return. Now every outcome is explicit:

| `decision` | `reason` | Meaning |
|------------|----------|---------|
| `accepted` | `under_limit` | Tokens available, passed immediately |
| `delayed` | `over_limit` | Waited for bucket to refill, no error returned |
| `throttled` | `over_limit` | Rejected — bucket empty in error mode |
| `cancelled` | `over_limit` | Was waiting in delay mode but client context was cancelled |

#### Internal: `RateLimiter.RateLimit` now returns `RateLimitResult`

The interface changed from `RateLimit(ctx, n) error` to `RateLimit(ctx, n) (RateLimitResult, error)`. The result carries `Decision`, `Delay`, `Tokens`, and `ConfigRate` so the processor layer can record all telemetry without the rate limiter needing any OTel imports.

### Test improvements

- `TestConsume_DelayMode` and `TestConsume_DelayMode_ContextCancelled` are now table-driven over all four signal types (logs, metrics, traces, profiles) — previously only logs were covered in delay mode.
- `local_test.go` now asserts `Decision`, `Delay`, and `ConfigRate` on all core behaviour tests, closing a gap where a wrong return value from `local.go` would have gone undetected.

### DIFF
--- a/processor/ratelimitprocessor/local.go
+++ b/processor/ratelimitprocessor/local.go
@@ -77,13 +77,20 @@ func (r *localRateLimiter) RateLimit(ctx context.Context, hits int) (RateLimitRe
 	})
 	state := v.(*keyState)
 
+	// preTokens is the bucket level before this request consumes anything.
+	// PreTokens < 0 means the bucket was already in deficit on arrival —
+	// a reliable signal for genuine sustained throttling (vs. a transient
+	// batching spike that clears before the next request).
+	preTokens := state.limiter.Tokens()
+
 	// makeResult captures the current token level and config for telemetry.
 	makeResult := func(decision Decision, delay time.Duration) RateLimitResult {
 		return RateLimitResult{
-			Decision:   decision,
-			Delay:      delay,
-			Tokens:     state.limiter.Tokens(),
-			ConfigRate: float64(cfg.Rate),
+			Decision:     decision,
+			Delay:        delay,
+			TokensBefore: preTokens,
+			TokensAfter:  state.limiter.Tokens(),
+			ConfigRate:   float64(cfg.Rate),
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


processor/ratelimitprocessor/local.go
```diff
@@ -65,43 +65,62 @@ func (r *localRateLimiter) Shutdown(_ context.Context) error {
...
+			TokensAfter:  state.limiter.Tokens(),
+			ConfigRate:   float64(cfg.Rate),
+		}
+	}
 
 	switch cfg.ThrottleBehavior {
 	case ThrottleBehaviorError:
-		if ok := limiter.AllowN(time.Now(), hits); !ok {
-			return errorWithDetails(errTooManyRequests, cfg)
+		if ok := state.limiter.AllowN(time.Now(), hits); !ok {
```
michalpristas, line 93: i mean yes, but at this time before tokens may be different value right? it's not inside a critical section nor it needs to be just for sake of tokens reporting.
this one also should not consume tokens if it fails but values will be different, it makes sense. if this is in reserveAll it could be in a single block behind a lock and this way it would probably be more accurate
it all depends on what are we trying to observe using before after.

michalpristas (approved):

processor/ratelimitprocessor/local.go
straistaru: Good shout.
I've fixed the accuracy of both `tokens_before` and `tokens_after` and placed them in `reserverAll()`.
I also made it so that these metrics will be emitted for `throttle_behavior: delay` only. It doesn't make sense to have them on `throttle_behavior: error`, as excess events will just be dropped.

michalpristas (approved):
beautiful

PR #1238, opened 2026-05-27
fix(receiver/elasticapmintake): reduce intake batch wait latency

vigneshshanmugam:
## Summary
- Consume non-empty logs, metrics, and traces concurrently within each Elastic APM intake batch.
- Preserve existing request semantics by waiting for all downstream consumers before returning.
- Add configurable `batch_size` for intake v2 processing, keeping the default at `10`.
- Add regression coverage for concurrent signal consumption and configured batch sizing.

## Test plan
- `go test -run 'TestLoadConfig|TestEventsHandlerUsesConfiguredBatchSize|TestConsumeOTelConsumesSignalsConcurrently' -count=1 ./...`
- `go test ./...`

### DIFF
--- a/receiver/elasticapmintakereceiver/receiver_test.go
+++ b/receiver/elasticapmintakereceiver/receiver_test.go
@@ -25,6 +25,7 @@ import (
 	"flag"
 	"io"
 	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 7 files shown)


receiver/elasticapmintakereceiver/receiver_test.go
```diff
@@ -681,6 +735,106 @@ func TestMetadataPropagation(t *testing.T) {
 	}
 }
 
+func TestConsumeOTelConsumesSignalsConcurrently(t *testing.T) {
```
lahsivjar, line 738: **nit, feel free to ignore** I don't think this is worth testing, the complexity of this test is greater than the value it provides.

lahsivjar (approved):

PR #1242, opened 2026-05-28
[receiver/elasticapm]perf: optimize shadow-mask paths and intake-v2 ctx cancel handling

lahsivjar:
## Summary

This PR improves `elasticapmintakereceiver` in two areas:

1. Intake-v2 cancellation/error handling is aligned with MIS behavior.
2. Shadow-mask batching logic is optimized and refactored for readability without changing allocation characteristics.

### Intake-v2 request handling

- Added early request-body close on context cancellation via `context.AfterFunc(...)` to unblock stream reads promptly.
- Updated intake error/status mapping to mirror MIS semantics:
  - client-canceled request context -> `499`
  - canceled/deadline exceeded -> `503`
  - invalid input -> `400` (or `413` when too large)
  - fallback internal errors -> `500`
- Kept `499` precedence behavior so client-cancel classification is not masked by later downstream errors.
- Set `Connection: close` for error responses.

### `processBatch` shadow-mask optimization/refactor

- Kept fast path when no global metadata keys are present (`len(keyIndex) == 0`).
- Kept compact `uint64` mask path for small key sets (`len(keyIndex) <= 64`) and `big.Int` fallback for larger sets.

## Why

- Intake-v2 requests can continue doing unnecessary work after cancellation unless body reads are unblocked promptly.
- MIS-aligned status classification improves consistency across services and observability.
- Shadow-mask code had duplicated branch logic; refactor keeps behavior/perf while improving maintainability.

## Validation

- `go test ./...` in `receiver/elasticapmintakereceiver`
- Lint checks on edited receiver files


receiver/elasticapmintakereceiver/receiver.go
```diff
@@ -184,50 +185,95 @@ func (r *elasticAPMIntakeReceiver) newElasticAPMEventsHandler(ctxFunc func(*http
...
+		processError := func(err error, isRequestContextErr bool) {
 			result.Errors = append(result.Errors, err.Error())
+			// Mirror MIS precedence: once we detect a client-canceled request
+			// (499), keep that status to avoid masking it with downstream errors.
+			if statusCode != statusClientClosed {
+				if errStatusCode := intakeStatusCodeFromErr(err, isRequestContextErr); errStatusCode > statusCode {
+					statusCode = errStatusCode
+				}
+			}
+		}
```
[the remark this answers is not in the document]
lahsivjar: I am a bit conflicted on this since even if client did cancel the connection we also failed with 5xx. IMO, client cancellation should dominate - so in favor of what this comment is implying, but, the current code is mimicking MIS behaviour ([ref](https://github.com/elastic/apm-ingest-service/blob/85932723c0743cd67316d1708e0f7c9c3cb64b83/internal/server/http/elastic.go#L177-L191)) and I would like to continue doing that for now atleast.

receiver/elasticapmintakereceiver/receiver.go
```diff
@@ -184,50 +185,95 @@ func (r *elasticAPMIntakeReceiver) newElasticAPMEventsHandler(ctxFunc func(*http
...
+			// streamErr as a client canceled request context error (MIS behavior).
+			processError(streamErr, requestContextErr != nil && errors.Is(streamErr, context.Canceled))
 		}
-
-		if len(result.Errors) > 0 {
-			statusCode = http.StatusBadRequest
-		} else if streamErr != nil {
-			statusCode = http.StatusInternalServerError
+		if requestContextErr != nil {
+			processError(requestContextErr, true)
```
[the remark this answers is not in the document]
lahsivjar: Same as https://github.com/elastic/opentelemetry-collector-components/pull/1242#discussion_r3317428817

receiver/elasticapmintakereceiver/receiver.go
```diff
@@ -184,50 +185,95 @@ func (r *elasticAPMIntakeReceiver) newElasticAPMEventsHandler(ctxFunc func(*http
...
 		statusCode := http.StatusAccepted
+		ctx := ctxFunc(req)
+		// Ensure stream decoding unblocks quickly once request context is canceled.
+		// In this net/http server handler, Request.Body.Close is safe to call
+		// even if the server later closes it as part of normal request teardown:
+		// the server body Close path is idempotent and concurrent Close unblocks Read.
+		stopBodyClose := context.AfterFunc(ctx, func() {
+			_ = req.Body.Close()
+		})
+		defer stopBodyClose()
```
carsonip, line 196: q: why are we calling it in defer instead of registering the AfterFunc immediately?

receiver/elasticapmintakereceiver/receiver.go
```diff
@@ -280,31 +326,53 @@ func (r *elasticAPMIntakeReceiver) processBatch(ctx context.Context, batch *mode
 		}
 	}
 
-	var shadowedBatches []shadowedBatch
-	var shadowIndex map[string]int // mask.String() → index, created lazily
-
-	for _, event := range *batch {
-		mask, shadowed := eventShadowMask(event, keyIndex)
-		if !shadowed {
+	// Fast path: no metadata-global keys were discovered in the batch.
```
carsonip, line 329: q: since we now have "fast", "less fast", and the original slow path, can we have corresponding tests, or is it a conscious decision to not have implementation-aware tests?

carsonip (commented):
looks right but prefer some test coverage on the new optimized code paths.

non-blocker: the masking logic is getting a bit out of hand and receiver.go is now at 1.2k LoC. Wonder if it'll be more maintainable to split it out which makes it more contained and testable as well

receiver/elasticapmintakereceiver/receiver.go
lahsivjar: The AfterFunc returns a stop function so it basically stops the body close. So, in the case request returns normally, we cancel the callback.

The role of the AfterFunc here is that if the request context is canceled, however, we are still in the process of reading the body then we will close the body. If the request context is not canceled then we deregister the after func (removing the request body close) and let the server close the request body. Does that make sense?

lahsivjar: Also, the stop function will NOT interrupt/deregister if callback has already started - so the afterfunc for us only applies for early context cancelations which are not captured by the upstream

receiver/elasticapmintakereceiver/receiver.go
lahsivjar: Good catch. The fast one and the short-circuit one is already covered but the slow one (big.Int) is no longer covered. Let me add a test.

lahsivjar:
> non-blocker: the masking logic is getting a bit out of hand and receiver.go is now at 1.2k LoC. Wonder if it'll be more maintainable to split it out which makes it more contained and testable as well

I have thought about this and I didn't think this separation is required at the moment. The shadow batch lives at the end of the file and is logically and in code, scoped to decoding data. I was planning on a follow-up to move decode data to a separate package rather than create a package boundary only for shadow-batch.

receiver/elasticapmintakereceiver/receiver.go
carsonip: > The AfterFunc returns a stop function

:facepalm: that's embarrassing for me

carsonip (approved):
lgtm based on passing tests

PR #1243, opened 2026-05-28
[receiver/elasticapm]perf: Make max concurrent decoders configurable

lahsivjar:
Related to: https://github.com/elastic/hosted-otel-collector/issues/2723

Makes the max concurrent decoders configurable to allow more concurrent requests per pod.

### DIFF
--- a/receiver/elasticapmintakereceiver/config.go
+++ b/receiver/elasticapmintakereceiver/config.go
@@ -33,6 +33,10 @@ type Config struct {
 	// BatchSize is the number of intake v2 events decoded and processed together.
 	BatchSize int `mapstructure:"batch_size"`
 
+	// MaxConcurrentDecoders is the maximum number of intake requests whose bodies
+	// can be decoded concurrently.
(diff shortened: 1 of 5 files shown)


carsonip (commented):
code change lgtm but what's the binary `receiver/elasticapmintakereceiver/elasticapmintakereceiver.test`?

lahsivjar:
Oops, good catch. That was a test artifact which was not in gitignore rules! Fixed now.

receiver/elasticapmintakereceiver/config.go
```diff
@@ -64,3 +71,10 @@ func (cfg *Config) batchSize() int {
 	}
 	return cfg.BatchSize
 }
+
+func (cfg *Config) maxConcurrentDecoders() int64 {
+	if cfg.MaxConcurrentDecoders <= 0 {
+		return int64(defaultMaxConcurrentDecoders)
```
carsonip, line 77: non blocker nit: not the biggest fan of setting default in both default config and also on read, but seeing that it is already bad for batchSize, let's fix it later.

carsonip (approved):

PR #1244, opened 2026-05-28
Add prw exporter in otelbench

gizas:
Adds the Prometheus Remote Write Exporter as a local fork with exporter helper enabled.

Related to https://github.com/elastic/hosted-otel-collector/pull/3133
See relevant discussion https://github.com/elastic/hosted-otel-collector/pull/3124#discussion_r3317610778


constanca-m:
I need a bit more context to review this PR.

So from what I understand this exporter is just a fork of [prometheusremotewriter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusremotewriteexporter), but it wants to replace `remote_write_queue + wal` with exporterhelper `sending_queue`. Is this correct? Is that the only change in this fork given the upstream implementation?

And what is the reason we are opening it here in this repository instead of opening the PR upstream with concrete results based on issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33137?

Also, what is our reasoning to go with `sending_queue` instead of what the exporter has? Not saying this is wrong, just want to understand why we are making these decisions.

gizas:
> but it wants to replace remote_write_queue + wal with exporterhelper sending_queue

Yes this is the goal !

> Is that the only change in this fork given the upstream implementation?

Yes, I removed remote_write_queue and added the exporter helper

Plus some small fixes like [suggestion 1](https://github.com/elastic/hosted-otel-collector/pull/3133#discussion_r3311158351) or [suggestion 2](https://github.com/elastic/hosted-otel-collector/pull/3133#discussion_r3311157090)

> what is our reasoning to go with sending_queue

Full details [here](https://github.com/elastic/hosted-otel-collector/issues/2913#issuecomment-4304472162). The reason is that current remote_write_queue does not support metadata_key population in order to support multiple project batching.

> what is the reason we are opening it here in this repository instead of opening the PR upstream with concrete results based on issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33137?

I tried to do it there but as you can see there is some discussion for this. Although with big delays. So final goal is to do it upstream. But until then in order to make things quicker we decided to open it internally first

constanca-m:
Thanks @gizas.

I read the full thread in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33137, and I think the key concerns from @Aneurysm9 are still not fully resolved.

From https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33137#issuecomment-3973998073:

> The remote write queue and WAL exist in this exporter to help ensure the ordered write requirements Prometheus has can be satisfied.

and

> First and foremost we need to understand how the in-order production guarantees will be maintained going forward. Without that I don't think anything else really matters.

I agree this is the core question.

I also checked `exporterhelper` code. My understanding is:
- normal retries happen in the retry sender loop, not by re-enqueueing every failed request;
- ordering risk still exists with multiple consumers and with restart handling of in-flight persistent-queue items.

So the concern remains valid: we still need a concrete ordering and durability story before migration.

If we want to proceed, I suggest an upstream-first, feature-gated plan:

1. Add a feature gate to enable `sending_queue` path for PRW, default off.
2. Keep current `remote_write_queue` + `wal` path as default during evaluation.
3. Define explicit acceptance criteria before merge:
   - ordering behavior (single consumer and multi-consumer),
   - retry classification (retryable vs permanent),
   - restart durability semantics.
4. Run this for a fixed window (for example, two releases), collect test and operational data, then decide whether to switch defaults or stop.

**If `exporterhelper` cannot satisfy PRW requirements, we should not keep a long-lived fork because drift will be expensive to maintain.**

> The reason is that current remote_write_queue does not support metadata_key population in order to support multiple project batching.

If this is the reason, then why not open an issue upstream to fix this? Why are we going with a completely different approach on the queue?

**I have never worked with the prometheus remote write exporter**, but looking at the full discussion I understand the codeowner concerns and questions, because I too feel a bit confused over our decisions:
- Why change queues?
- What do we lose/gain? Is the complexity even worth it?
- Why don't we tackle our issue directly instead of wanting a full queue change?

constanca-m:
> The reason is that current remote_write_queue does not support metadata_key population in order to support multiple project batching.

The changes for this seem to be minimal. According to my quick draft with codex 5.3:


- Added include_metadata_keys to PRW config:
  - exporter/prometheusremotewriteexporter/config.go
  - exporter/prometheusremotewriteexporter/config.schema.yaml
  - exporter/prometheusremotewriteexporter/README.md

- Propagated configured client metadata keys to outbound remote write HTTP headers:
  - exporter/prometheusremotewriteexporter/exporter.go: uses client.FromContext(ctx).Metadata.Get(key) and adds header values to the request.

gizas:
So we synced internally and decided:
- To close existing PR
- To open PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48767 to add the summport of metadata keys in the current remote_queue. This probably creates a small difference with queues that we have in esexporter and otlp but until the decision to add the exporterhelper support in PRW is finalised probably is the quickest solution
- At the moment I updated https://github.com/elastic/hosted-otel-collector/pull/3133 to use the upstream PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48767

Upstream main issue for [PRW](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33137#issuecomment-4577020119) updated as well

PR #1250, opened 2026-06-02
[Otelbench] Add support for prometheus Remote Write Exporter

gizas:
Adds the prometheus remote write exporter inside the otelbench

Related to https://github.com/elastic/hosted-otel-collector/issues/2913

We would like to be able to produce and be able to send metrics to a prometheus remote write endpoint


.github/workflows/checks.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: stable
+          go-version: '1.26.4'
```
gizas: For the reviewer: 
To make the tests pass:
https://github.com/elastic/opentelemetry-collector-components/actions/runs/26868691648/job/79238305324?pr=1250

constanca-m: That's a bit awkward fix, can we have a separate PR for it? My fear is that we will end up with a stale go version if we decide to fix it

constanca-m (commented):
If we fix the go version, then the PR looks good!

@gizas Can you tell me which PRs are pending review and the order to get them fixed? To reach load testing conditions

.github/workflows/checks.yml
gizas: Constanca I think easiest path for now is: https://github.com/elastic/opentelemetry-collector-components/issues/1252

constanca-m (commented):
I'm ok with approving, but we are still using a fixed go version in `checks.yml` which is not the best strategy long-term

constanca-m (approved):

PR #1254, opened 2026-06-03
[receiver/elasticapm]Add GET / handler with fake versioning

lahsivjar:
## Summary

Returns {"version":"8.9.0"} with HTTP 200 for APM agent server discovery, matching MIS behaviour. Registered as "GET /{$}" for exact path matching — without the {$} anchor Go ServeMux treats / as a rooted subtree match.

Related to: https://github.com/elastic/hosted-otel-collector/issues/3252

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ bin/
 .tools/
 .release/
 _build/
+*.test
 
 ### IntelliJ IDEA ###
 .idea/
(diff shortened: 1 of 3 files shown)


receiver/elasticapmintakereceiver/elasticapmintakereceiver.test
lahsivjar, line 1: [**For reviewers**] Unfortunately, this got added by me in the previous PR and will always remain in history 😢 . It was added by mistake and the file is basically a test binary. I have updated the .gitignore to not add this by mistake in the future.

raultorrecilla (approved):

carsonip (approved):

PR #1255, opened 2026-06-03
[Prometheus Remote Write] Adds additional config flag  for prw exporter

gizas:
This PR adds `--exporter-prometheusremotewrite` benchmark flag for Prometheus Remote Write load testing

We had the following errors:
```bash
make run-otelbench profile=prw-local

invalid configuration: service::pipelines::traces: references exporter "otlp" which is not configured
invalid configuration: service::pipelines::traces: references exporter "otlphttp" which is not configured
BenchmarkOTelbench/traces-otlphttp-1 	       0	               NaN ns/op
BenchmarkOTelbench/mixed-otlp-1      	       0	               NaN ns/op
invalid configuration: service::pipelines::logs: references exporter "otlp" which is not configured
BenchmarkOTelbench/mixed-otlphttp-1  	       0	               NaN ns/op
invalid configuration: service::pipelines::logs: references exporter "otlphttp" which is not configured

```

The above errors were present when I was trying to pass traffic to PRW exporter and realised that even without otlp/otlphttp being set those were trying to be used in code. This PR adds the additonal flag.

### Testing:

```bash
cd loadgen/cmd/otelbench
go build -o otelbench .\n
./otelbench \                
  --config $(pwd)/config-prw.yaml \
  --exporter-otlp=false \
  --exporter-otlphttp=false \
  --exporter-prometheusremotewrite=true \
  --metrics=true \
  --logs=false \
  --traces=false \
  --mixed=false \
  --concurrency=1 \
  --test.benchtime=1m \
  --test.count=1
...
```

See ingestion for prometheus happens e2e

Screenshot 2026-06-03 at 2 32 12 PM


loadgen/cmd/otelbench/config-prw.yaml
```diff
@@ -0,0 +1,40 @@
+receivers:
```
gizas, line 1: I decided to add this a future ref

loadgen/cmd/otelbench/flags.go
```diff
@@ -134,6 +135,7 @@ func Init() error {
 
 	flag.BoolVar(&Config.ExporterOTLP, "exporter-otlp", true, "benchmark exporter otlp")
 	flag.BoolVar(&Config.ExporterOTLPHTTP, "exporter-otlphttp", true, "benchmark exporter otlphttp")
+	flag.BoolVar(&Config.ExporterPRW, "exporter-prometheusremotewrite", false, "benchmark exporter prometheusremotewrite")
```
constanca-m, line 138: I don't understand why we would ever want these flags? They seem to only cause maintenance. We should want to benchmark every exporter that is in use in `config.yaml`

constanca-m (commented):
Approving, but not really a fan of how we are using flags to benchmark otlp, otlp/http and prometheusremotewrite. Seems really cumbersome if we want to add more exporters in the future

constanca-m (approved):

loadgen/cmd/otelbench/flags.go
gizas: Will also try to have a look after this PR to see if we can improve it

PR #1262, opened 2026-06-12
[receiver/akamaisiemreceiver] - Resolve storage extensions renamed by Fleet, fixing cursor persistence

### LINKED ISSUE #1261: [receiver/akamaisiemreceiver] - Cursor persistence cannot be enabled under Fleet-managed agents
## What happens

The receiver keeps a cursor — a bookmark of where it stopped reading the Akamai SIEM API — so that after a restart it can pick up where it left off. The cursor is saved through a storage extension (the collector's standard state-persistence mechanism, e.g. [file_storage](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/699a7669a8fe6647269c01e44d97eeb1e645b997/extension/storage/filestorage/README.md)), which the receiver points to by name in its config ([config.go#L92-L95](https://github.com/elastic/opentelemetry-collector-components/blob/31a5c7f560d0d179f48763590b0cc30cc70070ac/receiver/akamaisiemreceiver/config.go#L92-L95)):

```yaml
extensions:
  file_storage:
    directory: /var/lib/otelcol/akamai
receivers:
  akamai_siem:
    storage: file_storage   # "save my bookmark in file_storage"
service:
  extensions: [file_storage]
```

This works when you run the collector yourself. When the same config is delivered through Fleet, the receiver crashes at startup with `storage extension "file_storage" not found` ([receiver.go#L304-L318](https://github.com/elastic/opentelemetry-collector-components/blob/31a5c7f560d0d179f48763590b0cc30cc70070ac/receiver/akamaisiemreceiver/receiver.go#L304-L318)).

The receiver has not shipped in any release yet — it first ships with Elastic Agent 9.5, currently in snapshots. Unless this is fixed, 9.5 goes out with no way to enable cursor persistence on a Fleet-managed agent: every agent restart makes the receiver re-download up to 12 hours of events (deduplicated downstream, but wasted work and noisy 409 logs).

## Why it happens

An agent can run several integrations at once, and each one brings its own collector config. Fleet merges them all into a single config. To stop two integrations from clashing over the same names, Fleet renames every component by appending a unique per-policy suffix ([otel_collector.ts#L118](https://github.com/elastic/kibana/blob/822705e2991fa6137495dc985cfaea13dff6bde8/x-pack/platform/plugins/shared/fleet/server/services/agent_policies/otel_collector.ts#L118), extensions renamed at [#L162](https://github.com/elastic/kibana/blob/822705e2991fa6137495dc985cfaea13dff6bde8/x-pack/platform/plugins/shared/fleet/server/services/agent_policies/otel_collector.ts#L162)):

```yaml
# what the integration declared          # what the agent actually receives
extensions:                              extensions:
  file_storage: {...}                      file_storage/akamai-1234-siem: {...}
receivers:                               receivers:
  akamai_siem:                             akamai_siem/akamai-1234-siem:
    storage: file_storage                    storage: file_storage   # <- stale!
```

Fleet updates most places that mention a renamed component — pipeline lists and `service.extensions` ([otel_collector.ts#L187-L189](https://github.com/elastic/kibana/blob/822705e2991fa6137495dc985cfaea13dff6bde8/x-pack/platform/plugins/shared/fleet/server/services/agent_policies/otel_collector.ts#L187-L189)), and `auth.authenticator` references inside component bodies ([#L506-L528](https://github.com/elastic/kibana/blob/822705e2991fa6137495dc985cfaea13dff6bde8/x-pack/platform/plugins/shared/fleet/server/services/agent_policies/otel_collector.ts#L506-L528)). But it does not know that the receiver's `storage:` value is also a name reference — the word "storage" does not appear anywhere in that file — so it keeps the old name. The receiver then looks for an extension called `file_storage`, which no longer exists, and fails.

## The Idea behind the fix

Fleet gives **the same suffix to everything declared in one stream** — the suffix is built once per stream and applied to every component section ([otel_collector.ts#L118](https://github.com/elastic/kibana/blob/822705e2991fa6137495dc985cfaea13dff6bde8/x-pack/platform/plugins/shared/fleet/server/services/agent_policies/otel_collector.ts#L118)). The receiver and the extension it shipped with always end up with matching instance names:

```
akamai_siem/akamai-1234-siem
file_storage/akamai-1234-siem
```

So from inside the receiver, "the storage extension that was declared next to me" is findable with certainty: it is the storage extension whose instance name matches my own. No guessing involved.

## Proposal

Two small changes to how the receiver resolves its storage extension (today's exact-name lookup: [receiver.go#L304-L318](https://github.com/elastic/opentelemetry-collector-components/blob/31a5c7f560d0d179f48763590b0cc30cc70070ac/receiver/akamaisiemreceiver/receiver.go#L304-L318)):

1. **Follow the rename.** If `storage: file_storage` finds no exact match, look for a `file_storage/<something>` whose instance name matches the receiver's own (the Fleet case above), or — if there is exactly one `file_storage/*` — use that. Anything still unresolved or ambiguous fails at startup with a list of what was available, same as today. References that already name an instance (`file_storage/foo`) never fall back.

2. **Auto-bind when no `storage:` is set.** If the receiver itself has a suffixed name (which only happens under Fleet) and exactly one storage extension carries the same suffix, use it. A plain unnamed receiver — the standalone case — never does this, so existing standalone behavior does not change at all.

Auto-bind matters for rollout even though the receiver has not shipped yet: the integration package can ship just the extension, with no `storage:` line. Agent builds whose receiver predates this change — earlier 9.5 snapshots, or 9.5.0 itself if this lands in a later patch — leave the extension unused and keep working; builds that include it get persistence. If the package shipped the `storage:` reference instead, every receiver build without the fallback would crash on it.

## Why not fix it elsewhere

- **In Kibana** — teaching Fleet that `storage:` is a name reference, the same way it already handles `auth.authenticator` ([otel_collector.ts#L506-L528](https://github.com/elastic/kibana/blob/822705e2991fa6137495dc985cfaea13dff6bde8/x-pack/platform/plugins/shared/fleet/server/services/agent_policies/otel_collector.ts#L506-L528)), is the cleanest fix and I will file that separately. But the receiver should not be blocked on it, and a receiver-side fallback also covers policies generated by any Kibana version that lacks that rewrite.

ShourieG:
## Proposed Commit Message
```
receiver/akamaisiemreceiver: resolve storage extensions renamed by Fleet

Fleet renames every component in a policy stream to
<type>/<stream-suffix> without rewriting the receiver's storage
reference, so a bare `storage: file_storage` dangles and Start fails.
Resolve bare references via instance-name match, then unique type
match; explicit type/name references never fall back. When storage is
unset, auto-bind a single storage-capable extension sharing the
receiver's instance name so integration packages can enable cursor
persistence by declaring the extension alone. Standalone (unnamed)
receivers are unchanged.
```

## Issues
 - Closes https://github.com/elastic/opentelemetry-collector-components/issues/1261 
 
 ## Local Testing

Everything was tested against a local 9.5.0-SNAPSHOT stack with the akamai
integration installed, polling a mock Akamai SIEM endpoint with real EdgeGrid
credentials.

### 1. Reproduce the bug

Before touching any code, we built a baseline binary from current source and ran it
with a config shaped exactly like Fleet's output: receiver and `file_storage` both
suffixed, receiver body still saying `storage: file_storage`. It fails the way a
Fleet-managed agent fails today:

```
Error: cannot start pipelines: ... storage extension "file_storage" not found
```

### 2. Verify the fix on the same config

The fixed binary starts cleanly on the byte-identical config and says how it healed
the reference:

```
info  resolved storage extension via instance-name match
      {"configured": "file_storage", "resolved": "file_storage/akamai-0a1b2c3d-siem_otel"}
```

The cursor file appears on disk and advances with every poll.

### 3. Test the shape the integration package will actually ship

Same config with no `storage:` line at all — the package template deliberately omits
it. The receiver logs `auto-bound storage extension declared in this stream`, and
after a kill-and-restart it logs `loaded persisted cursor` and picks up where it left
off instead of re-fetching the lookback window. The ambiguous case (two candidate
extensions, neither matching) fails with an error listing both, as intended.

### 4. Check against real Kibana

We installed the akamai package into the stack, enabled the otelcol input with
storage on, and pulled the policy Fleet actually renders. It confirmed everything the
fix relies on: receiver and extension share the identical suffix, the receiver body
carries no `storage:` key, and the `${env:STATE_PATH:-...}` directory expression
passes through untouched.

### 5. Prove it works inside a container

We built the collector Docker image from this branch and ran it with the real suffix
Kibana rendered in step 4. The state file showed up inside the container at
`/usr/share/elastic-agent/state`, survived `docker restart` (cursor loaded on the way
back up), and moved when `STATE_PATH` was overridden.

### 6. Unit tests

`storage_resolution_test.go` adds 13 cases covering every resolution and auto-bind
path, including the guard that an unnamed receiver never auto-binds — so the
standalone behavior stays exactly as documented. `go test`, `go vet`, and `make lint`
are clean.

### DIFF
--- a/receiver/akamaisiemreceiver/README.md
+++ b/receiver/akamaisiemreceiver/README.md
@@ -414,7 +414,7 @@ The example pipelines compose standard Collector components. Each one is indepen
 
 **`file_storage` extension** (optional) — `extensions.file_storage`
 
-- `directory` — where the BoltDB-backed key-value file lives. Used by both the receiver's cursor and the exporter's persistent sending queue. Both live in the same DB but in separate keyspaces — no conflict.
+- `directory` — where the BoltDB-backed key-value file lives. Used by both the receiver's cursor and the exporter's persistent sending queue. Both live in the same DB but in separate keyspaces — no conflict. **The directory must be writable by the collector process**: `file_storage` fails at service initialization when it cannot create or write the directory (even with `create_directory: true`, e.g. a non-root container user and a path outside any writable mount), and an extension that fails to start takes the whole collector down — no receiver-side handling can degrade gracefully around it, because the failure happens before receivers start.
 - `compaction.on_rebound` — when enabled, BoltDB is compacted after a spike (queue drained from large to small). Without this, the file only grows. Strongly recommended for long-running deployments.
 - `fsync` — fsync each write. Trades throughput for durability. Set to `true` if a power loss must not lose the last few queued events; leave default for higher-throughput-but-eventually-durable setups.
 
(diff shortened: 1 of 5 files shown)


receiver/akamaisiemreceiver/README.md
```diff
@@ -664,12 +664,82 @@ The `_id` (`1774777751-zW7QLk7MmGNbuha8XEVw2iitcJc=`) is set by the ingest pipel
...
+without any receiver-body reference. Unnamed receivers — the standalone
+shape — never auto-bind: keep using the explicit `storage:` reference.
+
+Auto-bind failures (no candidate, multiple candidates, client creation error)
+only disable persistence — a config without `storage:` never fails to start
+because of them. Note the limit of that guarantee: it covers receiver-side
+resolution only. A declared `file_storage` extension that cannot start —
+typically an unwritable `directory` — fails the collector at service
+initialization, before any receiver runs (see the `file_storage` notes in the
+components section).
```
efd6, line 704: This troubles me; spooky action at a distance is a future support load waiting to happen. I can absolutely see cases where a user has accidentally configured auto-bind behaviour (because it worked) and then at some time in the future add another case (again, because the last one worked) and now their storage fails.

efd6 (commented):
I think this should be done in kibana; it's a three line fix. I'm not convinced of the blocking argument, and the fix here only fixes the issue for akamaisiemreceiver while the fix at kibana would fix the issue for *all* receivers that make use of a "storage" (though see note below for a more correct global fix).

I also am very uncomfortable with the addition of the auto-bind feature. If it _is_ decided that the fix should go here, that should be a separate addition (it's a new feature, not a bug fix).

---

Global fix.

Really, the definition of receivers should have a boolean saying that they have a reference to a real-world object that can only have a single owner. This would prevent the need for matching "storage" on the key and avoid the whack-a-mole game on key values, and the potential for foot-gun issue due to incorrect name matching.

ShourieG:
@efd6, [PR Fix ](https://github.com/elastic/kibana/pull/273330) has been raised in the Kibana repo, so this PR will be closed in soon once that goes through.

ShourieG:
Resolved by https://github.com/elastic/kibana/issues/273281

PR #1265, opened 2026-06-16
Fix otelbench multi-arch image publishing

constanca-m:
Currently, the release workflow builds `linux/amd64` and `linux/arm64` in separate matrix jobs, but both jobs push the same tag, `v0.8.0`:

```yaml
matrix:
  platform: ["linux/amd64", "linux/arm64"]

ko build --bare --platform=${{ matrix.platform }} --tags ${{ env.OTELBENCH_VERSION }}
```

That does not publish one multi-arch manifest. It publishes the same tag twice. Whichever matrix job pushes last owns `docker.elastic.co/observability-ci/otelbench:v0.8.0`.

This PR fixes that to support both architectures.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,6 +2,8 @@
 * @elastic/otel-devs
 
 # Per-component owners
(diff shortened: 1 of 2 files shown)


marclop (approved):
Thanks!

carsonip (approved):
nice catch

PR #1281, opened 2026-06-17
[bug]Refactor and fix data_stream attribute handling

lahsivjar:
## Summary

Fix two `data_stream` routing divergences between MIS and mOTLP on the `.apm.` endpoint.

**Namespace preservation** (`data_stream.namespace`): mOTLP was unconditionally writing `"default"`,
ignoring any namespace set by the user as an OTLP resource attribute. All five encode functions
(`EncodeDataStream`, `EncodeErrorDataStream`, `internalMetricDataStream`,
`internalIntervalMetricDataStream`) now write `"default"` only when the attribute is absent.

**Dataset preservation** (`data_stream.dataset`): mOTLP was unconditionally overwriting
`data_stream.dataset` with the computed default (e.g. `"apm"`, `"apm.app.<service>"`), ignoring
any value set by the user. `EncodeDataStream` now reads existing resource attributes via
`ecs.TranslateResourceMetadata` (as `ResourceAttrContext`) before encoding and skips the write
when a dataset is already present. `EncodeErrorDataStream` retains its unconditional write of
`"apm.error"`, matching MIS behaviour.

## Tests

- `TestDataStreamEncoders` — unit tests for all encode functions covering namespace/dataset
  preservation, defaults, and service-name normalisation. Refactored to use `map[string]any`
  attrs and `Attributes().FromRaw()`.
- `TestECSMappingModeTriggersDataStreamRouting` — existing integration test validates namespace
  preservation end-to-end for all three signal types.
- Three new golden tests (`ecs_span_custom_dataset`, `ecs_log_custom_dataset`,
  `ecs_metric_custom_dataset`) validate dataset preservation through the full ECS pipeline for
  traces, logs, and metrics.
- Fixed three existing golden test inputs (`elastic_log`, `elastic_metric`, `elastic_span_db`)
  that carried a `data_stream.dataset: "default"` placeholder which was previously harmless but
  caused the new conditional logic to preserve `"default"` instead of computing the proper value.

Related to: https://github.com/elastic/hosted-otel-collector/issues/3324

### DIFF
--- a/processor/elasticapmprocessor/internal/enrichments/enricher.go
+++ b/processor/elasticapmprocessor/internal/enrichments/enricher.go
@@ -129,9 +129,8 @@ func ecsAPMConfig(cfg config.Config) config.Config {
 // resource. This is common across all signal types (traces, logs, metrics)
 // in ECS mode.
 func ecsPreProcessResource(ctx context.Context, resource pcommon.Resource, dataStreamType string, serviceNameInDataStreamDataset bool, hostIPEnabled bool, sanitizeExistingLabels bool) {
-	ecs.TranslateResourceMetadata(resource, sanitizeExistingLabels)
-	ecs.ApplyResourceConventions(resource)
-	routing.EncodeDataStream(resource, dataStreamType, serviceNameInDataStreamDataset)
+	resourceContext := ecs.TranslateResourceMetadata(resource, sanitizeExistingLabels)
+	routing.EncodeDataStream(resource, dataStreamType, serviceNameInDataStreamDataset, resourceContext)
 	if hostIPEnabled {
 		ecs.SetHostIP(ctx, resource.Attributes())
 	}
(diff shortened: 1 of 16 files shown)


carsonip (commented):
why is there change around host and k8s fields? It is not explained in the PR description

lahsivjar:
> why is there change around host and k8s fields? It is not explained in the PR description

Just refactoring, the code does too many gets whereever it wants and this is an attempt to unify it. It should not change the behaviour around the k8s and host name fields.

carsonip (approved):
lgtm thanks for the tests

processor/elasticapmprocessor/internal/enrichments/enricher.go
```diff
@@ -129,9 +129,8 @@ func ecsAPMConfig(cfg config.Config) config.Config {
 // resource. This is common across all signal types (traces, logs, metrics)
 // in ECS mode.
 func ecsPreProcessResource(ctx context.Context, resource pcommon.Resource, dataStreamType string, serviceNameInDataStreamDataset bool, hostIPEnabled bool, sanitizeExistingLabels bool) {
-	ecs.TranslateResourceMetadata(resource, sanitizeExistingLabels)
-	ecs.ApplyResourceConventions(resource)
```
lahsivjar, line 133: @carsonip this is the main point of the refactor you asked about earlier, both these translate resource metadata but for some reason are 2 separate functions and require more Get's over the resource attribute due to this design. The changes around k8s and host basically merges them.

PR #1291, opened 2026-06-23
[beatsencodingextension] Add CSV format support and a baseline @timestamp

MichaelKatsoulis:
This PR adds a CSV decoder in beatsencodingextension and a baseline timestamp

When configured `format: csv`  it decodes CSV input the same way the Beats `aws-s3 decoding.codec.csv codec` does.
The first record is the header  and each subsequent record becomes a log record whose message is a JSON object keyed by the header. 

Also in this PR:
- Every record now gets a default `@timestamp` (the time it was decoded, like a
  Beats doc). If an integration pipeline sets its own `@timestamp`, that wins.
  This applies to all formats (text/json/csv), not just CSV. It's needed because
  some pipelines (e.g. Netskope) only set `@timestamp` on certain paths and
  otherwise expect the input to have set it already — without a default, those
  docs get rejected by the data stream.
- The `csv.*` options use the same names as the Beats aws-s3 `decoding.codec.csv`
  settings (comma, comment, fields_names, lazy_quotes, trim_leading_space).
- Updated the README to cover the `csv` format, the options, and the `@timestamp` change.


MichaelKatsoulis:
Closing as PR has been reopened from a fork

PR #1300, opened 2026-06-30
Upgrade prometheus to v0.312.0 to fix GO-2026-5710/5381/5264/5662-pre…

lahsivjar:
…decessor

The PR updates the Makefile.Common as per upstream to allow ignoring vulns and ignores [GO-2026-5662](https://pkg.go.dev/vuln/GO-2026-5662) as per https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49343/changes PR upstream


ChrsMark (approved):
LGTM, looks like what upstream does too: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/Makefile.Common#L280-L305

PR #1314, opened 2026-07-07
apikeyauth: add metadata conditions

constanca-m:
Add support for conditional application privileges via `require_metadata`/`exclude_metadata` and for optional dynamic resources and cache-key metadata through a `required: false` flag, so privilege checks only apply when the relevant client metadata is present or absent, and consolidate the application-privilege request tests into a single table-driven test with added config validation coverage.

---

For @andrewvc PoC this:

```yaml
apikeyauth:
  cache:
    key_metadata: [x-elastic-target-id]
  application_privileges:
    - application: apm
      privileges: ["event:write"]
      resources: ["-"]

apikeyauth/sources:
  cache:
    key_metadata: [x-elastic-target-id, x-elastic-source-id]
  application_privileges:
    - application: ingest
      privileges: ["write"]
      dynamic_resources:
        - metadata: x-elastic-source-id
          format: "source:%s"

source_auth:
  sources: apikeyauth/sources
  fallback: apikeyauth

authmiddleware:
  auth: source_auth
```

Becomes:
```yaml
apikeyauth:
  cache:
    key_metadata:
      - metadata: x-elastic-target-id
      - metadata: x-elastic-source-id
        required: false
  application_privileges:
    - exclude_metadata: [x-elastic-source-id]
      application: apm
      privileges: ["event:write"]
      resources: ["-"]

    - require_metadata: [x-elastic-source-id]
      application: ingest
      privileges: ["write"]
      dynamic_resources:
        - metadata: x-elastic-source-id
          format: "source:%s"

authmiddleware:
  auth: apikeyauth
```

### DIFF
--- a/extension/apikeyauthextension/config.go
+++ b/extension/apikeyauthextension/config.go
@@ -103,6 +103,14 @@ type ClientRetryConfig struct {
 }
 
 type ApplicationPrivilegesConfig struct {
+	// RequireMetadata lists client metadata keys that must be present for
+	// this application privilege check to apply.
+	RequireMetadata []string `mapstructure:"require_metadata,omitempty"`
+
+	// ExcludeMetadata lists client metadata keys that must be absent for
+	// this application privilege check to apply.
+	ExcludeMetadata []string `mapstructure:"exclude_metadata,omitempty"`
+
 	// Application holds the name of the application for which
 	// privileges are defined, e.g. "apm".
 	Application string `mapstructure:"application"`
@@ -134,6 +142,17 @@ type DynamicResource struct {
 	// the metadata value. Must contain exactly one %s placeholder.
 	// Defaults to "%s" if empty.
 	Format string `mapstructure:"format"`
+
+	// Required controls whether missing Metadata is an error. Defaults to true.
+	Required *bool `mapstructure:"required,omitempty"`
+}
+
+type MetadataKeyConfig struct {
+	// Metadata holds the client metadata key to include in the cache key.
+	Metadata string `mapstructure:"metadata"`
+
+	// Required controls whether missing Metadata is an error. Defaults to true.
+	Required *bool `mapstructure:"required,omitempty"`
 }
 
 type CacheConfig struct {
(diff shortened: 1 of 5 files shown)


extension/apikeyauthextension/config.go
```diff
@@ -134,6 +142,17 @@ type DynamicResource struct {
 	// the metadata value. Must contain exactly one %s placeholder.
 	// Defaults to "%s" if empty.
 	Format string `mapstructure:"format"`
+
+	// Required controls whether missing Metadata is an error. Defaults to true.
+	Required *bool `mapstructure:"required,omitempty"`
+}
+
+type MetadataKeyConfig struct {
```
constanca-m, line 150: This is a breaking change, but only MOTel is using it: https://github.com/search?q=org%3Aelastic+key_metadata%3A&type=code, so it would be easy to fix

axw:
```yaml
apikeyauth:
  cache:
    key_metadata:
      - metadata: x-elastic-target-id
      - metadata: x-elastic-source-id
        required: false
```

Do we need `x-elastic-source-id` to be optional? Perhaps it would be simpler if legacy requests are authorised with a separate apikeyauth extension, with the old static privileges?

If we can get away without making metadata optional, I think we could collapse static/dynamic resources to a single path using `text/template`. Something like:

```yaml
  # legacy path
  application_privileges:
    - application: apm
      privileges: ["event:write"]
      resources: ["-"]
```

```yaml
  # new path
  application_privileges:
    - application: ingest
      privileges: ["write"]
      resources: ["source:{{index .Metadata \"X-Elastic-Source-Id\"}}"]
```

WDYT?

constanca-m:
@axw That's ok with me.

I was trying to avoid adding a middleware whose only job is to select between two `apikeyauth` extensions.

If we go that direction, I think the config would look roughly like this:

```yaml
apikeyauth:
  cache:
    key_metadata: [x-elastic-target-id]
  application_privileges:
    - application: apm
      privileges: ["event:write"]
      resources: ["-"]

apikeyauth/sources:
  cache:
    key_metadata: [x-elastic-target-id, x-elastic-source-id]
  application_privileges:
    - application: ingest
      privileges: ["write"]
      resources: ["source:{{index .Metadata \"X-Elastic-Source-Id\"}}"]

conditional_middleware:
  conditions:
    - when:
        metadata_exists: x-elastic-source-id
      use: apikeyauth/sources
  fallback: apikeyauth

authmiddleware:
  auth: conditional_middleware
```

(I'm trying to make `conditional_middleware` generic, maybe we will have use for it somewhere else).

axw:
@constanca-m that proposed config LGTM

axw:
Actually: I think conditional_middleware should wrap authmiddleware? Otherwise it's more like conditional_auth, but that feels less generalisable

constanca-m:
Sounds good! @axw I am going to close this PR in favor of the new approach

PR #1316, opened 2026-07-07
receiver/elasticapmintakereceiver: pool NDJSON stream decoders

lahsivjar:
HandleStream allocated a new NDJSONStreamDecoder per request, including a
bufio buffer sized to max_event_size (1MiB by default). This single
allocation accounted for ~34% of allocation volume and a constant ~1MiB/op
regardless of payload size.

Reuse decoders via a sync.Pool, resetting them on reuse and discarding
pooled decoders whose buffer size does not match the configured
max_event_size.

Benchmarks (Apple M4 Pro, no-op consumer):

| Metric | main | branch | Δ |
| --- | --- | --- | --- |
| time/op | 160.3µs | 65.5µs | -59% |
| throughput | 43.8 MiB/s | 107.0 MiB/s | +145% |
| bytes/op | 1.33 MiB | 111 KiB | -92%  |


internal/exporter/prometheusremotewriteexporter/exporter.go
```diff
@@ -477,7 +477,9 @@ func (prwe *prwExporter) execute(ctx context.Context, buf []byte) error {
 		}
 		defer func() {
 			_, _ = io.Copy(io.Discard, resp.Body)
-			resp.Body.Close()
+			if err := resp.Body.Close(); err != nil {
+				prwe.settings.Logger.Warn("failed to close response body", zap.Error(err))
+			}
```
lahsivjar, line 482: [**For reviewers**]: This is to fix the lint error in the prometheus remote write exporter which is failing the build.

receiver/elasticapmintakereceiver/internal/ndjsondecoder/stream.go
```diff
@@ -56,7 +61,18 @@ func HandleStream(
 	logger *zap.Logger,
 	consumer BatchConsumer,
 ) (int, []error) {
-	dec := NewNDJSONStreamDecoder(body, maxLineLength)
+	dec, ok := decoderPool.Get().(*NDJSONStreamDecoder)
+	if ok && dec.lineReader.maxLineLength == maxLineLength {
```
lahsivjar, line 65: [**For reviewers**] The `maxLineLength` check is a bit unnecessary IMO but Claude was very insistent in keeping this and I couldn't reason otherwise as it correctly pointed out that this is a correctness check requirement and will keep this safe even for future.

carsonip: i don't exactly like how it is poking into implementation details of the dec, but it is not a blocker

carsonip (approved):

PR #1320, opened 2026-07-10
connector/elasticapmconnector: derive agent.name for metrics and logs signals

### LINKED ISSUE #1319: elasticapmconnector: agent.name defaults to "unknown" for service_summary (and other non-trace-derived) aggregated metrics
### Component

`connector/elasticapmconnector`

### What happened

The same service instance ends up with two different `agent.name` values depending on signal type: traces get the correctly derived value (e.g. `opentelemetry/nodejs`), while metrics/logs-derived aggregates (`service_summary`, service transaction, span destination, etc.) get the literal string `unknown`.

This comes from `signaltometricsConfig()`, which hardcodes `"unknown"` as the `agent.name` default for these aggregates:

https://github.com/elastic/opentelemetry-collector-components/blob/9e457b5f2fec8751ebe253e849dcb63c287ef1a6/connector/elasticapmconnector/config.go#L245-L250

`telemetry.sdk.language` is already carried through as a resource attribute on the same aggregate (`commonResourceAttributes`, line 242), so the information needed to derive a real `agent.name` (the same way `elasticapmprocessor.setAgentName` does it, see `processor/elasticapmprocessor/internal/enrichments/resource.go:187-211`) is available at this point but unused.

### Why it matters

Same service, same instance, two conflicting `agent.name` values purely because of which signal path produced the aggregate. Reproduced with both a Node.js and a Java EDOT auto-instrumented service on 9.4.2, no legacy instrumentation involved.

Sample affected document (`service_summary.1m.otel` data stream):

```json
{
  "resource": {
    "attributes": {
      "agent.name": "unknown",
      "deployment.environment": "prod",
      "service.name": "my-service",
      "telemetry.sdk.language": "nodejs"
    }
  },
  "scope": { "name": "github.com/open-telemetry/opentelemetry-collector-contrib/connector/signaltometricsconnector" }
}
```

### Suggested fix

Derive `agent.name` for the metrics/logs signal paths the same way `elasticapmprocessor` does for traces (e.g. `opentelemetry/<telemetry.sdk.language>`), instead of defaulting to the literal string `"unknown"`.

macdewee:
agent.name previously defaulted to the literal string "unknown" for aggregates (service_summary, etc.) built from the metrics/logs signal path, since only the traces path had it set upstream by elasticapmprocessor. This meant a single service instance could end up with two different agent.name values depending on which signal produced the aggregate.

Derive it the same way elasticapmprocessor does for traces, from telemetry.sdk.name/telemetry.sdk.language/telemetry.distro.name, before handing off to the wrapped signaltometricsconnector. The "unknown" default remains as a fallback for resources with no telemetry.* attributes.

Closes https://github.com/elastic/opentelemetry-collector-components/issues/1319

### DIFF
--- a/connector/elasticapmconnector/connector.go
+++ b/connector/elasticapmconnector/connector.go
@@ -181,9 +181,13 @@ func (e *logsResourceEnricher) Capabilities() consumer.Capabilities {
 
 // setAgentName derives the agent.name resource attribute from
 // telemetry.sdk.name, telemetry.sdk.language and telemetry.distro.name,
-// the same way elasticapmprocessor does for traces. It is a no-op if
-// agent.name is already set, e.g. by a classic Elastic APM agent, or if
-// none of the telemetry.* attributes are present.
+// the same way elasticapmprocessor does for traces (see setAgentName in
+// processor/elasticapmprocessor/internal/enrichments/resource.go; keep the
+// two in sync). It is a no-op if agent.name is already set, e.g. by a
+// classic Elastic APM agent. Otherwise it always sets a value, defaulting
+// to "otlp" when no telemetry.* attributes are present, so that a service
+// never ends up with a different agent.name on its traces than on its
+// metrics/logs-derived aggregates.
 func setAgentName(resource pcommon.Resource) {
 	attrs := resource.Attributes()
 	if _, ok := attrs.Get("agent.name"); ok {
@@ -202,9 +206,6 @@ func setAgentName(resource pcommon.Resource) {
 		}
 		return true
 	})
-	if sdkName == "" && sdkLanguage == "" && distroName == "" {
-		return
-	}
 
 	agentName := "otlp"
 	if sdkName != "" {
--- a/connector/elasticapmconnector/connector.go
+++ b/connector/elasticapmconnector/connector.go
@@ -19,10 +19,14 @@ package elasticapmconnector // import "github.com/elastic/opentelemetry-collecto
 
 import (
 	"context"
+	"fmt"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/connector"
 	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 	"go.opentelemetry.io/collector/processor"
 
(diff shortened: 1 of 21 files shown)


macdewee:
@elastic/ingest-otel-data, @elastic/obs-ds-intake-services, @elastic/obs-ds-hosted-services Can I get review on this one please?

connector/elasticapmconnector/connector.go
```diff
@@ -127,6 +143,87 @@ func (e *spanEnricher) Capabilities() consumer.Capabilities {
...
+// setAgentName derives the agent.name resource attribute from
+// telemetry.sdk.name, telemetry.sdk.language and telemetry.distro.name,
+// the same way elasticapmprocessor does for traces (see setAgentName in
+// processor/elasticapmprocessor/internal/enrichments/resource.go; keep the
+// two in sync). It is a no-op if agent.name is already set, e.g. by a
+// classic Elastic APM agent. Otherwise it always sets a value, defaulting
+// to "otlp" when no telemetry.* attributes are present, so that a service
+// never ends up with a different agent.name on its traces than on its
+// metrics/logs-derived aggregates.
+func setAgentName(resource pcommon.Resource) {
```
lahsivjar, line 195: What is the reason for doing it here instead of in the elasticapmprocessor? Additionally, this will only apply to the aggregated metrics, is that what we want?

macdewee: actually we need to make the change in both places, because:
- connector can be used without the processor in the pipeline
- when using [skip_enrichment](https://github.com/elastic/opentelemetry-collector-components/tree/aaeb978eb2a7895c0a6cb0613522b844bb54b9f1/processor/elasticapmprocessor#configuration) in processor we do not set the agent's name for logs/metrics but we still do for traces

> Additionally, this will only apply to the aggregated metrics, is that what we want?

As the connector output is the aggregated metrics, we should restrict to them. if someone wants to change the other flowing signals, the should use proccesors

lahsivjar: The connector (or rather the aggregation service) will always use the data enriched by the ingest collector and keeping enrichment logic at multiple places is pain to maintain long term. I would lean towards keeping the enrichment in the apm processor - we can change how skip_enrichment works if required.

> connector can be used without the processor in the pipeline

That should break the service, no? If processor is not used then that is the wrong/incomplete pipeline. Are there any edge cases where processor SHOULDN't be run?

macdewee: > I would lean towards keeping the enrichment in the apm processor - we can change how skip_enrichment works if required

It would be beneficial to unify how `skip_enrichment` works, but the question is what can we change?

1. always enrich agent name (ignore `skip_enrichment`)
2. skip enrichment for traces (sounds like breaking change)

> Are there any edge cases where processor SHOULDN't be run?

I am not familiar with apm enough to judge it, but the connector's readme doesn't point that the processor is required. Also looks like both solutions for processor are breaking change (sending data which were explicitly skipped before or stop sending data which were send)

hegerchr: > Also looks like both solutions for processor are breaking change (sending data which were explicitly skipped before or stop sending data which were send)

Do you know what breaks exactly? We have features in Kibana depending on the agent.name and analytics using it.

macdewee: By breaking change I mean it is significant change if we were ignoring `skip_enrichment` for traces and now we will stop doing that

I am trying to understand what is the proposed solution. Is it to always set the agent name regardless of the `skip_enrichment` option?

lahsivjar: `skip_enrichment` is a bit of misnomer as what it intends to do is to define enrichment for `ecs` mode. Can you confirm if the `agent.name` enrichment is required for OTel or for ECS (this is data sent for elastic agents, the non-otel ones OR data sent via intake-v2 protocol)?

lahsivjar: > I am not familiar with apm enough to judge it, but the connector's readme doesn't point that the processor is required.

I don't think this is a requirement of the connector, rather, it should be pointed out in the Elastic agent docs or in MOTel pipelines.

MiriamAparicio: Hi @macdewee,
`agent.name` is a first-class signal in Kibana APM, this will definitely break the APM UI, some of the usage on top of my head now:

- Service identity / icons / inventory
- Metrics dashboard selection (classic vs OTel via getDashboardFileName)
- Agent-scoped metric queries (e.g. Java heap/GC filter on agent.name)
- Agent explorer, agent config, telemetry

...might be more...
I think there are some API paths with a fallback, don't remember which ones, but they do something like: `getAgentName(agent.name, telemetry.sdk.language, telemetry.sdk.name)`, but many ES queries filter on `agent.name` directly
Changing `skip_enrichment` to stop enriching traces would break APM. Always enriching `agent.name` (carving it out of `skip_enrichment`) is the safer direction for the UI.
APM can tolerate always setting `agent.name` , it cannot tolerate removing it from traces, "unknown" on aggregates is already a problem for the UI , it can show wrong icons, pick the wrong metrics dashboards, or split one service into inconsistent agent identities.
I would push back on anything that makes traces less enriched.

lahsivjar: > this will definitely break the APM UI, some of the usage on top of my head now:

@MiriamAparicio do you mean the UI will break if the `agent.name` changes are only done in the connector (affects aggregated data only)?

lahsivjar:
This is not the correct solution, see:

https://github.com/elastic/hosted-otel-collector/pull/1595
https://github.com/elastic/elastic-agent/pull/10959

Closing this PR, feel free to reopen if the rationale doesn't sound correct.

PR #1323, opened 2026-07-13
[Metricsgen] Add metricsgen-benchmark flag to simulate soak mode

gizas:
With `-metricsgen-benchmark=false`:
(and start_now_minus: 10m insie config-prw.yaml)

```shell
tmp/otelbench -metricsgen -metricsgen-benchmark=false -config config-prw.yaml

BenchmarkOTelbench/metricsgen          1        2499062375 ns/op                 2.499 duration_s                0 failed_metric_points/s            34253 metric_points/s
```

See below that with false runs for 2.49 secs and creates metrics for 10m

Screenshot 2026-07-13 at 6 27 15 PM

With -metricsgen-benchmark=true it uses the becnhmark mode and 

```
/tmp/otelbench -metricsgen  -config config-prw.yaml

BenchmarkOTelbench/metricsgen          1        1252625708 ns/op                 1.253 duration_s                0 failed_metric_points/s             1708 metric_points/s
```


carsonip (commented):
More of a high level question before diving into the technical details: since this is for the soak use case, is there a reason why we don't introduce a separate code path that spins up a collector with `run_indefinitely` instead of using testing.Benchmark?

edit: Although i think it'd be fine if we use testing.Benchmark with run_indefinitely. It is unclear to me why we would want a load that terminates itself in a soak mode.

gizas:
@carsonip sorry for big delay here, so to transform your question above:

The current metricsgen code path has by default N=1.

```
/tmp/otelbench -metricsgen -config config-prw.yaml
```
(This is the default -metricsgen-benchmark=true)

In that mode, otelbench uses Go’s testing.Benchmark, and b.N is passed into metricsgen.

Use this if you want the start_now_minus from config-prw.yaml:

```yaml
/tmp/otelbench -metricsgen -metricsgen-benchmark=false -config config-prw.yaml
```
(so -metricsgen-benchmark=false and it runs for the period u define)

For true soak mode, use:

```
/tmp/otelbench -metricsgen -metricsgen-benchmark=false -config config-prw-soak.yaml -duration-metrics=10m
```

See that this has `run_indefinitely: true` so it is close to soak. 
Let me know wdyt

carsonip (commented):
- As `config-prw-soak.yaml` has run_indefinitely: true, what does `-duration-metrics=10m` mean?
- I'll take another look at how to make it more ergonomic to use, especially around using it in soak mode with an embedded yaml, and configurable through flags

gizas:
> what does -duration-metrics=10m mean?

Means run metricsgen indefinitely based on the config, but let otelbench stop the collector after 10 minutes
Seehttps://github.com/elastic/hosted-otel-collector/blob/main/.buildkite/pipeline.tests-staging.yaml#L35-L48, we have execution_time and timeout_in_minutes: 15 to stop the soak. Something like that

carsonip (approved):

PR #1326, opened 2026-07-14
[Extension][beats_encoding]  feat: use maps instead of dotted field names

Kavindu-Dodan:
### Overview

This PR improves `beats_encoding` extension by using maps for `event.*` & `dataset.*` fields. 

Prior to this change, extension used dotted field keys such as `event.created`.

Note - Updated go version for vulnerability check as it's failing

### Why

Dotted field keys has known incompatibility with ingest pipelines. The fix proposed here helps to utilize this extension with existing pipelines without modifications.

### DIFF
--- a/extension/beatsencodingextension/testdata/json_nested_complex_expected.yaml
+++ b/extension/beatsencodingextension/testdata/json_nested_complex_expected.yaml
@@ -21,23 +21,31 @@ resourceLogs:
                   - key: Nano
                     value:
                       intValue: "1779801400123456000"
+                  - key: event
+                    value:
+                      kvlistValue:
+                        values:
+                          - key: dataset
+                            value:
+                              stringValue: custom
+                  - key: data_stream
+                    value:
+                      kvlistValue:
+                        values:
+                          - key: type
+                            value:
+                              stringValue: logs
+                          - key: dataset
+                            value:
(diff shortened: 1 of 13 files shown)


extension/beatsencodingextension/testdata/json_nested_complex_expected.yaml
```diff
@@ -93,23 +109,31 @@ resourceLogs:
...
+                          - key: type
+                            value:
+                              stringValue: logs
+                          - key: dataset
+                            value:
+                              stringValue: custom
+                          - key: namespace
+                            value:
+                              stringValue: default
                   - key: input.type
```
MichaelKatsoulis: Is `input.type` intentionally left as dotted?

MichaelKatsoulis (approved):
I have just one question regarding `input.type`. 
Looks very good!

extension/beatsencodingextension/testdata/json_nested_complex_expected.yaml
Kavindu-Dodan: Only focused on event.* and data_stream.* fields. But it makes sense to convert all body fields. I will address with next commit.

girodav (approved):

PR #1337, opened 2026-08-04
Fix govulncheck failures in CI

### LINKED ISSUE #1252: [CI] Revert the hardcoded value of go1.26.4
The [CI  gogovulncheck](https://github.com/elastic/opentelemetry-collector-components/actions/runs/26873031410/job/79252950777?pr=1250) were failing  because  of go version vulnerabilities:
   ```yaml 
   Found in: net/textproto@go1.26.3
    Fixed in: net/textproto@go1.26.4
```

This PR https://github.com/elastic/opentelemetry-collector-components/pull/1250/changes#diff-3ea54af4839eb75404d71b28252bead7e7ec8f676b1f815e1cde02629a75c165R110 pinned the 1.26.4 version to make CI pass.

According to https://go.dev/doc/devel/release#go1.26.minor the 1.26.4 released yesterday and soonish the stable flag should recognise it
This story will revert the above hardcoded value

### PR DESCRIPTION
**Pull request by gizas**
Fixes https://github.com/elastic/opentelemetry-collector-components/issues/1252

Fix gogovulncheck errors in CI:
See https://github.com/elastic/opentelemetry-collector-components/actions/runs/30792955755/job/91620191800?pr=1323

Mailny: 
Vuln | library | Version
-- | -- | --
GO-2026-5970 | golang.org/x/text | ≥ v0.39.0
GO-2026-6061 | google.golang.org/grpc | ≥ v1.82.1

### CODE REVIEW

File: .github/workflows/checks.yml [hunk 1]
```diff
@@ -107,7 +107,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v6
         with:
-          go-version: '1.26.5'
+          go-version: 'stable'
```
carsonip: q: does this line make a difference?

carsonip (approved):

.github/workflows/checks.yml
gizas: not in this PR specifically at this point of time but it is time to Fix https://github.com/elastic/opentelemetry-collector-components/issues/1252

mrodm (approved):

gizas:
@elastic/ingest-otel-data , @elastic/ingest-managed-jobs , @elastic/security-service-integrations can I have a review please?

straistaru (approved):

zmoog (approved):

gizas:
@elastic/ingest-otel-data , @elastic/security-service-integrations could I have a review please?

.github/workflows/checks.yml
florianl: Using `stable` at this point is misleading, I think. Looking at random components, we see that most of them use `oldstable` and so `oldstable` should be used here as well, otherwise govulncheck will miss certain CVEs.

florianl: Maybe a suggestion:
Instead of the various indirections via `Makefile`, maybe consider adding a `go.mod` at the top level to be able to add tools like, `govulncheck`, `lint` and others as regular go tool.

florianl: I showcased the approach in https://github.com/elastic/opentelemetry-collector-components/pull/1342.

gizas: @florianl I see that oldstable is also followed in upstream and by a quick check I see 28 modules here follow the oldstable.

So np with me, we can change this to oldstable. So quicker we will catch any issues (and of course this depends on what versions our internal modules have)

But now that I change it to oldstable I have:

```yaml
Expected. oldstable is 1.25.12 with GOTOOLCHAIN=local, but these modules require 1.26.0:
```

For:
processor/elasticapmprocessor
processor/elastictraceprocessor
receiver/elasticapmintakereceiver

So we have 3 with major versions and we should downgrade :) 
Shall I leave it with stable then? This is what it was. 

Also for your second suggestion I opened ~(or I will open a second PR [here](https://github.com/elastic/opentelemetry-collector-components/compare/fix_go_mod?expand=1) )~ https://github.com/elastic/opentelemetry-collector-components/pull/1344
I would say not to have a common go.mod for 33 modules but better to have the tools build in common way

- Standard Go — same pattern as upstream collector/contrib and what Go 1.24+ recommends.  No fake tools.go imports as we have toda
- tool versions live in go.mod/go.sum
- Now the local make command can not drift from what is run in CI

florianl (approved):

PR #1344, opened 2026-08-10
[Make build tools] Change to common tools 

gizas:
This PR updates the tools.go from building binaries into .tools/ (via tools.go) to Go native `-tool` directives .
Update Makefile.Common, otelbench changelog targets, and CI/workflows so they resolve the new tools.

Pros:

- Standard Go — same pattern as upstream collector/contrib and what Go 1.24+ recommends. No fake tools.go imports as we have toda. See [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/286294e8e029a10fc80427174013a9d8e51b4ff4/Makefile.Common#L69)
- tool versions live in go.mod/go.sum
- Now the local make command can not drift from what is run in CI

Related to: https://github.com/elastic/opentelemetry-collector-components/pull/1337#discussion_r3748776178


florianl (approved):

.github/workflows/bump_otelbench.yaml
```diff
@@ -66,23 +66,9 @@ jobs:
           go-version-file: internal/tools/go.mod
           cache: 'false'
 
-      - name: Cache go tools
-        id: go-tools-cache
-        uses: actions/cache@v5
```
carsonip, line 71: non blocking q: why are we removing the cache on CI? we can keep the mod cache and build cache right? then make install-tools would transparently speed up if cache exists

.github/workflows/bump_otelbench.yaml
```diff
@@ -66,23 +66,9 @@ jobs:
...
-          path: ./.tools
-          key: go-tools-${{ runner.os }}-${{ hashFiles('internal/tools/go.sum') }}
-
-      - name: Add .tools to PATH
-        run: echo "$GITHUB_WORKSPACE/.tools" >> $GITHUB_PATH
-
-      # If there is no cache with the go tools,
-      # install them. We need the changelog
-      # tool for this workflow to work.
+      # Warm the go tool cache (chloggen is required below).
```
carsonip: nit: this is a misleading comment. It doesn't "warm" anything because there is no cache to begin with

carsonip (approved):
approving to unblock. Some nits

.github/workflows/bump_otelbench.yaml
gizas: [fix comment](https://github.com/elastic/opentelemetry-collector-components/pull/1344/commits/8bb93085e59a2727d39ee144886b00259ca1050b)

.github/workflows/bump_otelbench.yaml
gizas: The above cache was only for the old tools.
So I think was not needed 

So sth like the below:
```yaml
- uses: actions/setup-go@v6
  with:
    go-version-file: internal/tools/go.mod
    cache: 'false'
- name: Cache Go
  id: go-cache
  uses: actions/cache@v5
  with:
    path: |
      ~/go/pkg/mod
      ~/.cache/go-build
    key: go-tools-${{ runner.os }}-${{ hashFiles('internal/tools/go.sum') }}
    restore-keys: |
      go-tools-${{ runner.os }}-
```

or why dont we do `cache: 'true'`

But again it will for the cache of make install-tool on rerun

carsonip: let's test cache: true later as a follow up

carsonip (approved):

PR #1369, opened 2026-08-17
[connector/dynamicrouting]Introduce static routing for known keys and pipelines

lahsivjar:
Allows the connector to route to known-pipelines based on OTTL conditions. If static routing is defined then dynamic-routing based cardinality estimation is skipped. The docs are updated with more details on how this works and the configuration shapes.

The conditionals for OTTL introduced in this PR will eventually be introduced as conditional pipelines, something like:

```
routing_pipelines:
  - pipelines: []pipeline.ID # will serve as default pipeline if contitional_pipelines are NOT defined
    max_cardinality: float64
    conditional_pipelines:
      - conditions: []string # OTTL conditions
        pipelines: []pipeline.ID
```

Benchmark results:

```
no_static_routes     182 ns/op    172 B/op    4 allocs/op
first_match           70 ns/op     48 B/op    3 allocs/op   ← fastest, short-circuits before dynamic routing
no_match/1_routes    238 ns/op    206 B/op    6 allocs/op
no_match/10_routes   760 ns/op    495 B/op   24 allocs/op
```
As expected, there are no surprises. With small number of static routes (<=2) the impact on performance will be around 100ns.

Related to: https://github.com/elastic/hosted-otel-collector/issues/3634


connector/dynamicroutingconnector/router.go
```diff
@@ -222,15 +254,56 @@ func (r *router[C]) Shutdown(ctx context.Context) error {
...
+	}
+	var zero C
+	return zero, false
+}
+
+// newStaticConditionSequence compiles a list of OTTL condition strings into a
+// reusable condition sequence evaluated against the ottlotelcol context.
+// Conditions are OR'd: any matching condition triggers the route.
+func newStaticConditionSequence(conditions []string, settings component.TelemetrySettings) (*ottl.ConditionSequence[*ottlotelcol.TransformContext], error) {
+	parser, err := ottlotelcol.NewParser(
```
constanca-m, line 287: I am not against OTTL, but do you see us adding any more conditions?

Because we are pulling the whole OTTL just to check `len(meta.Get("x-elastic-source-id)) > 0`

I see that it is just 50ns and 2 allocs, but it feels so much was pulled here just for a condition

connector/dynamicroutingconnector/router.go
```diff
@@ -222,15 +254,56 @@ func (r *router[C]) Shutdown(ctx context.Context) error {
...
+		return c
+	}
+	meta := client.FromContext(ctx).Metadata
+	pk := r.partitionKey(meta)
+	return r.getNextConsumerAndMaybeRecord(ctx, pk, meta)
+}
+
+func (r *router[C]) matchStaticRoute(ctx context.Context) (C, bool) {
+	for _, sr := range r.staticRoutes {
+		tCtx := ottlotelcol.NewTransformContextPtr()
```
constanca-m, line 268: Why not create it before the `for` and reuse it?

```go
	var zero C
	if len(r.staticRoutes) == 0 {
		return zero, false
	}
	tCtx := ottlotelcol.NewTransformContextPtr()
	defer tCtx.Close()
	for _, sr := range r.staticRoutes {
		match, err := sr.conditions.Eval(ctx, tCtx)
...
```

connector/dynamicroutingconnector/config.go
```diff
@@ -55,6 +69,18 @@ func (c *Config) Validate() error {
...
 	}
+	nopSettings := component.TelemetrySettings{Logger: zap.NewNop()}
+	for i, sr := range c.StaticRoutes {
+		if len(sr.Conditions) == 0 {
+			return fmt.Errorf("static_routes[%d]: at least one condition must be specified", i)
+		}
+		if _, err := newStaticConditionSequence(sr.Conditions, nopSettings); err != nil {
+			return fmt.Errorf("static_routes[%d]: invalid condition: %w", i, err)
+		}
+		if len(sr.Pipelines) == 0 {
```
constanca-m: Maybe put this before `newStaticConditionSequence`?

connector/dynamicroutingconnector/router.go
lahsivjar: The conditions are parsed ONCE during startup - they are NOT on the hot-path. Having conditional flexibility allows for doing routing based on data types and other functions without having any extra complexity in the code or the config declaration.

What are your concerns with multiple conditions?

connector/dynamicroutingconnector/router.go
lahsivjar: There should be minimal performance impact of keeping this in for loop (it is backed by a sync pool). I think keeping it in for loop gives better readability and we hold on to the context ptr for much less time, wdyt?

connector/dynamicroutingconnector/router.go
constanca-m: I put this comment here just because I was talking about OTTL, I was actually referring to the use of it. The parse happens once, but then we do eval every time we have a condition

> based on data types and other functions without having any extra complexity in the code or the config declaration.

Well, this parser is mostly for metadata. If we wanted other types like log body, we would need a new parser

Anyway, this doesn't matter much

connector/dynamicroutingconnector/router.go
constanca-m: I'm ok with it

constanca-m (approved):

connector/dynamicroutingconnector/router.go
lahsivjar: > Well, this parser is mostly for metadata. If we wanted other types like log body, we would need a new parser

Right, I referred to that only as an example (a bad one in hindsight). Theoretically since it allows full OTTL we can do what we want using just one condition too (like ottl_1 AND ottl_2 AND ottl_3), this format gives a bit more readability to such cases. It is also consistent with the support provided by other upstream components like routing connector and signal to metrics connector.

PR #1370, opened 2026-08-17
chore(deps): bump OTel Collector to v0.158.0 in internal/exporter/httpexporter

andrzej-stencel:


carsonip (commented):
@gizas is going to move it out of internal. It is causing issues for us as well

carsonip (approved):
approving to unblock

PR #1406, opened 2026-08-18
[extension/apikeyauth] change error handler to return `codes.ResourceExhausted` for exhausted ES 502 and 503 responses

isaacaflores2:
# Overview

This PR updates the `extension/apikeyauth` logic for handling typed  errors (`ElasticsearchError`) that return a 502 or 503. These errors indicated the ES instance is not healthy and the ES client will internally retry (when retries are enabled).  The extension will now return `codes.ResourceExhausted` so clients can throttle and retry,

### DIFF
--- a/extension/apikeyauthextension/authenticator.go
+++ b/extension/apikeyauthextension/authenticator.go
@@ -414,7 +414,9 @@ func (a *authenticator) Authenticate(ctx context.Context, headers map[string][]s
 			switch elasticsearchErr.Status {
 			case http.StatusUnauthorized, http.StatusForbidden:
 				return ctx, status.Error(codes.Unauthenticated, err.Error())
-			case http.StatusTooManyRequests, http.StatusGatewayTimeout:
+			case http.StatusTooManyRequests, http.StatusGatewayTimeout, http.StatusBadGateway, http.StatusServiceUnavailable:
+				// 502/503/504/429 are transient ES overload/unavailability errors (retried by the ES client when enabled);
+				// surface as backpressure so callers throttle+retry
 				return ctx, errorWithDetails(status.New(
 					codes.ResourceExhausted, "failed to check privileges",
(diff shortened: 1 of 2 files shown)


extension/apikeyauthextension/authenticator.go
```diff
@@ -414,7 +414,9 @@ func (a *authenticator) Authenticate(ctx context.Context, headers map[string][]s
 			switch elasticsearchErr.Status {
 			case http.StatusUnauthorized, http.StatusForbidden:
 				return ctx, status.Error(codes.Unauthenticated, err.Error())
-			case http.StatusTooManyRequests, http.StatusGatewayTimeout:
+			case http.StatusTooManyRequests, http.StatusGatewayTimeout, http.StatusBadGateway, http.StatusServiceUnavailable:
```
constanca-m, line 417: I don't fully get it.

I read your comment https://github.com/elastic/hosted-otel-collector/issues/3633#issuecomment-5331679452

On one of your screenshots we see:

Image

How is this request running for over 1k seconds?

constanca-m: I am being dumb, this is not 1k seconds, but 1k ms.

I wonder if it is worth it caching the target as unavailable for a few seconds to stop the cascade of retries when it is down?

constanca-m (approved):
Thanks Isaac, looks good!

carsonip (approved):

extension/apikeyauthextension/authenticator.go
isaacaflores2: Hmm that would help reduce the load on ES but I do not know if we should cache a temporary error like this. I think it will be a good follow up to add this type of negative cache/circuit breaker if we continue to have issues. A similar strategy would be to not retry 503s or increase the back off interval. Thanks!

PR #1418, opened 2026-09-02
Add dynamicbatch processor

lahsivjar:
The dynamicbatch processor allows creating isolated batch processors (with its own sending queue and batching configs) based on metadata keys. The batch processors are GCd if they are idle for a long time. The idle timeout can be configured.


constanca-m:
This dynamic batching processor is meant to replace the adaptive concurrency processor we have. Basically we would go from this YAML:

```yaml
exporters:
  elasticsearch:
    sending_queue:
      enabled: true
      num_consumers: 200
      wait_for_result: true
      block_on_overflow: true
processors:
  adaptive_concurrency/by_target:
    metadata_keys: [x-elastic-target-id]
    shared_slots: 200          # copied from elasticsearch.sending_queue.num_consumers
    max_share: 0.1             # factory default if omitted
    slow_request_threshold: 6s # factory default if omitted
service:
  pipelines:
    traces:
      processors: [adaptive_concurrency/by_target]
      exporters: [elasticsearch]
    # logs and metrics: same processor, same exporter
```

To this:

```yaml
exporters:
  elasticsearch:
    sending_queue:
      enabled: false
processors:
  dynamicbatch/by_target:
    metadata_keys: [x-elastic-target-id]
    wait_for_result: true
    block_on_overflow: true
    # factory defaults if omitted: num_consumers 10, metadata_cardinality_limit 1000
service:
  pipelines:
    traces:
      processors: [dynamicbatch/by_target]
      exporters: [elasticsearch]
    # logs and metrics: same processor, same exporter
```

So basically my question is: **what are we gaining from this change**?

I am going to use last night's data from `prd-awsuse1-hosted-1` because that's the first time we have proper data for how many resources each target can take independently.

I wrote much of this data on https://github.com/elastic/hosted-otel-collector/issues/3569#issuecomment-5506932132.

Same 12 targets that peaked at `in_flight` 30, each with the HTTP P99 we actually measured. The pair that mattered is already in this table: slow `324d2a` at 48s held 30 of the 200, healthy `8ace85` at 3s needed 15.

| tenant | `/_bulk` delay (HTTP P99) |
|---|---|
| `324d2a444cb34f1398497d78b58a0a77` | 48s |
| `4c85028238adaddb331d0e184a2617fa` | 6s |
| `8ace8532fe9c42208a5f50e09651c163` | 3s |
| `cdb79ee231b84d72a4ce9a3e75f29124` | 24s |
| `6f35a93526a341ffb55484b2af4cec18` | 3s |
| `bacf47c46cdf49a58e595760b7e361cc` | 18s |
| `51550d253d825aa4a16c6e7024b2f420` | 0.38s |
| `6e3fcee832104fbdad1a63427941f7bc` | 3s |
| `5ac984baeeff4bd89c566035d280569f` | 0.56s |
| `1bbc8834e0f24c2cb680052dcfafe301` | 0.18s |
| `f5e72aacd664f50aeb0c0870583902e2` | 0.38s |
| `27a1d13cfa7117b5252e05d9e98dc63a` | 0.12s |

The thing we hold is an in-flight `/_bulk`: the body, the HTTP connection, the worker. It stays in the pod until ES answers. Fast tenants (`0.12s`-`0.56s`) recycle that slot. Slow ones (`18s`, `24s`, `48s`) will hold it. We saw that last night.

With dynamicbatch, every target gets a fixed shard (`num_consumers` 10 if we take the factory default, 20 if we match today's `max_share` ceiling). There is no way to shrink that. So:

- Factory 10: last night's 12 is 120 `/_bulk` in flight (10 slots * 12 targets). The replay held `324d2a` at 10 on a 48s ES (adaptive left it at 1) and stuck `8ace85` at 10. Fat took 81s to drain. Heap was 71MB.
- Workers 20: last night's 12 is 240 `/_bulk` in flight (20 slots * 12 targets), already over today's pod cap (which is 200). The replay held `324d2a` at 20 and `8ace85` at 20. Fat drained in 42s. Heap was 79MB.

With adaptive, all 12 targets share one pool of 200 workers (`shared_slots`). `max_share: 0.1` only sets the per-target ceiling: 20 of those 200. It is not the count a 48s ES should keep. The 6s timer is. A call slower than 6s halves that target toward 1. A healthy call can climb toward 20. I replayed last night's 12 delays. Adaptive held slow `324d2a` at 1 worker, let fat `8ace85` use 20, and the 12 together used 124 of the 200.

So using this processor removes the shared queue, which I think is a good approach, but just kicks the problem somewhere else. And removing a shared queue removes the only knob we have that is bounding the pod (num consumer, max bytes on the shared queue). We would then have to invent a global cap again, which is what `shared_slots` is atm.

Slow tenants would still hold the buffers, because dynamicbatch never takes the slot away. Fast tenants would still clean up. This is the behavior we saw last night.

I don't think this tradeoff is worth it. I think our shared pool is not finely tuned at all with the new approach, but nuking it is not the fix.

Out of curiosity, I ran a quick benchmark that repeated this exact scenario of last night and got:

| | adaptive | db 10 | db 20 |
|---|---:|---:|---:|
| slow `324d2a` in flight | **1** | 10 | 20 |
| fat `8ace85` in flight | **20** | 10 | 20 |
| total in flight | **124** | 120 | **240** |
| heap | **12MB** | 71MB | 79MB |

It's quite a big difference. I would take this numbers with a grain of salt, but I think the numbers for memory are not that far off.

PR #1457, opened 2026-09-07
test(receiver/entity_analytics): harden flaky integration test `TestADReceiverLifecycle_ES`

andrzej-stencel:
`TestADReceiverLifecycle_ES` fails intermittently in CI with "timed out waiting for N log records", both on PRs and on main. The in-memory variant never fails, so the culprit is timing against the freshly started Elasticsearch container.

- Wait for cluster health yellow before running tests; GET / returns 200 before the cluster state is recovered and document requests issued in that window block.
- Pre-create the test index so the first commit under test does not pay for auto-creation and dynamic mapping.
- Capture receiver logs with a zap observer and fail fast with the underlying error when the receiver reports "sync failed", instead of an opaque 30s timeout.
- Raise the ES-backed wait budgets (60s/30s) and set the module GOTEST_TIMEOUT to 480s, as the package starts one ES container per ES-backed test and was running close to the 240s default.

Assisted-by: Claude Fable 5.1

### DIFF
--- a/receiver/entityanalyticsreceiver/Makefile
+++ b/receiver/entityanalyticsreceiver/Makefile
@@ -1,1 +1,5 @@
+# The integration tests start a dedicated Elasticsearch container per
+# ES-backed test, so the package needs more than the default budget.
+GOTEST_TIMEOUT=480s
+
 include ../../Makefile.Common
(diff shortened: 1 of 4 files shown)


efd6 (approved):
Thanks

PR #1462, opened 2026-09-09
fix(apikeyauth): normalize API key authorization scheme

zmoog:
## Summary

Normalize accepted API key authorization headers before forwarding them to Elasticsearch and deriving the cache fingerprint.

HTTP authentication schemes are case-insensitive, so `APIKey <credential>` and `ApiKey <credential>` identify the same API key. Previously, parsing accepted both representations but PBKDF2 hashed the original header. Once one representation populated the ID-based cache entry, the other was mistaken for the same ID with a different secret and rejected with HTTP 401 until the cache expired.

## Reproduction

Against a deployed Managed OTLP endpoint with a 35-second auth cache TTL:

- prime with `APIKey`: 40/40 returned 200
- immediately use `ApiKey`: 40/40 returned 401
- expire the cache and prime with `ApiKey`: 40/40 returned 200
- immediately use `APIKey`: 40/40 returned 401

## Changes

- canonicalize valid API key headers to `ApiKey <canonical-base64>`
- use that canonical value for both `_has_privileges` and the cache fingerprint
- verify both scheme orders share one cache entry and make only one Elasticsearch request
- preserve the existing same-ID/different-secret collision behavior

## Testing

- `make test` (race detector enabled, 94.3% statement coverage)
- `make lint` (0 issues)
- `make fmt`
- `git diff --check`

### DIFF
--- a/extension/apikeyauthextension/authenticator.go
+++ b/extension/apikeyauthextension/authenticator.go
@@ -220,7 +220,7 @@ func (a *authenticator) Shutdown(ctx context.Context) error {
 }
 
 // parseAuthorizationHeader checks that the Authorization header follows the expected format,
-// and returns the full header value, and the API Key ID.
+// and returns the canonical header value and the API Key ID.
 func (a *authenticator) parseAuthorizationHeader(headers map[string][]string) (string, string, error) {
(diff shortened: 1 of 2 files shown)


isaacaflores2 (approved):
Thanks for the update!

carsonip (approved):

PR #1469, opened 2026-09-17
Add new confighttp Server config for prw receiver

gizas:
Relates  #https://github.com/elastic/hosted-otel-collector/issues/3927 

Specifically:

- Switches the Prometheus remote Write receiver to use  confighttp.NewDefaultServerConfig()
- Adds a test that confirms that default configuration sets `KeepAlivesEnabled: true`


constanca-m (approved):
Thanks!

carsonip (approved):

